PR #484, opened 2026-03-13
feat: configurable brokerage cash reserve

### LINKED ISSUE #407: Untouchable cash reserve for rebalancing
Rebalancing currently operates on the full Alpaca cash balance. Need the ability to designate a portion of the cash balance as off-limits so rebalancing only operates on available balance minus the reserve. This prevents rebalancing from draining the account below a safety threshold.

0xgleb:
## Motivation

Closes [RAI-68](https://linear.app/makeitrain/issue/RAI-68/untouchable-cash-reserve-for-rebalancing).

USDC auto-rebalancing operates on the full Alpaca cash balance, which can
deplete the brokerage account entirely — leaving no buffer for operational
needs like margin requirements or manual interventions. There's no way to
tell the system "keep at least $X in the brokerage account."

## Solution

Adds an optional `reserved` field to `CashAssetConfig` (USD amount) that is
subtracted from the reported offchain cash balance before rebalancing
decisions are made. Negative results are preserved (they represent margin
debt or an under-reserved account and are meaningful signals downstream).

### How it works

1. **Config**: `reserved = 50000.0` under `[assets.cash]` sets a $50k reserve
2. **Construction**: the `Usd` reserve is converted to `i64` cents once in
   `InventoryPollingService::new()` — errors surface at startup, not per poll
3. **Inventory polling**: `poll_and_record` does `inventory.usd_balance_cents
   .checked_sub(self.reserved_cash_cents)` and emits the result as the
   `OffchainUsd` snapshot event — no Float arithmetic in the hot path
4. **Rebalancing sees reduced balance**: downstream triggers operate on the
   post-reserve cash; if the reserve exceeds available cash, the ratio
   computation sees a negative number and acts accordingly

### Typed error for Usd → cents conversion

`Usd::to_cents()` now returns `Result<i64, UsdToCentsError>` instead of
`Option<i64>`, with named variants (`SubCentPrecision`, `Overflow`,
`Arithmetic`) each carrying the offending value — so when config yields
`reserved = 50.005`, the startup error names both the field and the value.

### E2E coverage for the subtraction path

`usdc_imbalance_triggers_base_to_alpaca` now configures a $3k reserve and,
after the rebalance flow settles, calls `assert_offchain_usd_reflects_reserve`
— which polls until the latest `OffchainUsd` snapshot equals
`broker.cash_balance() - reserved` in cents. If the reserve subtraction
regresses, the test fails.

## Checks

- [x] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs
- [ ] included screenshots (if this involves a change to the dashboard)

## Release Notes

* **New Features**
  * Configurable cash reserve setting that reduces available cash for rebalancing operations.
  * Dashboard now displays configured cash reserve amount in settings panel.
  * Unified dashboard layout simplifies navigation across mobile and desktop.

* **Bug Fixes**
  * Improved USD conversion error handling with explicit error reporting.

* **Testing**
  * Enhanced integration tests for cash reserve accounting scenarios.
  * Dashboard CI now includes linting validation step.


0xgleb:
* **#484** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/484?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#492](https://github.com/ST0x-Technology/st0x.liquidity/pull/492) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/492?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#506](https://github.com/ST0x-Technology/st0x.liquidity/pull/506) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/506?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/484?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#483** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/483?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 7 other dependent PRs ([#490](https://github.com/ST0x-Technology/st0x.liquidity/pull/490) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/490?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#495](https://github.com/ST0x-Technology/st0x.liquidity/pull/495) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/495?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#498](https://github.com/ST0x-Technology/st0x.liquidity/pull/498) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/498?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 4 others)
* **#558** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/558?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#557** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/557?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=484).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/bridge/src/cctp/mock_attestation.rs
0xgleb: over-clippied

flake.nix
```diff
@@ -251,7 +251,8 @@
             additionalBuildInputs = [ ragenix.packages.${system}.default ];
             body = ''
               ${infraPkgs.parseIdentity}
-              ragenix --rules ./secret/secrets.nix -i "$identity" -e "$@" && exec ${rekeySecrets}
+              ragenix --rules ./secret/secrets.nix -i "$identity" -e "$@"
+              exec ${rekeySecrets}
             '';
```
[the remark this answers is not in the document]
0xgleb: ${infraPkgs.parseIdentity}

does that not do the trick?

tests/e2e/rebalancing/assertions.rs
```diff
@@ -228,6 +230,7 @@ pub(crate) fn build_usdc_rebalancing_ctx<BP>(
     equity_tokens: &[(String, Address, Address)],
     usdc_vault_id: B256,
     cctp: CctpOverrides,
+    reserved: Option<Positive<Usd>>,
```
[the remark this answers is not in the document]
0xgleb: double check. might be done already

src/config.rs
```diff
@@ -69,6 +69,9 @@ pub struct CashAssetConfig {
     pub vault_id: Option<B256>,
     pub rebalancing: OperationMode,
     pub operational_limit: Option<Positive<Usdc>>,
+    /// USD amount subtracted from offchain cash to compute available balance.
+    /// Prevents the system from rebalancing funds that should remain untouched.
+    pub reserved: Option<Positive<Usd>>,
```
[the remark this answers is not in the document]
0xgleb: the comment is stale, coverage seems sufficient now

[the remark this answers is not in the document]
0xgleb: check if this comment is stale

0xgleb:
@coderabbitai review

src/queue.rs
```diff
@@ -41,15 +40,50 @@ pub(crate) struct QueuedEvent {
...
+            log_index,
+            block_number,
+            event,
+            processed: false,
+            created_at: None,
+            processed_at: None,
+            block_timestamp,
+        })
+    }
+}
```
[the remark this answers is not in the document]
0xgleb: not needed

dashboard/src/lib/env.ts
```diff
@@ -24,9 +24,7 @@ export const getWebSocketUrl = (): string => {
   return val !== undefined && val !== '' ? val : getDefaultWsUrl()
 }
 
-export const getApiBaseUrl = (): string => {
-  return window.location.origin
-}
+export const getApiBaseUrl = (): string => window.location.origin
```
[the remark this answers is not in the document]
0xgleb: not needed

0xgleb:
@coderabbitai review

JuaniRios (changes requested):
We should carry both the gross and available balance through the snapshot instead of stripping out the reserve before it hits the inventory. Right now the reserve is invisible downstream — the dashboard just shows a smaller cash number with no explanation. If we keep both, rebalancing still uses the available amount, but operators can actually see the breakdown and won't wonder why the numbers don't match Alpaca.

0xgleb:
> We should carry both the gross and available balance through the snapshot instead of stripping out the reserve before it hits the inventory. Right now the reserve is invisible downstream — the dashboard just shows a smaller cash number with no explanation. If we keep both, rebalancing still uses the available amount, but operators can actually see the breakdown and won't wonder why the numbers don't match Alpaca.

​great point. sorted

src/inventory/polling.rs
```diff
@@ -508,6 +528,26 @@ where
...
+fn compute_available_cash(
+    broker_usd_balance_cents: i64,
+    reserved_cash: Usd,
+) -> Result<(i64, i64), ReserveError> {
+    let gross =
+        Usd::from_cents(broker_usd_balance_cents).ok_or(ReserveError::BrokerCentsConversion {
+            cents: broker_usd_balance_cents,
+        })?;
+    let available = (gross - reserved_cash)?;
+    let available = st0x_finance::NonNegative::new(available)?;
```
JuaniRios, line 562: `NonNegative::new(available)?` means that if the broker ever reports a negative cash balance (common during margin calls, intraday settlement, or borrowed-funds scenarios), this function returns an error and `poll_offchain` fails every cycle. The USD balance becomes permanently stale until the broker balance goes positive again, and the logs fill with errors.

Previously negative balances were stored and propagated downstream, which let rebalancing see the real state. With `reserved_cash = ZERO`, a broker reporting -$50 now hits this path and breaks.

If this is intentional (i.e., margin accounts are explicitly not supported), it should be documented. Otherwise, consider clamping to zero with a warning log so polling degrades gracefully rather than stalling entirely.

src/inventory/snapshot.rs
```diff
@@ -401,6 +417,8 @@ pub(crate) enum InventorySnapshotEvent {
     OffchainUsd {
         #[serde(alias = "cash_balance_cents")]
         usd_balance_cents: i64,
+        #[serde(default)]
+        gross_usd_cents: i64,
```
JuaniRios: This should be `Option<i64>` rather than `i64` with `#[serde(default)]`.

Here's what happens on upgrade: old `OffchainUsd` events in the event store don't have a `gross_usd_cents` field. Serde sees the missing key, applies `#[serde(default)]`, and sets it to `0`. The view then does `offchain_gross_usd_cents: Some(*gross_usd_cents)` which becomes `Some(0)`. The DTO builder calls `Usdc::from_cents(0)` which returns `Some(Usdc::ZERO)`. The dashboard receives `offchainGross: "0"` and renders "Gross: $0.00" for about 15 seconds until the first live poll replaces it with the real value.

With `Option<i64>`, old events deserialize to `None`, the view skips setting the gross field, and the dashboard hides the row entirely for legacy data. The command struct stays `i64` since the value is always known at write time:

```rust
// Event
#[serde(default)]
gross_usd_cents: Option<i64>,

// View apply
offchain_gross_usd_cents: *gross_usd_cents,  // None propagates naturally
```

The schema version bump to 5 clears snapshots (good), but the events themselves are still replayed from the log, so this matters for any deployment that has accumulated history.

Interestingly, `docs/cqrs.md` added in this very PR calls out this exact pitfall: use `Option<T>` with `#[serde(default)]` when you need to distinguish "field not present" from "field was zero." This is exactly that case.

dashboard/src/lib/components/available-inventory.svelte
```diff
@@ -278,6 +278,14 @@
         <Table.Cell class="font-mono font-medium">Cash</Table.Cell>
         <Table.Cell class="text-right font-mono opacity-90">
           <span class={approxClass(cashRow.alpaca)} title={cashRow.alpaca.truncated ? cashRow.alpaca.full : undefined}>{cashRow.alpaca.display}</span>
+          {#if usdc?.offchainGross}
```
JuaniRios, line 281: This row shows up for every deployment, even ones without a reserve configured. Since `compute_available_cash` runs unconditionally, `gross_usd_cents` is always populated (equal to `usd_balance_cents` when reserve is zero). The guard `{#if usdc?.offchainGross}` is truthy for any non-empty string including `"0"`, so operators without a reserve always see "Gross: $X · Rsv: $0".

If the event field becomes `Option<i64>` (see snapshot.rs comment), only emit `Some(gross)` when `reserved_cash > Usd::ZERO`, so `offchainGross` is `null` at the DTO level for no-reserve deployments. Alternative: change the guard to `{#if usdc?.offchainGross && usdc.offchainGross !== usdc.offchainAvailable}`.

src/inventory/polling.rs
```diff
@@ -83,11 +100,11 @@ where
...
-    orderbook: Address,
-    order_owner: Address,
+    snapshot_id: InventorySnapshotId,
     snapshot: Arc<Store<InventorySnapshot>>,
     wallet_polling: Option<WalletPollingCtx>,
     tokenizer: Option<Arc<dyn Tokenizer>>,
+    reserved_cash: Usd,
 }
 
 impl<Chain, Exe> InventoryPollingService<Chain, Exe>
```
JuaniRios, line 109: A user could configure `reserved = 50.005` and the system starts fine (it's a valid `Positive<Usd>`). But at poll time, `(gross - reserved_cash)` produces a sub-cent result, and `to_cents()` returns `Err(SubCentPrecision(...))` every cycle, permanently blocking USD balance updates. The failure is deferred from startup to runtime.

Validate `reserved_cash.to_cents()` here in the constructor so sub-cent values are rejected immediately at startup with a clear error message.

src/inventory/polling.rs
```diff
@@ -65,6 +66,22 @@ pub(crate) enum InventoryPollingError<ExecutorError> {
...
+}
+
+/// Errors that can occur when computing available cash after subtracting
+/// the configured reserve from the gross broker balance.
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum ReserveError {
+    #[error("gross balance is less than reserve: {0}")]
+    NotNonNegative(#[from] NotNonNegative<Usd>),
+    #[error(transparent)]
+    Arithmetic(#[from] FloatError),
```
JuaniRios, line 80: The error message says "gross balance is less than reserve" but it also fires when `reserved_cash` is zero and the broker itself reports a negative balance. In that case the message is confusing because the reserve isn't the problem at all.

Consider including both values: `"available cash is negative (gross: {gross}, reserved: {reserved})"` so operators can immediately see whether the issue is a large reserve or a negative broker balance.

tests/e2e/rebalancing/assertions.rs
```diff
@@ -1315,6 +1319,59 @@ pub(crate) async fn assert_base_wallet_usdc_event_exists(
...
+    broker: &AlpacaBrokerMock,
+    reserved: Positive<Usd>,
+) -> anyhow::Result<()> {
+    let reserved_cents = reserved.inner().to_cents()?;
+
+    let timeout = Duration::from_secs(DEFAULT_POLL_TIMEOUT_SECS);
+    let deadline = tokio::time::Instant::now() + timeout;
+    let context = "OffchainUsd snapshot matching broker cash minus reserve".to_string();
+
+    loop {
```
JuaniRios, line 1307: This assertion verifies `usd_balance_cents == broker_cash - reserved` but never checks that `gross_usd_cents == broker_cash`. If the values were accidentally swapped in the event construction (gross and available), the test would still pass. Worth adding an assertion on `gross_usd_cents` too, since that's the main new invariant of this feature.

src/integration_tests/rebalancing.rs
```diff
@@ -1072,7 +1074,131 @@ async fn usdc_onchain_imbalance_triggers_base_to_alpaca() {
...
+/// With $300 reserve: effective offchain = 200, ratio = 500/700 = 71.4%
+/// -> above 70% upper bound -> TooMuchOnchain -> base_to_alpaca.
+/// Excess = 500 - 350 (target = 0.5 * 700) = $150.
+#[tokio::test]
+async fn cash_reserve_shifts_offchain_balance_triggering_base_to_alpaca() {
+    let pool = setup_test_db().await;
+
+    let (event_sender, _) = broadcast::channel::<Statement>(16);
+    let inventory = Arc::new(BroadcastingInventory::new(
+        InventoryView::default(),
```
JuaniRios, line 1112: The docstring says "Simulates: broker reports $500, reserve is $300, so polling sends $200" but the test manually builds inventory with `offchain: Usdc::new(float!(200))` directly. No reserve is configured on the trigger and no polling subtraction happens here. The test correctly validates that a 500/200 imbalance triggers rebalancing, but the reserve feature itself isn't exercised at this integration level (it is covered by the unit test `reserved_cash_subtracted_from_offchain_usd` and the e2e test).

Either clarify the docstring to say it tests the rebalancing outcome given a pre-computed available balance, or wire up an actual reserve config to make the test truly end-to-end for this feature.

src/inventory/polling.rs
```diff
@@ -231,14 +245,16 @@ where
...
+        let raw_balance = self
             .raindex_service
             .get_usdc_balance::<OpenChainErrorRegistry>(
-                self.order_owner,
+                self.snapshot_id.owner,
                 RaindexVaultId(usdc_vault.vault_id),
             )
             .await?;
 
+        let usdc_balance = raw_balance;
```
JuaniRios: nit: `let usdc_balance = raw_balance;` is a leftover noop from the refactoring. The offchain path has a meaningful rename (raw -> available after subtraction) but the onchain path just aliases the same value. Can inline it.

JuaniRios (changes requested):

JuaniRios (approved):

findolor (approved):

PR #562, opened 2026-04-10
fix: check offchain inventory before placing counter trades

findolor:
## Motivation

Closes #263.
Closes [RAI-45](https://linear.app/makeitrain/issue/RAI-45/check-offchain-inventory-before-placing-counter-trades).

Counter-trades were being submitted to Alpaca without proving that the broker
account could actually support them.

That left three concrete gaps:

- sell-side counter-trades could be placed without confirming the broker held
enough shares
- buy-side counter-trades could be preflighted against raw cash instead of
margin-safe buying power
- concurrent submission paths could approve multiple counter-trades against the
same broker capacity before pending orders were recorded

The result was that the bot could still oversubscribe inventory or rely on
margin even though the goal of the issue was explicitly to avoid that.

## Solution

- add Alpaca counter-trade preflight for both broker integrations
- require sell-side preflight to confirm sufficient offchain equity inventory
- require buy-side preflight to use `non_marginable_buying_power`, capped by
available cash, instead of raw cash
- buffer buy-side cost estimation with configurable
`broker.counter_trade_slippage_bps`
- return explicit preflight reservation data so the conductor can reason about
approved broker capacity
- serialize counter-trade submission through a shared in-process lock so the
preflight-to-persist step cannot race across queued and periodic submission
paths
- move the counter-trade gating logic into a dedicated
`src/conductor/counter_trade.rs` module instead of keeping it spread across
the conductor
- remove conductor-side double-counting of already-open Alpaca buy orders once
broker buying power is trusted as the live source of truth
- make USDC rebalance amounts quantize to 6 decimals at the trigger boundary so
execution keeps rejecting lossy financial input instead of silently truncating
it

## Checks

By submitting this for review, I'm confirming I've done the following:

- [x] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs
- [ ] included screenshots (if this involves a change to the dashboard)

Verification run:

- [x] `nix develop -c cargo check --workspace`
- [x] `nix develop -c cargo nextest run --workspace --all-features`
- [x] `nix develop -c cargo clippy --workspace --all-targets --all-features`
- [x] `nix develop -c cargo fmt --all`

## Anything else

`nextest` finished green, but it reported one flaky retry in the existing e2e
test `hedging::direct_high_precision_sell_price_still_hedges`, which passed on
retry.

This fix is intended for the current single-process deployment model. If we ever
run multiple bot processes against the same broker account, the next step would
[...]

PR #570, opened 2026-04-13
remove: schwab and alpaca trading api

findolor:
## Motivation

This PR removes the dormant Schwab and Alpaca Trading API integrations so the codebase matches the set of brokers we actually support and maintain.

Keeping those unused paths around was adding maintenance cost across the execution crate, runtime config, CLI, API surface, dashboard wiring, and documentation. It also made the counter-trade work in #562 harder to reason about because that branch only needs to operate on the current supported broker path.

## Solution

- remove the Schwab and Alpaca Trading API executor implementations from `st0x-execution`
- remove their runtime wiring from config parsing, executor dispatch, CLI commands, and the HTTP API
- remove the Schwab auth flow and broker branching from the dashboard/backend paths
- update tests that still assumed Schwab or Alpaca Trading API behavior so the active runtime model is now `alpaca-broker-api` plus `dry-run`
- remove the last legacy `SupportedExecutor::{Schwab, AlpacaTradingApi}` compatibility seam so the executor model only contains supported runtime variants
- update `README.md`, `SPEC.md`, `docs/domain.md`, and execution docs so they describe only the supported system
- make this the parent cleanup for #562 so the counter-trade preflight work sits on top of the reduced broker surface

## Checks

- [x] added test coverage or updated existing coverage for changed logic
- [x] kept the PR focused on removing unsupported broker paths
- [x] linked the dependent PR that now stacks on top of this one
- [ ] included screenshots (if this involves a change to the dashboard)

Verification run:

- [x] `cargo check --workspace`
- [x] `cargo nextest run -p st0x-hedge`
- [x] `cargo nextest run --workspace --all-features`
- [x] `cargo clippy --workspace --all-targets --all-features`
- [x] `cargo fmt --all --check`
- [x] `bun install` in `dashboard/`
- [x] `bun x vitest run src/lib/env.test.ts src/lib/stores/broker.svelte.test.ts`

## Anything else

This PR is the intended parent for #562:
- child PR: #562 `fix: check offchain inventory before placing counter trades`

After the latest cleanup, this branch removes both unsupported runtime integrations and the remaining legacy executor compatibility layer, so the broker model now reflects only the supported system.


findolor:
* **#562** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/562?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#570** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/570?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/570?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#576** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/576?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=570).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

JuaniRios (changes requested):
### High priority

1. **Dead Schwab credential tables on fresh installs.** `migrations/20250703115746_trades.sql:81` still creates `schwab_auth`, and `migrations/20251105152853_schwab_auth_view.sql:1` still creates `schwab_auth_view`. The legacy cleanup migration at `20260217103751_drop_legacy_tables.sql` drops `offchain_trades` and friends but forgot these two. The original migration flagged `schwab_auth` as storing sensitive OAuth tokens, so leaving it behind on new deployments is worth fixing. Suggest `sqlx migrate add drop_schwab_auth_tables` with `DROP TABLE IF EXISTS schwab_auth; DROP TABLE IF EXISTS schwab_auth_view;`.

2. **`supports_fractional_shares` is now a constant-true method.** Both remaining executors (`AlpacaBrokerApi` and `DryRun`) return `true`, so the `else` branch in `Position::is_ready_for_execution` at `src/position.rs:396` that floors to whole shares can never run. Worth deleting the method in `crates/execution/src/lib.rs:251`, collapsing the `if/else` so `capped_shares` is returned directly, and merging the two rewritten tests that now exercise the same path.

3. **`test_run_function_invalid_orderbook_address` was deleted without replacement** in `src/lib.rs`. The sibling error-propagation test was adapted, but this one was removed entirely. If the intent is that both collapsed to the same error path once Schwab init went away, can you either keep it with a distinct assertion (for example, use a valid websocket URL so a different error surfaces) or leave a short comment explaining the consolidation? As is, it just looks like silent coverage loss.

4. **Six generic CLI validation tests were deleted without replacements in `src/cli/mod.rs`.** Most of the CLI test deletions are fine because they tested Schwab-specific flows that are gone, but these six covered generic behavior that still applies:
   - `test_sell_command_rejects_zero_quantity` (the buy version was kept)
   - `test_wrap_equity_command_rejects_zero_quantity`
   - `test_unwrap_equity_command_rejects_zero_quantity`
   - `test_sell_command_parses_fractional_quantity` (buy version kept)
   - `test_wrap_equity_command_parses_fractional_quantity`
   - `test_parse_positive_shares_rejects_zero`

   Can we restore these? None are Schwab or Alpaca Trading API specific.

5. **SPEC.md still documents the removed auth surface.** `SPEC.md:2193`, `:2296`, `:2424`, `:2434`, and `:2570` still mention "auth status" in the initial WebSocket state, list "auth" as an HTTP mutation, and talk about executor maintenance refreshing auth tokens. `/auth/refresh` and the CLI auth command are both gone now. The spec should be updated in the same PR that removes the behavior.

6. **Root `AGENTS.md` still references Schwab and removed commands.** Line 177 says "Charles Schwab or Alpaca Markets", line 204 instructs `cargo run --bin cli -- auth`, line 359 lists `SchwabExecutor` and `AlpacaTradingApi` as supported, line 402 references `SchwabError`, and line 519 frames HTTP mocking around Schwab. New contributors will try to follow workflows that no longer exist.

### Medium priority

7. **Grafana datasource still points at Schwab.** `grafana-datasource.yaml:4-10` has a "Schwab SQLite" datasource pointing at `/databases/schwab.db`, marked `isDefault: true`. That database is never created anymore, so Grafana will show a broken default datasource. Suggest removing the block and promoting the Alpaca SQLite datasource to default.

8. **`PUBLIC_SCHWAB_WS_URL` is a dead config knob.** `dashboard/.env.example:6` still advertises it, but `dashboard/src/lib/env.ts` only reads `PUBLIC_ALPACA_WS_URL` now. The comment on line 4 ("WebSocket URLs for each broker backend") is also stale with only one backend. Worth deleting the line and simplifying the comment.

9. **`authStatus` is dead API surface.** `crates/dto/src/lib.rs:43` still exposes `auth_status` and `:480` still defines `AuthStatus`, always defaulting to `NotConfigured`. The dashboard still hydrates a `['auth']` query from it at `dashboard/src/lib/websocket.svelte.ts:42` and `:113`, and there is a test at `dashboard/src/lib/websocket.svelte.test.ts:122` and `:175`. With `/auth/refresh` and the CLI auth flow gone, nothing produces it anymore. Suggest removing the field, deleting the dashboard hydration and test, and regenerating the TS bindings so `dashboard/src/lib/api/InitialState.ts` updates.

10. **Orphaned Schwab OpenAPI spec file.** `account_orders_openapi.yaml` is a 400 plus line Schwab Trader API spec that nothing references now. Probably kept for schema lookup during development. Safe to delete.

11. **`TradeAccountingError` still couples to `AlpacaBrokerApiError`.** The old enum had a TODO comment right above the Alpaca variants saying it should not be coupled to a concrete executor. This PR removed the TODO and the `AlpacaTradingApi` variant, but left `AlpacaBrokerApi(#[from] AlpacaBrokerApiError)` at `src/trading/onchain/trade_accountant.rs:193-194` intact. The architectural warning was erased while the violation still stands. Resolving it properly is non-trivial because `ExecutionError` lost its broker-specific variant in this same PR and no longer acts as a catch-all. At minimum, please restore the TODO so the next person knows the coupling is known-wrong.

### Low priority

12. **Single-option broker selector in the dashboard.** Now that `VALID_BROKERS = ['alpaca']` in `dashboard/src/lib/env.ts:4`, the `<Select>` dropdown in `dashboard/src/lib/components/header-bar.svelte:41` has one item whose `onValueChange` ignores the chosen value. `defaultBroker` at `dashboard/src/lib/env.ts:10` has a fallback branch that can never trigger. `brokerStore` at `dashboard/src/lib/stores/broker.svelte.ts` is tautological, and its test just sets `'alpaca'` to `'alpaca'`. Worth removing the selector, the store, `VALID_BROKERS`, `defaultBroker`, and the `broker` parameter of `getWebSocketUrl`.

13. **`run_conductor_loop` no longer loops.** At `src/lib.rs:198`, the function used to retry on Schwab refresh token expiration. It now runs `dispatch_to_executor` once and returns. The name is misleading and the docstring ("Retry loop that initializes the broker executor...") is outright wrong. While you are in there, worth rechecking whether `#[allow(clippy::cognitive_complexity)]` on `dispatch_to_executor` is still earned after two match arms were removed.

14. **`DRY_RUN_MIN_SHARES` uses `unreachable!()`.** In `src/config.rs`, the `unwrap_or_else(|_| unreachable!())` is a hidden panic in production. It pre-existed as `SCHWAB_MIN_SHARES` so it is not a regression, but this PR renames and moves it, so it is a natural spot to switch to a construction that returns a `Result` or uses a const-safe path.

15. **`/api/alpaca/ws` route path is vestigial.** `dashboard/nginx.conf:10` and `os.nix:150` both still use the `/api/<broker>/ws` scheme. With only one broker, the broker segment does not do anything. Not broken, just dead weight. Could collapse to `/api/ws`, or leave it with a comment saying it is kept for future multi-broker expansion.

16. **Stale `.sqlx` query cache files.** Three files under `.sqlx/` still describe `SELECT ... FROM schwab_auth` queries. They do not break the build, but `cargo sqlx prepare --check` in CI would complain. A `cargo sqlx prepare` and commit would clean them up.

JuaniRios (approved):

hardyjosh (approved):

PR #571, opened 2026-04-13
feat: restore tailnet-only CI deploys + split tailscale secrets

JuaniRios:
## What

Related: [RAI-186](https://linear.app/makeitrain/issue/RAI-186/server-hardening-do-firewall-tailscale-vpn-fail2ban) (original hardening PR that this restores).

Restores the tailnet-only posture that #556 implemented and #558 silently regressed, closes the public SSH port that was kept open as a temporary bootstrap escape hatch, and splits the shared Tailscale auth key secret into per-env files.

- Per-env Tailscale MagicDNS hostnames (`keys.nix`)
- CI deploy workflows join the tailnet and resolve the droplet via MagicDNS (`deploy-staging.yaml`, `deploy-prod.yaml`)
- Local tooling (`remote`, `bot-*`, `status`, `dashboard`) uses per-env MagicDNS for staging too, not just prod (`infra/default.nix`)
- DO Cloud Firewall drops the public port-22 rule; only 41641/UDP (Tailscale WireGuard) remains inbound (`infra/modules/stack/main.tf`)
- NixOS firewall drops `22` and `80` from `allowedTCPPorts` — both were dead config once the tailnet is the only ingress path (`os.nix`)
- **New:** `tailscale-authkey.age` split into `tailscale-authkey-prod.age` + `tailscale-authkey-staging.age`, each encrypted only to its own env's service recipients. `os.nix` templates `services.tailscale.authKeyFile` and the `age.secrets` entry on `${environment}` so each droplet only decrypts its own secret.

## Why

- **#556 shipped CI-over-Tailscale in `ci.yaml`** as part of "server hardening — firewall, Tailscale VPN, fail2ban" (commit `dba72f4`). The goal: everything inbound to the droplet goes through the tailnet; public 22/80 get blocked.
- **#558 ("separate staging and production deployment workflows", commit `2d91283c`) split the deploy job out of `ci.yaml`** into fresh `deploy-staging.yaml` and `deploy-prod.yaml` files. Those files were written from scratch and did not carry over the `tailscale/github-action` step or MagicDNS resolution — they went back to the pre-#556 pattern of raw `SSH_KEY` + public-IP resolution. The PR description even says *"Both deploy workflows use the same SSH_KEY secret"*, showing Tailscale was not in mind during the split.
[...]

PR #568, opened 2026-04-11
fix: robust status command and readonly database crash loop

JuaniRios:
## What

Closes [RAI-167](https://linear.app/makeitrain/issue/RAI-167/fix-prod-dashboard-nix-command)
Related: [RAI-152](https://linear.app/makeitrain/issue/RAI-152/create-simple-way-to-get-prod-bot-status)

- Rewrite `status.sh` and `dashboard.sh` scripts from Bash to Nushell for better error handling and native TOML/JSON parsing
- Fix the staging bot's readonly database crash loop by migrating from `DynamicUser` to a static `st0x` system user
- Fix the status command to always complete and download logs/DB even when the bot is stopped or the database is uninitialized

## Why

- **Crash loop**: The staging bot was in a restart loop (2400+ restarts) because `DynamicUser=true` allocates an ephemeral UID per run. After a restart cycle, systemd allocated a different UID that couldn't write to the existing SQLite database at `/mnt/data/st0x-hedge.db`, causing `(code: 8) attempt to write a readonly database` on every startup
- **Status command fragility**: The old Bash status script used `set -euo pipefail`, which meant any failed SSH/DB query would abort the entire script before downloading logs and the database — the most important artifacts for debugging
- **Nushell benefits**: Native TOML/JSON parsing replaces fragile `jq`/`awk` pipelines; structured error handling via `try/catch` per section

### NixOS service (`os.nix`)
- Replace `DynamicUser=true` + `SupplementaryGroups` with static `User=st0x` / `Group=st0x`
- Add `users.users.st0x` as a system user (the `st0x` group already existed)
- Add tmpfiles rule to ensure `/mnt/data` is owned by `st0x:st0x` with `0755` permissions on boot
- Remove `ReadWritePaths` (no longer needed — the user directly owns the data directory)

### Deploy script (`deploy.nix`)
- Add `chown st0x:st0x /mnt/data/*.db*` during deploy to fix ownership of existing DB files created by the old dynamic user
- Wrapped in a subshell `(... || true)` so the `|| true` (needed because glob may not match on fresh deploys) doesn't mask failures in the preceding `&&`-chain steps

### Status script (`scripts/status.nu`)
- Each display section (Deployed Version, Active Config, Raindex Orders, DB queries, etc.) wrapped in independent `try/catch` so failures are isolated
- Footer conditionally reports which files were actually saved vs failed to download
- Fixed `get -i` -> `get -o` for nushell deprecation, and `\[`/`\]` invalid escape sequences

[...]

PR #567, opened 2026-04-11
fix: include cutoff block in backfill to prevent data loss

JuaniRios:
## What

Closes [RAI-196](https://linear.app/makeitrain/issue/RAI-196/block-n-lost-between-backfill-and-live-stream)

Fixes a bug where events in the cutoff block are silently dropped during the backfill-to-live-stream handoff in `Conductor::run`.

## Why

- `get_cutoff_block` **consumes** one event from the WS stream to learn the current block number `n`
- Backfill then ran up to block `n-1` (`cutoff_block.checked_sub(1)`)
- The live stream resumes at block `n+1` (since the event at `n` was already consumed)
- **Block `n` is never processed by either path** — any order fills in that block are silently lost, causing missed hedges and stale position data

## How

- Remove the `checked_sub(1)` and pass `cutoff_block` directly to `backfill_events` so backfill includes block `n`
- The consumed stream event and the backfilled events from block `n` may overlap, but this is harmless — `OnChainTrade` aggregates already deduplicate by `(tx_hash, log_index)`, logging "Trade already processed (duplicate event), skipping"
- `backfill_events` already handles `start_block > end_block` gracefully (skips with a log), so no guard needed

## Testing

- All 32 conductor tests pass
- The existing `test_get_cutoff_block_with_timeout` test verifies the timeout fallback path
- The deduplication path is covered by existing `OnChainTrade` aggregate tests

## Screenshots

N/A

## Anything else

- The issue suggested three possible fixes: (1) backfill through block `n` inclusive, (2) use `provider.get_block_number()` instead of consuming the stream, (3) redesign with simultaneous backfill + WS dedup. Option 1 was chosen as the simplest correct fix — it's a one-line change with no architectural risk
- The timeout path (`provider.get_block_number()`) was already correct — no stream event is consumed, so no gap exists

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed event backfilling during startup to include the cutoff block, ensuring complete and uninterrupted historical event recovery during system initialization without gaps.

PR #566, opened 2026-04-11
fix: make service config editable without redeploy

JuaniRios:
## What

Closes [RAI-61](https://linear.app/makeitrain/issue/RAI-61/editable-service-config-without-redeploy)
Closes [#398](https://github.com/ST0x-Technology/st0x.liquidity/issues/398)

Changes the deploy activation script to **copy** the config file instead of **symlinking** it to the Nix store, making it a regular mutable file that operators can edit on the server.

## Why

- Config is currently a symlink to the Nix store (`ln -sf ${configFile} ${cfg.configPath}`), which is immutable — you literally cannot edit it on the server
- Changing any config value (e.g. enabling trading for a symbol) requires a full NixOS rebuild + redeploy
- For quick prototyping on staging, operators need to stop the bot, edit the TOML, and restart without going through the full deploy cycle

## How

- Replace `ln -sf` with `cp` in `deploy.nix`'s `mkServiceProfile` activation script
- Set `root:st0x` ownership and `0640` permissions on the copied file, matching the pattern already used for decrypted secrets
- The service runs with `SupplementaryGroups = [ "st0x" ]` so it can read the file

**No other changes needed** — the Rust binary reads whatever file path `--config` points to, and `services.nix`/`os.nix` are unchanged.

**Workflow:**
1. Deploy copies config from repo to `/run/st0x/st0x-hedge.config` (as before, but now a regular file)
2. Operator edits the file directly on the server
3. Operator runs `systemctl restart st0x-hedge`
4. Service reads updated config — no redeploy needed
5. Next deploy overwrites with the repo version (operator changes are not persistent across deploys)

## Testing

- Nix linters pass (deadnix, nil, nixfmt-classic, statix)
- Functional verification requires deploying to staging:
  - `file /run/st0x/st0x-hedge.config` should show a regular file, not a symlink
  - Edit config, restart service, confirm new config takes effect

## Screenshots

N/A

## Anything else

- Secrets remain unchanged — still decrypted from `.age` files to `/run/agenix/` on each deploy
- Config changes do NOT persist across deploys — this is intentional. The repo config is the source of truth; mutable editing is for quick iteration, not permanent divergence

## Summary by CodeRabbit

* **Chores**
  * Updated deployment configuration file handling for improved system reliability.

PR #572, opened 2026-04-14
fix: project snapshot events into view regardless of rebalancing

JuaniRios:
## What

- New `InventoryProjection` (`src/inventory/projection.rs`) that folds `InventorySnapshotEvent`s into the in-memory `BroadcastingInventory` view the dashboard WebSocket reads from.
- Extracted the event-fold logic into two shared methods on `InventoryView` — `apply_snapshot_event` (normal) and `force_apply_snapshot_event` (recovery) — so every caller folds events identically.
- Wired the projection into both conductor paths:
  - **Rebalancing-enabled**: `RebalancingTrigger` owns the projection internally. `on_snapshot` applies the event via the projection, then runs threshold checks — a failed apply short-circuits the checks so rebalancing never runs against a stale view. The trigger remains the sole `InventorySnapshot` subscriber in this path.
  - **Rebalancing-disabled**: the projection is registered directly as the sole `InventorySnapshot` subscriber in the `else` branch of `run_conductor_loop`'s snapshot-store construction so the dashboard view updates without a rebalancing context.

## Why

Closes [RAI-206](https://linear.app/makeitrain/issue/RAI-206/dashboard-shows-empty-inventory-when-rebalancing-is-disabled).

The bug surfaced while fixing the empty dashboard on staging (parent PRs #568 and #571). After wiring nginx correctly, the WebSocket handshake succeeded and the dashboard rendered — but every panel showed `—` / empty arrays, even though `staging-status` could print live RKLB and USDC balances from the event store.

**Root cause walkthrough:**

- `InventoryPollingService` polls brokers and chains every minute and sends `InventorySnapshotCommand`s to the `InventorySnapshot` CQRS store. This runs on every deploy regardless of config — confirmed in `journalctl -u st0x-hedge` on staging.
- The aggregate's `initialize` / `transition` in `src/inventory/snapshot.rs` emits `InventorySnapshotEvent::*` and persists them. Events exist in the `events` table — `staging-status` reads them with a raw SQL query (`scripts/status.nu:361`).
- The dashboard WS (`src/dashboard/mod.rs:28`) doesn't read the event store. It reads `BroadcastingInventory` — an `RwLock<InventoryView>` that's initialized at startup with `InventoryView::default()` and only mutates when something calls `inventory.write().await`.
- The only subscriber on the snapshot store that called `inventory.write().await` was `RebalancingTrigger::on_snapshot` — which was only instantiated when `ctx.rebalancing_ctx()` returned `Ok(...)`.
- Staging has `rebalancing = "disabled"` for cash and every equity, so `rebalancing_ctx()` returns `Err(NotRebalancing)`, `spawn_rebalancing_infrastructure` is skipped, the trigger never exists, and the snapshot store was built with zero queries attached. Events flowed through the store untouched, view stayed `default()` forever, dashboard showed empty state.

So the responsibility "project snapshot events into the in-memory view" was bundled inside a class that also does "check thresholds and trigger rebalancing actions", and the latter's gate was killing the former too. The fix decouples the two concerns.

**Why not a periodic task / apalis job**: we already have an event-driven mechanism (`sqlite_cqrs` query dispatch) that runs the instant an event is committed, in the same code path as the write, with zero latency and no new tokio lifecycle to supervise. A periodic task would have a tick interval to tune, would re-derive state from the event store on every tick, and would run out-of-step with the write path.

### 1. Shared apply logic on `InventoryView` (`src/inventory/view.rs`)

Two new methods, consuming `self` and returning a new `InventoryView`, matching the existing `update_equity` / `update_usdc` style:

```rust
pub(crate) fn apply_snapshot_event(
    self,
    event: &InventorySnapshotEvent,
    now: DateTime<Utc>,
) -> Result<Self, InventoryViewError>

pub(crate) fn force_apply_snapshot_event(
    self,
    event: &InventorySnapshotEvent,
    now: DateTime<Utc>,
    reason: Arc<InventoryViewError>,
) -> Result<Self, InventoryViewError>
```

Body is a straight lift of the existing match arms from `RebalancingTrigger::on_snapshot` / `on_snapshot_recovery`. Wallet-read event variants (`EthereumUsdc`, `BaseWalletUsdc`, `AlpacaWalletUsdc`, `BaseWalletUnwrappedEquity`, `BaseWalletWrappedEquity`) remain explicit no-ops on the view — they feed accounting/reconciliation elsewhere, not the dashboard inventory.

`force_apply_snapshot_event` runs against the *existing* view rather than a `default()` reset, so a single failing event cannot wipe unrelated venue/symbol state. `OffchainUsd` and `InflightEquity` intentionally reuse the non-forced conversion: silently inventing a `Usdc` value from an invalid cents payload would corrupt financial state, so the original error resurfaces instead of being masked.

### 2. New `InventoryProjection` (`src/inventory/projection.rs`)

Holds an `Arc<BroadcastingInventory>` and exposes `apply(event)`. On each event it takes the write lock, calls `apply_snapshot_event`, and on `InventoryViewError` falls back to `force_apply_snapshot_event` against the existing view. Also implements `Reactor` so it can be registered as a direct CQRS subscriber in the rebalancing-disabled path.

### 3. `RebalancingTrigger` owns the projection (`src/rebalancing/trigger/mod.rs`)

The trigger gains an internal `projection: Arc<InventoryProjection>` field (constructed from the shared `BroadcastingInventory` inside `new()`). `on_snapshot` is now:

```rust
async fn on_snapshot(
    &self,
    event: InventorySnapshotEvent,
) -> Result<(), RebalancingTriggerError> {
    self.projection.apply(event.clone()).await?;
    self.check_and_trigger_after_snapshot(&event).await
}
```

Gating the checks on `apply`'s return value is load-bearing: reactor errors are otherwise logged and swallowed per-subscriber by `ReactorBridge::dispatch`, so without this gate a projection failure would still let the trigger run against a stale view. The trigger remains subscribed to `InventorySnapshot` via `deps!` — it just owns the fold now instead of doing it inline.

### 4. Wiring — `QueryManifest` (`src/conductor/manifest.rs`)

The manifest no longer has a separate `inventory_projection` field — the trigger owns it internally, so the trigger is the sole subscriber on the snapshot store when rebalancing is enabled:

```rust
let snapshot = StoreBuilder::<InventorySnapshot>::new(pool.clone())
    .with(rebalancing_trigger)
    .build(())
    .await?;
```

### 5. Wiring — non-rebalancing path (`src/conductor.rs`)

In the `else` branch of `run_conductor_loop`'s snapshot-store construction, the projection is built standalone and attached as the sole subscriber:

```rust
let inventory_projection = Arc::new(InventoryProjection::new(inventory.clone()));
let snapshot = StoreBuilder::<InventorySnapshot>::new(pool.clone())
    .with(inventory_projection)
    .build(())
    .await?;
```

That's the single line that fixes the staging dashboard.

## Testing

- **Projection unit tests** (`src/inventory/projection.rs`): `apply_onchain_equity_updates_view`, `apply_offchain_usd_converts_and_updates_view`, `apply_wallet_read_events_are_noops_on_view`, `apply_preserves_unrelated_venue_state`, `force_apply_preserves_unrelated_state_and_overwrites_target` (view-level semantic test for the recovery branch, which is not reachable via a real event since the only possible `InventoryViewError` fails identically in both apply paths), and `snapshot_command_through_store_updates_view_without_rebalancing` (end-to-end dispatch through a store with only the projection attached, mirroring the rebalancing-disabled wiring).
- **Manifest live-dispatch test** (`src/conductor/manifest.rs`): `build_frameworks_dispatches_live_snapshot_events_to_view` builds the manifest, sends an `InventorySnapshotCommand` through the built store, and asserts the shared `BroadcastingInventory` reflects the event via the trigger's internal projection.
- **Pre-existing snapshot-fold tests** in `src/rebalancing/trigger/mod.rs` continue to pass unchanged — the fold logic is bit-identical, just routed through the shared helper.
- **Full workspace quality gates** (`cargo fmt --check`, `cargo check --workspace`, `cargo clippy --workspace --all-targets --all-features`, `cargo nextest run --workspace --all-features`) all pass locally. `cargo nextest run -p st0x-hedge` runs 1201 tests, all pass.

### Post-merge verification on staging

After the staging deploy from the tip of the stack:

- [ ] Dashboard at `http://st0x-liquidity-staging.taildeb04.ts.net/` shows non-empty inventory within one poll cycle (~60s) of startup. Alpaca positions, USDC balances, and any onchain equity should appear in the right-hand panels.
- [ ] `staging-status` and the dashboard show the *same* values for onchain/offchain equity and USDC. If they diverge, one of the two readers is stale.
- [ ] `journalctl -u st0x-hedge | grep "Applied inventory snapshot event"` shows the debug log firing on every poll cycle.

Prod behavior is unchanged (rebalancing is enabled there, so the trigger owns the projection and folds events exactly as before — the projection is not a second writer, it's the same writer moved behind a different field). Still worth eyeballing the prod dashboard after the prod deploy to confirm nothing regressed.

## Anything else

- **No periodic task was added.** The user originally framed the fix as a "periodic task via apalis / TaskSupervisor"; after digging into the CQRS wiring the right shape was "fold events into the view via an event-driven subscriber". The poller remains unchanged.
- **No double-write when rebalancing is enabled.** The trigger owns the projection and calls it exactly once before running threshold checks — only one writer per event.
- **Follow-up on rehydration**: `BroadcastingInventory` still starts from `InventoryView::default()` at boot and only populates after the first poll cycle completes (~60s). `InventorySnapshot` is a non-projected entity, so `StoreBuilder::build()` does not `catch_up` reactor subscribers from historical events — the dashboard therefore shows empty state for up to one poll interval after a restart. Not addressed here; a separate concern that would require loading the latest `InventorySnapshot` aggregate state at startup and folding it into the view. Worth a follow-up PR.

## Summary by CodeRabbit

* **Architecture Updates**
  * Added a unified inventory projection to drive live dashboard inventory updates and route snapshot events through a single, consistent path.
  * Projection now handles automatic recovery on apply failures and ensures setups with rebalancing disabled still receive live snapshots.

* **Tests**
  * Added tests for snapshot dispatch, recovery behavior, wallet no-ops, and multi-venue update sequencing.


JuaniRios:
* **#575** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/575?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#572** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/572?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/572?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#571** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/571?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#568** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/568?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#567** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/567?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#566** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/566?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#565** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/565?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#564** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/564?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#559](https://github.com/ST0x-Technology/st0x.liquidity/pull/559) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/559?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#563** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/563?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#561** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/561?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#560** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/560?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=572).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hardyjosh (approved):

findolor (approved):

JuaniRios:
### Merge activity

* **Apr 14, 5:49 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/572).
* **Apr 14, 6:41 PM UTC**: [Graphite](https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/572) couldn't merge this pull request because a downstack PR #561 failed to merge.
* **Apr 15, 6:16 AM UTC**: @JuaniRios merged this pull request with [Graphite](https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/572).

PR #573, opened 2026-04-14
Handle timed out rebalancing transfers

findolor:
## What

- Closes [RAI-65](https://linear.app/makeitrain/issue/RAI-65/stuck-transfer-blocks-rebalancing-with-no-timeout-or-recovery)
- Handle timed out mint, redemption, and USDC rebalancing transfers in the trigger layer
- Add required `rebalancing.transfer_timeout_secs` config and thread it through trigger construction
- Clear stale in-progress guards and inflight inventory when a transfer exceeds the timeout
- Ignore late events for timed-out aggregates so stale CQRS updates do not re-block rebalancing
- Suppress only stale `InflightEquity` snapshots after timeout cleanup, and automatically accept newer snapshots again
- Prune timeout tombstones and snapshot-suppression markers after a bounded retention window

## Why

A transfer that gets stuck in a non-terminal state can leave rebalancing permanently blocked.

The trigger already uses in-progress guards and inflight inventory to prevent overlapping transfers. When a mint, redemption, or USDC rebalance never reaches a terminal event, that state is never cleared, so future rebalance checks keep seeing work as already in progress and stop indefinitely.

This PR fixes the deadlock without inventing synthetic aggregate failure semantics or pretending we know where the stuck funds ended up.

## How

- Extend trigger-side tracking for mints, redemptions, and USDC rebalances to store stage and `last_progress_at`
- Run timeout sweeps before trigger checks and during snapshot handling
- On timeout:
    - log the aggregate id, stage, and elapsed age
    - remove trigger-local tracking
    - clear the corresponding in-progress guard
    - zero the relevant inflight inventory
    - record a timeout tombstone so late events are ignored
- Add inventory helpers for clearing inflight balances without restoring source/destination balances
- Replace symbol-wide inflight suppression with timestamp-based suppression:
    - suppress `InflightEquity` entries only when `fetched_at <= cleared_at`
    - retire suppression automatically once a newer inflight snapshot arrives
- Keep timeout tombstones bounded by pruning old entries during the same sweep

## Testing

- `nix develop -c cargo check --workspace --all-features`
- `nix develop -c cargo nextest run --workspace --all-features`
- `nix develop -c cargo clippy --workspace --all-targets --all-features`
- `nix develop -c cargo fmt`

Added/updated regression coverage for:

- timed out USDC cleanup and late-event ignore
- timed out redemption cleanup
- stale inflight snapshot suppression after timeout cleanup
- newer inflight snapshot acceptance after suppression
- timeout tombstone pruning

## Screenshots

- None

## Anything else

- This keeps timeout recovery in the trigger layer on purpose. It unblocks future rebalancing but does not attempt automatic capital recovery or restart-persistent timeout tracking.

## Summary by CodeRabbit

- **New Features**
  - Added configurable transfer timeout for rebalancing (default: 30 minutes).
  - Automatic timeout cleanup for stuck mint, redemption, and USDC rebalance operations.
  - Actions to clear inflight balances for equities and USDC.

- **Improvements**
  - Operation tracking records stages and progress timestamps for better observability.
  - Suppression and pruning of stale inflight data to avoid late-event side effects.
  - Config validation to prevent zero transfer-timeout.

- **Tests**
  - Expanded coverage for timeout, cleanup, and late-event suppression behaviors.

- **Documentation**
  - Updated USDC deposit/state-flow documentation to reflect conversion sub-flow and terminal states.

### DIFF
--- a/src/inventory/view.rs
+++ b/src/inventory/view.rs
@@ -841,6 +841,44 @@ impl InventoryView {
         })
     }
 
+    pub(crate) fn clear_equity_inflight(
+        self,
+        symbol: &Symbol,
+        venue: Venue,
+        now: DateTime<Utc>,
+    ) -> Result<Self, InventoryViewError> {
+        let Some(inventory) = self.equities.get(symbol).cloned() else {
+            return Ok(self);
+        };
+
+        let cleared = Inventory::set_inflight(venue, FractionalShares::ZERO)(inventory)?;
+        let cleared = Inventory::with_last_rebalancing(now)(cleared)?;
+
+        let mut equities = self.equities;
+        equities.insert(symbol.clone(), cleared);
+
+        Ok(Self {
+            equities,
+            last_updated: now,
+            usdc: self.usdc,
+        })
+    }
+
+    pub(crate) fn clear_usdc_inflight(
+        self,
+        venue: Venue,
+        now: DateTime<Utc>,
+    ) -> Result<Self, InventoryViewError> {
+        let cleared = Inventory::set_inflight(venue, Usdc::ZERO)(self.usdc)?;
+        let cleared = Inventory::with_last_rebalancing(now)(cleared)?;
+
+        Ok(Self {
+            usdc: cleared,
+            last_updated: now,
+            equities: self.equities,
+        })
+    }
+
     /// Returns the set of symbols that currently have inflight balances
     /// at any venue.
     #[cfg(test)]
--- a/src/rebalancing/trigger/mod.rs
+++ b/src/rebalancing/trigger/mod.rs
@@ -10,13 +10,14 @@ use alloy::providers::RootProvider;
 use alloy::rpc::client::RpcClient;
 use alloy::transports::layers::RetryBackoffLayer;
 use async_trait::async_trait;
-use chrono::Utc;
+use chrono::{DateTime, Utc};
 use serde::Deserialize;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, Ordering};
-use tokio::sync::{RwLock, mpsc};
-use tracing::{debug, info, warn};
+use std::time::Duration;
+use tokio::sync::{Mutex, RwLock, mpsc};
+use tracing::{debug, error, info, warn};
 use url::Url;
 
 use rain_math_float::Float;
@@ -27,11 +28,11 @@ use st0x_finance::Usdc;
 
 use crate::config::AssetsConfig;
 use crate::equity_redemption::{EquityRedemption, EquityRedemptionEvent, RedemptionAggregateId};
-use crate::inventory::projection::{InventoryProjection, InventoryProjectionError};
+use crate::inventory::projection::InventoryProjectionError;
 use crate::inventory::snapshot::{InventorySnapshot, InventorySnapshotEvent};
 use crate::inventory::{
-    BroadcastingInventory, ImbalanceThreshold, Inventory, InventoryViewError, Operator, TransferOp,
-    Venue,
+    BroadcastingInventory, ImbalanceThreshold, Inventory, InventoryView, InventoryViewError,
+    Operator, TransferOp, Venue,
 };
 use crate::position::{Position, PositionEvent};
 use crate::tokenized_equity_mint::{
@@ -87,6 +88,8 @@ pub(crate) enum TokenAddressError {
 pub enum RebalancingCtxError {
     #[error("rebalancing requires alpaca-broker-api broker type")]
     NotAlpacaBroker,
+    #[error("rebalancing transfer_timeout_secs must be non-zero")]
+    ZeroTransferTimeout,
     #[error("invalid wallet config: {0}")]
     WalletConfig(#[from] toml::de::Error),
     #[error(transparent)]
@@ -151,6 +154,7 @@ impl std::fmt::Debug for RebalancingSecrets {
 pub(crate) struct RebalancingConfig {
     pub(crate) equity: ImbalanceThreshold,
     pub(crate) usdc: UsdcRebalancing,
+    pub(crate) transfer_timeout_secs: u64,
     pub(crate) redemption_wallet: Address,
     pub(crate) wallet: toml::Value,
 }
@@ -168,6 +172,7 @@ pub(crate) struct RebalancingConfig {
 pub struct RebalancingCtx {
     pub(crate) equity: ImbalanceThreshold,
     pub(crate) usdc: Option<ImbalanceThreshold>,
+    pub(crate) transfer_timeout: Duration,
     /// Issuer's wallet for tokenized equity redemptions.
     pub(crate) redemption_wallet: Address,
     pub(crate) alpaca_broker_auth: AlpacaBrokerApiCtx,
@@ -243,6 +248,10 @@ impl RebalancingCtx {
         secrets: RebalancingSecrets,
         broker_auth: AlpacaBrokerApiCtx,
     ) -> Result<Self, RebalancingCtxError> {
+        if config.transfer_timeout_secs == 0 {
+            return Err(RebalancingCtxError::ZeroTransferTimeout);
+        }
+
         let WalletKindTag { kind } = WalletKindTag::deserialize(config.wallet.clone())?;
 
         let (base_wallet, ethereum_wallet) = tokio::try_join!(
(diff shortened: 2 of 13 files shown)


findolor:
* **#573** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/573?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/573?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=573).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

hardyjosh (approved):

src/rebalancing/trigger/usdc.rs
```diff
@@ -179,88 +287,147 @@ impl RebalancingTrigger {
...
+            stage
+                .timestamp(event)
+                .ok_or(RebalancingTriggerError::MissingUsdcTrackingContext {
+                    id: id.clone(),
+                    event: UsdcTrackingEvent::Initiated,
+                })?;
+
+        let mut tracking = self.usdc_tracking.write().await;
+
+        if let Some(existing) = tracking.get_mut(id) {
```
JuaniRios, line 470: CRITICAL: `Initiated` no longer starts the inventory transfer. For the `AlpacaToBase` flow, `ConversionInitiated` now creates a `usdc_tracking` entry via `upsert_conversion_tracking`, so when the later `Initiated` event arrives, the `if let Some(existing) = tracking.get_mut(id)` branch returns early and never calls `Inventory::transfer(source_venue, TransferOp::Start, amount)`. Source inflight stays at zero, and any subsequent `ConversionFailed` or terminal settlement will be run against a missing inflight balance. Keep the timeout refresh but still apply the start transfer when the withdrawal leg actually begins.

This should have been caught by a test. Add new integration tests in `src/rebalancing/trigger/mod.rs` (alongside the existing `usdc_*` tests) that drive `on_usdc_rebalance` through the complete lifecycle via `ReactorHarness` and assert inflight is correctly started, tracked, and cleared at each stage: one `AlpacaToBase` flow (`ConversionInitiated` -> `ConversionConfirmed` -> `Initiated` -> `WithdrawalConfirmed` -> `BridgingInitiated` -> `BridgeAttestationReceived` -> `Bridged` -> `DepositInitiated` -> `DepositConfirmed`) and one `BaseToAlpaca` flow (`Initiated` -> `WithdrawalConfirmed` -> `BridgingInitiated` -> `BridgeAttestationReceived` -> `Bridged` -> `DepositInitiated` -> `DepositConfirmed` -> `ConversionInitiated` -> `ConversionConfirmed`). Add failure-path variants that end in `WithdrawalFailed` / `BridgingFailed` / `DepositFailed` / `ConversionFailed` and assert inflight is cancelled. Keep these as new tests rather than extending `usdc_pre_withdrawal_conversion_confirmation_refreshes_timeout_tracking` or `timed_out_usdc_rebalance_clears_guards_and_ignores_late_events`, which are scoped to stage refresh and timeout tombstoning. The existing tests either skip `Initiated` or write tracking directly into the map, so the transfer-start path is never exercised end to end.

src/rebalancing/trigger/mod.rs
```diff
@@ -477,6 +648,373 @@ impl RebalancingTrigger {
...
+        if elapsed < self.config.transfer_timeout {
+            return Ok(None);
+        }
+
+        let mut inventory = self.inventory.write().await;
+        *inventory = inventory.clone().clear_usdc_inflight(now)?;
+        drop(inventory);
+
+        tracking_guard.remove(id);
+        drop(tracking_guard);
```
JuaniRios, line 992: **Tombstone race between cleanup and event handling.** This helper removes the tracking entry and drops the write guard, but the tombstone is only inserted afterwards by the caller in `expire_stuck_usdc_rebalances`. A late event for the same aggregate arriving in that window sees neither a tracking entry nor a tombstone, so `on_usdc_rebalance` falls through the top-of-function tombstone check and processes the event normally, and `upsert_conversion_tracking` can even resurrect the tracking entry after it was just cleared. The tombstone insert then lands a moment later and silently blocks all future events for this id, leaving a ghost tracking row with inflight already zeroed. Same shape exists for mints (line ~825) and redemptions (line ~855). Fix: insert the tombstone in the same write section that removes the tracking entry, so no observer can see the "neither map has it" state. Please also add an integration test in `src/rebalancing/trigger/mod.rs` next to the existing timeout tests: inject a stale `usdc_tracking` entry, call `cleanup_timed_out_usdc_rebalance` directly (not the full `expire_stuck_usdc_rebalances` sweep), then fire a late `ConversionInitiated` via `on_usdc_rebalance` and assert the tracking map is still empty. Add symmetric tests for mint and redemption cleanups.

src/inventory/view.rs
```diff
@@ -840,6 +840,45 @@ impl InventoryView {
         })
     }
 
+    pub(crate) fn clear_equity_inflight(
```
JuaniRios, line 844: **`clear_equity_inflight` zeros both venues.** A mint only records inflight on `Venue::Hedging` and a redemption only on `Venue::MarketMaking`, but this helper unconditionally zeros both. Today the per-symbol `equity_in_progress` guard hides the damage, but any non-zero inflight on the other venue (for example from a polling snapshot) gets silently wiped on every timeout cleanup. Same bug in `clear_usdc_inflight`. Fix: add a `venue: Venue` parameter to both helpers and have each cleanup site pass the venue its operation actually wrote to. `cleanup_timed_out_mint` passes `Venue::Hedging`, `cleanup_timed_out_redemption` passes `Venue::MarketMaking`, and `cleanup_timed_out_usdc_rebalance` passes the direction's source venue via `UsdcRebalanceTracking::source_venue()`.

src/rebalancing/trigger/mod.rs
```diff
@@ -504,49 +1054,54 @@ impl RebalancingTrigger {
...
+                        .clone()
+                        .apply_inflight_snapshot(mints, redemptions, fetched_at, now)
+                } else {
+                    Ok(inventory.clone())
+                }
+            }
+        }?;
 
         *inventory = updated;
         drop(inventory);
```
JuaniRios, line 1129: <p class='graphite__hidden'><i>[Re: line +1119]</i></p>
**Recovery path skips the timeout sweep.** `on_snapshot` runs `expire_stuck_operations` before filtering the snapshot, so aged-out operations get tombstoned and their stale inflight dropped in one pass. `on_snapshot_recovery` only filters, so if the clock has moved past a timeout between when `on_snapshot` ran its sweep and when recovery takes over, recovery will reapply stale inflight for an operation that should have been expired on this cycle.

Fix: add `self.expire_stuck_operations_with_logging().await;` as the first line of `on_snapshot_recovery`, mirroring `on_snapshot`.

The production window is microseconds today because `on_snapshot` always sweeps before recovery can run, so this is latent rather than live. Flagging it because the two handlers now have different contracts for the same job: any future change that moves the sweep, adds another caller of the recovery path, or widens the gap between `on_snapshot` and recovery will turn the race into something reproducible. One line keeps both paths symmetric.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/573?utm_source=unchanged-line-comment">Graphite</a>.</p>

JuaniRios (changes requested):
Left in-line comments on important things to address.
The following are minor, but still worth adding before merging:

1. **`filter_suppressed_inflight_snapshot` mutates state under a read-sounding name.** `src/rebalancing/trigger/mod.rs:737`. The function name suggests a pure filter, but it also takes a write lock and retires expired suppression entries. Split the retirement into its own helper (`retire_stale_suppression` or similar) so the filter stays read-only and the side effect is visible at the call site.
<br>

2. **USDC lifecycle docs contradict the new timeout behavior.** `src/usdc_rebalance.rs` module-level docstring around line 6. The state-flow comment still treats `DepositConfirmed` as the terminal success state, but this PR depends on the post-deposit conversion leg staying timeout-tracked for `BaseToAlpaca`. Update the docstring so operators and future reviewers do not misread which stages can legitimately time out.
<br>
3. **`transfer_timeout_secs` is a new required field with no inline doc.** `example.config.toml:17` and `config/cli.toml:16`. Existing deployments that do not update their toml will fail to start with the `ZeroTransferTimeout` / missing-field error, and the example has no comment explaining the unit or default. Add a short comment ("transfer timeout in seconds, 1800 = 30 minutes") so operators know what they are setting.

JuaniRios (changes requested):
critical comment needs addresssing

JuaniRios (approved):

PR #577, opened 2026-04-15
Resume interrupted tokenization aggregates after restart

findolor:
## What

- Closes [RAI-115](https://linear.app/makeitrain/issue/RAI-115/resume-interrupted-tokenization-aggregates-after-restart)
- Resume non-terminal mint and redemption aggregates during conductor startup
- Restore trigger-side tracking and inflight inventory for interrupted tokenization before rebalancing resumes
- Add end-to-end restart coverage for interrupted mint and redemption flows
- Add a test-only tokenization mock control for holding requests pending long enough to exercise restart recovery deterministically

## Why

When the bot crashes mid-tokenization, the aggregate state is still in the event store, but the in-memory workflow state is gone. On restart, the reactor only sees new events, so interrupted mints and redemptions can remain stuck mid-lifecycle with inflight state never cleared.

That leaves the symbol blocked indefinitely until manual intervention. The correct fix is to discover interrupted aggregates at startup, restore the trigger state they depend on, and resume the real workflow from persisted aggregate state.

## How

- Query the event store for `TokenizedEquityMint` and `EquityRedemption` aggregates whose latest event is non-terminal
- During rebalancing startup:
  - load those aggregates from the CQRS stores
  - repopulate `mint_tracking` / `redemption_tracking`
  - restore the relevant `equity_in_progress` guards
  - restore inflight inventory for interrupted tokenization
- Resume mint workflows from persisted states:
  - `MintAccepted` -> poll Alpaca
  - `TokensReceived` -> verify, wrap, deposit
  - `TokensWrapped` -> deposit
- Resume redemption workflows from persisted states:
  - `WithdrawnFromRaindex` -> unwrap
  - `TokensUnwrapped` -> send
  - `TokensSent` -> poll detection
  - `Pending` -> poll completion
- Add focused tests for interrupted aggregate discovery and trigger-state recovery
- Add restart e2e coverage that shuts the bot down after a non-terminal mint/redemption event and verifies the same aggregate completes after restart

## Testing

- `nix develop -c cargo check --workspace`
- `nix develop -c cargo nextest run --workspace --all-features`
- `nix develop -c cargo clippy --workspace --all-targets --all-features -- -D clippy::all -D warnings`
- `nix develop -c cargo fmt`

Added/updated coverage for:

- interrupted mint aggregate discovery
- interrupted redemption aggregate discovery
- trigger recovery for restored mint/redemption state
- resume-from-accepted mint workflow
- resume-from-sent redemption workflow
[...]

PR #592, opened 2026-04-21
fix: checkpoint backfill and clean up apalis jobs

findolor:
## What

[RAI-220](https://linear.app/makeitrain/issue/RAI-220/checkpoint-backfill-progress-clean-up-apalis-jobs-and-rate-limit)

Closes RAI-220

Persist historical backfill progress, periodically clean up terminal apalis job rows, and
bound service restart loops so crashes do not cause repeated full backfills or unbounded
SQLite growth.

## Why

The conductor always backfilled from `deployment_block` on startup. After a crash or
restart, that meant the service re-scanned the same historical range and re-enqueued
already-seen work, relying on downstream idempotency to absorb duplicates.

Apalis job rows are queue bookkeeping, not audit history. Leaving completed, killed, or
terminally failed rows around lets the queue table grow without bound during normal
operation and repeated restarts.

The deployed service also used an aggressive `Restart=always` loop with a short delay
and no explicit burst limit, which can amplify startup failures into repeated RPC
consumption.

## How

- **Backfill checkpoints** (`src/onchain/backfill.rs`, migration) — added
  `backfill_checkpoints` keyed by orderbook. Backfill now resumes from
  `checkpoint + 1`, falls back to `deployment_block` only when no checkpoint exists,
  and advances the checkpoint only after the full requested range is enqueued
  successfully.
- **Apalis job cleanup** (`src/conductor/job.rs`, `src/conductor.rs`) — added cleanup
  for terminal job rows (`Done`, `Killed`, and exhausted `Failed`) followed by SQLite
  `VACUUM`.
- **Explicit cleanup cadence config** (`src/config.rs`, repo TOMLs) — added required
  `apalis_finished_job_cleanup_interval_secs`. Missing config fails TOML parsing;
  zero fails startup validation. No production default or fallback is used.
- **Conductor lifecycle wiring** (`src/conductor.rs`, `src/conductor/builder.rs`) — the
  cleanup task is spawned with the configured interval, included in
  `wait_for_completion`, and aborted with the rest of the conductor.
- **Deployment restart limits** (`os.nix`) — increased restart delay to 30s and added
  systemd start-limit settings to stop runaway restart loops.
- **Docs/spec** (`SPEC.md`, `docs/conductor.md`, `README.md`) — documented
  checkpointing, cleanup behavior, and the required cleanup interval config.

## Testing

Added coverage for:
- Backfill start block selection with and without checkpoints.
- Checkpoint persistence after successful backfill.
- No checkpoint advancement after partial batch failure.
- Skipping backfill when the checkpoint is already caught up.
- Terminal apalis job cleanup SQL behavior.
- Required/non-zero `apalis_finished_job_cleanup_interval_secs` config validation.
- Cleanup task spawn, abort, completion selection, and failure propagation through
  conductor lifecycle.

Validation run:

- `nix develop -c cargo check --workspace`
- `nix develop -c cargo nextest run --workspace --all-features`
  - 1616 passed, 3 skipped; nextest reported 1 leaky existing `rain-math-float` test
    and exited successfully.
- `nix develop -c cargo clippy --workspace --all-targets --all-features`
- `nix develop -c cargo fmt`
- `git --no-pager diff --check`

## Screenshots

N/A

## Anything else

This intentionally makes `apalis_finished_job_cleanup_interval_secs` required config.
Checked-in config files were updated, but any external operator config must add the
field before deploying this change.

## Summary by CodeRabbit

* **New Features**
  * Backfill now resumes from persisted checkpoints instead of always starting from deployment block, preventing reprocessing of historical data.
  * Automatic periodic cleanup of finished jobs to manage queue storage.

* **Configuration**
  * New required configuration parameter: `apalis_finished_job_cleanup_interval_secs` (must be set to a non-zero value).

* **Documentation**
  * Updated operational constraints and startup sequencing documentation.

* **Infrastructure**
  * Enhanced systemd service restart rate limiting and increased restart delay for improved stability.

### DIFF
--- a/src/conductor.rs
+++ b/src/conductor.rs
@@ -17,7 +17,8 @@ use std::sync::Arc;
 use std::time::Duration;
 use task_supervisor::SupervisorHandle;
 use tokio::sync::{Mutex, broadcast, mpsc};
-use tokio::task::JoinHandle;
+use tokio::task::{JoinError, JoinHandle};
+use tokio::time::MissedTickBehavior;
 use tracing::{debug, error, info, warn};
 
 use st0x_dto::Statement;
@@ -30,6 +31,7 @@ use st0x_execution::{
 
 use crate::alpaca_wallet::AlpacaWalletService;
 use crate::bindings::IOrderBookV6::{self, IOrderBookV6Instance};
+use crate::conductor::job::cleanup_finished_jobs;
 use crate::config::{AssetsConfig, BrokerCtx, Ctx, CtxError};
 use crate::dashboard::Broadcaster;
 use crate::equity_redemption::symbols_with_stuck_redemptions;
@@ -107,6 +109,8 @@ pub(crate) struct Conductor {
     /// Polls wallet balances and onchain state on a timer. Absent when
     /// rebalancing (and therefore wallet context) is not configured.
     inventory_poller: Option<JoinHandle<()>>,
+    /// Periodically removes terminal rows from the persistent job queue.
+    job_cleanup: JoinHandle<()>,
 }
 
 fn base_wallet_unwrapped_equity_token_addresses(ctx: &Ctx) -> HashMap<Symbol, Address> {
@@ -185,6 +189,7 @@ impl Conductor {
         backfill_events(
             &provider,
             &ctx.evm,
+            &pool,
             cutoff_block,
             get_backfill_retry_strat(),
             job_queue.clone(),
@@ -286,6 +291,10 @@ impl Conductor {
         };
 
         let hedge_queue = crate::conductor::job::JobQueue::new(&pool);
+        let job_cleanup = spawn_finished_job_cleanup(
+            pool.clone(),
+            Duration::from_secs(ctx.apalis_finished_job_cleanup_interval_secs),
+        );
 
         let conductor_ctx = builder::ConductorCtx {
             ctx: ctx.clone(),
@@ -307,6 +316,7 @@ impl Conductor {
             .dex_streams(dex_streams)
             .maybe_executor_maintenance(executor_maintenance)
             .maybe_rebalancer(rebalancer)
+            .job_cleanup(job_cleanup)
             .call();
 
         info!("Conductor running");
@@ -318,22 +328,13 @@ impl Conductor {
 
 impl Conductor {
     pub(crate) async fn wait_for_completion(&mut self) -> anyhow::Result<()> {
-        tokio::select! {
-            result = self.supervisor.wait() => {
-                result?;
-                info!("Supervisor exited");
-            }
-            result = &mut self.monitor => {
-                if let Err(join_error) = result
-                    && !join_error.is_cancelled()
-                {
-                    return Err(anyhow::anyhow!("Apalis monitor failed: {join_error}"));
-                }
-                info!("Apalis monitor exited");
-            }
-        }
+        let exit = tokio::select! {
+            result = self.supervisor.wait() => ConductorExit::Supervisor(result.map_err(anyhow::Error::new)),
+            result = &mut self.monitor => ConductorExit::Monitor(result),
+            result = &mut self.job_cleanup => ConductorExit::JobCleanup(result),
+        };
 
-        Ok(())
+        handle_conductor_exit(exit)
     }
 
     pub(crate) fn abort_all(&self) {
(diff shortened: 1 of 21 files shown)


findolor:
* **#577** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/577?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#595** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/595?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#594** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/594?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#593** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/593?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#592** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/592?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/592?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#598** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/598?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#597** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/597?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#596** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/596?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=592).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/conductor/job.rs
```diff
@@ -99,6 +99,24 @@ where
...
+         WHERE status = 'Done' \
+         OR status = 'Killed' \
+         OR (status = 'Failed' AND max_attempts <= attempts)",
+    )
+    .execute(pool)
+    .await?
+    .rows_affected();
+
+    if deleted > 0 {
+        pool.execute("VACUUM").await?;
```
JuaniRios: **Major: VACUUM in WAL mode risks blocking all concurrent DB operations**

SQLite `VACUUM` rebuilds the entire database file and requires an exclusive lock. In WAL mode (the default with multiple connections), this blocks all readers and writers for the duration. The production database has concurrent CQRS writes, apalis job polling, backfill enqueue operations, and inventory projections.

Running VACUUM whenever `deleted > 0` (even for a single row) means periodic exclusive-lock stalls proportional to total DB size, not to the number of deleted rows. Additionally, the first `interval.tick()` fires immediately at startup (tokio behavior), so VACUUM may run during the backfill write window on cold starts.

**Suggested fix:** Either:
- Remove VACUUM entirely (SQLite reuses free pages automatically via WAL auto-checkpoint)
- Use `PRAGMA incremental_vacuum` which is non-blocking in WAL mode
- Use `PRAGMA wal_checkpoint(PASSIVE)` after the DELETE, which never blocks
- If VACUUM is intentionally desired, add a deletion threshold (e.g., `deleted > 100`) and document the tradeoff

src/conductor.rs
```diff
@@ -310,22 +321,13 @@ impl Conductor {
...
-                {
-                    return Err(anyhow::anyhow!("Apalis monitor failed: {join_error}"));
-                }
-                info!("Apalis monitor exited");
-            }
-        }
+        let exit = tokio::select! {
+            result = self.supervisor.wait() => ConductorExit::Supervisor(result.map_err(anyhow::Error::new)),
+            result = &mut self.monitor => ConductorExit::Monitor(result),
+            result = &mut self.job_cleanup => ConductorExit::JobCleanup(result),
```
JuaniRios, line 334: **Major: job_cleanup clean exit causes full conductor shutdown**

`spawn_finished_job_cleanup` contains an infinite loop and should never return `Ok(())`. However, `wait_for_completion` races `job_cleanup` alongside `supervisor` and `monitor`. If the cleanup loop exits cleanly for any reason (accidental `break`, future code change), the conductor interprets this as graceful completion, logs "Job cleanup exited", returns `Ok(())`, and `abort_all()` tears down the entire service.

The test `wait_for_completion_returns_when_job_cleanup_exits` confirms this behavior -- it asserts that a cleanly-exiting cleanup task causes successful conductor completion. This is the wrong invariant: an infinite background task returning `Ok(())` should be treated as an error.

**Suggested fix:** Treat `Ok(())` from the cleanup task as an unexpected exit:
```rust
ConductorExit::JobCleanup(result) => match result {
    Ok(()) => return Err(anyhow::anyhow!("Job cleanup task exited unexpectedly")),
    Err(join_error) if join_error.is_cancelled() => {}
    Err(join_error) => return Err(anyhow::anyhow!("Job cleanup failed: {join_error}")),
}
```

src/onchain/backfill.rs
```diff
@@ -90,6 +90,54 @@ pub(crate) async fn backfill_events<P: Provider + Clone, B: BackoffBuilder + Clo
...
+    pool: &SqlitePool,
+    evm_ctx: &EvmCtx,
+    last_processed_block: u64,
+) -> Result<(), OnChainError> {
+    let last_processed_block = i64::try_from(last_processed_block)?;
+
+    sqlx::query(
+        "INSERT INTO backfill_checkpoints (orderbook, last_processed_block) \
+         VALUES (?, ?) \
+         ON CONFLICT(orderbook) DO UPDATE SET \
```
JuaniRios, line 134: **Minor: No checkpoint monotonicity guard -- checkpoint can regress**

The `ON CONFLICT DO UPDATE` unconditionally overwrites `last_processed_block`. If backfill were ever called with a lower `end_block` than the existing checkpoint (chain reorg, bug in cutoff calculation), the checkpoint would regress, causing previously processed blocks to be re-backfilled. Safe due to CQRS idempotency but weakens the checkpoint's guarantee.

**Suggested fix:**
```sql
ON CONFLICT(orderbook) DO UPDATE SET
  last_processed_block = MAX(excluded.last_processed_block, backfill_checkpoints.last_processed_block),
  updated_at = strftime('%Y-%m-%dT%H:%M:%fZ', 'now')
```

src/onchain/backfill.rs
```diff
@@ -36,18 +35,19 @@ pub(crate) fn get_backfill_retry_strat() -> ExponentialBuilder {
...
     evm_ctx: &EvmCtx,
+    pool: &SqlitePool,
     end_block: u64,
     retry_strategy: B,
     job_queue: DexTradeAccountingJobQueue,
 ) -> Result<(), OnChainError> {
-    let start_block = evm_ctx.deployment_block;
+    let start_block = backfill_start_block(pool, evm_ctx).await?;
 
     if start_block > end_block {
```
JuaniRios, line 52: **Minor: Checkpoint not saved when backfill is skipped (start > end)**

When `start_block > end_block`, the early return skips saving the checkpoint. On repeated restarts where the service is already caught up, this means backfill re-evaluates the same skip condition every time rather than knowing the state was already verified as current. Not a data-loss risk, but the "Skipping backfill" log message repeats on every restart until actual events arrive and the full path executes.

Also worth adding a test for the partial-success scenario: first batch succeeds, second batch fails -> verify checkpoint is NOT saved. The existing `test_backfill_events_fails_partway` only fails on the first batch.

src/conductor.rs
```diff
@@ -344,7 +346,63 @@ impl Conductor {
...
+
+    Ok(())
+}
+
+fn handle_task_exit(result: Result<(), JoinError>, task: &'static str) -> anyhow::Result<()> {
+    if let Err(join_error) = result
+        && !join_error.is_cancelled()
+    {
+        return Err(anyhow::anyhow!("{task} failed: {join_error}"));
     }
```
JuaniRios, line 390: **Minor: handle_task_exit drops JoinError source chain**

`anyhow::anyhow!("{task} failed: {join_error}")` stringifies the `JoinError` rather than preserving it as a source error. Consider:
```rust
return Err(anyhow::Error::new(join_error).context(format!("{task} failed")));
```

src/conductor/job.rs
```diff
@@ -99,6 +99,24 @@ where
     }
 }
 
+pub(crate) async fn cleanup_finished_jobs(pool: &SqlitePool) -> Result<u64, sqlx::Error> {
+    let deleted = sqlx::query(
+        "DELETE FROM Jobs \
```
JuaniRios, line 120: **Minor: String-based status matching couples to apalis internals**

These status strings (`'Done'`, `'Killed'`, `'Failed'`) are apalis internal representations. The tests insert them via raw SQL but don't verify apalis actually produces these exact strings. Consider adding a comment citing the apalis version/source where these are defined, so future upgrades can be checked.

JuaniRios (changes requested):

JuaniRios (approved):

hardyjosh (approved):

PR #645, opened 2026-05-07
refactor: import InventorySnapshotEvent variants in fn bodies

0xgleb:
## What

Adds an `InFlightEquityLocation` enum and an `inflight_equity` map to `InventoryView` to track equity tokens observed at intermediate wallet locations (`BaseWalletUnwrapped` and `BaseWalletWrapped`) between venues. Previously, `BaseWalletUnwrappedEquity` and `BaseWalletWrappedEquity` snapshot events were silently ignored (treated as no-ops) in both `InventoryView` and `RebalancingTrigger`. They now populate the new `inflight_equity` map.

## Why

Equity tokens in transit between venues — unwrapped issuer tokens awaiting wrapping, or vault shares awaiting deposit into Raindex — were invisible to the inventory view. Recording them at their intermediate wallet locations gives observability into capital that has left one venue but has not yet arrived at the other, without affecting imbalance math.

## How

`InFlightEquityLocation` mirrors the existing `InFlightCashLocation` pattern. The `inflight_equity` field is a `HashMap<(Symbol, InFlightEquityLocation), FractionalShares>` keyed by symbol and location. `set_inflight_equity_at_location` replaces all entries for a given location with the latest wallet-read balances, making the chain reading authoritative — symbols absent from the new snapshot drop out of the map entirely.

The `inflight_equity` map is intentionally excluded from imbalance math; it is read-only visibility data. Both `apply_snapshot_event` and `force_apply_snapshot_event` now route `BaseWalletUnwrappedEquity` and `BaseWalletWrappedEquity` through `set_inflight_equity_at_location` instead of returning early. The same change is applied in `RebalancingTrigger` so speculative inventory projections also reflect wallet equity readings.

## Testing

- `apply_snapshot_event_populates_inflight_equity_for_base_wallet_unwrapped` — verifies the `BaseWalletUnwrapped` slot is populated and the `BaseWalletWrapped` slot is untouched.
- `apply_snapshot_event_populates_inflight_equity_for_base_wallet_wrapped` — symmetric test for the wrapped slot.
- `apply_snapshot_event_replaces_inflight_equity_at_same_location` — verifies that a subsequent wallet read replaces prior balances, including dropping symbols absent from the new snapshot.
- `venue_inflight_and_inflight_equity_are_tracked_independently` — verifies that the venue-level inflight system and the location-keyed `inflight_equity` map coexist without interfering.
[...]

PR #644, opened 2026-05-07
feat: track all usdc stages

0xgleb:
## What

Introduces an `inflight_cash` map on `InventoryView` that tracks USDC observed at intermediate wallet locations (`Ethereum` and `BaseWallet`) between the two venues. `EthereumUsdc` and `BaseWalletUsdc` snapshot events now populate this map instead of being silently discarded. The `RebalancingTrigger` is updated to forward these events through `apply_snapshot_event` and `force_apply_snapshot_event` accordingly.

## Why

Wallet-read USDC balances represent funds in transit between venues (e.g. between an Alpaca withdrawal and a CCTP burn, or between a CCTP mint and a vault deposit). Previously these events were no-ops, meaning the information was thrown away entirely. Capturing them in a dedicated `inflight_cash` map makes the in-transit state visible for observability and lays the groundwork for orphan-state detection without polluting the venue imbalance math.

## How

A new `InFlightCashLocation` enum (`Ethereum`, `BaseWallet`) keys a `HashMap<InFlightCashLocation, InFlightCashEntry>` stored on `InventoryView`. Each entry carries the `amount` and the `fetched_at` timestamp from the originating snapshot.

`set_inflight_cash` enforces a freshness guard: if the incoming `fetched_at` is strictly older than the entry already on file, the update is dropped and a `warn!` is emitted. This prevents concurrent or out-of-order RPC polls from overwriting a fresher reading with a stale one. Equal timestamps are treated as valid replacements.

`check_usdc_imbalance` is intentionally left unchanged — wallet readings never enter the imbalance math. The design keeps imbalance detection strictly venue-scoped so wallet residue cannot mask or compensate a real venue imbalance. Suppression based on wallet residue is deferred to a future orphan-state detection mechanism.

## Testing

- `apply_ethereum_usdc_populates_inflight_cash_only` and `apply_base_wallet_usdc_populates_inflight_cash_only` verify that wallet-read events populate only the correct `inflight_cash` slot and leave venue inventory untouched.
- `apply_wallet_equity_events_remain_noops_on_view` confirms that `BaseWalletUnwrappedEquity` and `BaseWalletWrappedEquity` remain no-ops on both venue equity and `inflight_cash`.
- `venue_inflight_and_inflight_cash_are_tracked_independently` asserts that per-venue transfer inflight and the wallet-level `inflight_cash` map coexist without interfering.
[...]

PR #643, opened 2026-05-07
feat: track active transfer aggregate IDs in inventory view

0xgleb:
Closes RAI-95.

## What

`InventoryView` now tracks the aggregate IDs of in-flight USDC rebalances, equity mints, and equity redemptions via three new fields: `active_usdc_rebalance`, `active_mints`, and `active_redemptions`. Setter and clearer methods are exposed on `InventoryView`, and `RebalancingTrigger` calls them at the appropriate lifecycle points — setting IDs when operations begin and clearing them on terminal events or timeout cleanup.

## Why

During recovery, the system needs to know which aggregate currently owns each in-flight operation so it can reload stalled aggregates from their stores. Without persisting these IDs in `InventoryView`, that information is lost across restarts.

## How

- Added `active_usdc_rebalance: Option<UsdcRebalanceId>`, `active_mints: HashMap<Symbol, IssuerRequestId>`, and `active_redemptions: HashMap<Symbol, RedemptionAggregateId>` to `InventoryView`, all serialized with `#[serde(default)]`.
- Added `set_active_usdc_rebalance`, `clear_active_usdc_rebalance`, `set_active_mint`, `clear_active_mint`, `set_active_redemption`, and `clear_active_redemption` builder-style methods on `InventoryView`.
- `RebalancingTrigger` calls these setters in `on_mint`, `on_redemption`, and `on_usdc_rebalance` event handlers, and calls the clearers in the corresponding timeout cleanup paths (`cleanup_timed_out_mint`, `cleanup_timed_out_redemption`, `cleanup_timed_out_usdc_rebalance`).
- Two new private helpers, `update_active_mint` and `update_active_redemption`, centralize the set-or-clear logic for mint and redemption events.
- Existing struct literal constructions in `InventoryView` update methods and tests are simplified to use `..self` / `..InventoryView::default()` spread syntax now that the struct has additional fields.

## Testing

New tests verify the full lifecycle for each operation type:
- `no_active_aggregate_ids_when_idle` — all three fields are `None` at rest.
- `mint_accepted_records_active_mint_id` / `mint_terminal_event_clears_active_mint_id` — ID is set on `MintAccepted` and cleared on `DepositedIntoRaindex`.
- `redemption_withdrawn_records_active_redemption_id` / `redemption_completed_clears_active_redemption_id` — ID is set on `WithdrawnFromRaindex` and cleared on `Completed`.
[...]

PR #632, opened 2026-05-05
fix: backfill on reconnect

0xgleb:
Closes RAI-198.

## What

Introduces a durable `BackfillRange` apalis job and a `BackfillJobQueue` so that events emitted during a WebSocket disconnect window are recovered automatically on every (re)connect. The `OrderFillMonitor` now enqueues a `BackfillRange` covering `(checkpoint+1, cutoff_block)` before entering the live listen loop, rather than performing a one-shot backfill at startup. A `StreamWatermarks` tracker ensures the backfill checkpoint only advances past blocks that both the `ClearV3` and `TakeOrderV3` streams have crossed, preventing straggling events from being silently skipped.

## Why

Previously, backfill was performed once at startup by waiting for the first WebSocket event to determine a cutoff block. Any events emitted while the subscription was unavailable — including across supervisor-driven restarts — were silently dropped. The checkpoint also had no protection against rewinding on out-of-order events.

## How

- `BackfillRange` implements `Job<AccountantCtx>` and is processed by a new `backfill-worker` registered in the apalis `Monitor`. Its `perform` delegates to `backfill_range`, which fetches logs in batches and enqueues `AccountForDexTrade` jobs.
- `OrderFillMonitor::run` no longer consumes pre-established streams. On every invocation it calls `prepare_and_listen`, which reads the current provider head as the cutoff, enqueues a `BackfillRange` durably before any awaitable work, then enters the live listen loop.
- `StreamWatermarks` tracks the per-stream high-water mark and exposes `safe_advance_to` (the minimum of both high-water marks minus one), which gates checkpoint writes so neither stream can be skipped.
- The `Job` trait gains `Output`, `WORKER_NAME`, `TERMINAL_FAILURE_MSG`, and `JOB_KIND` associated items. A `build_supervised_worker!` macro replaces the repeated inline worker-wiring blocks in `builder.rs`.
- `AccountantCtx` now carries a `SqlitePool` and `DexTradeAccountingJobQueue` so `BackfillRange::perform` has everything it needs without additional context types.
- The `get_cutoff_block` / `await_next_block` helpers and the `DexEventStreams` hand-off struct are removed.
- Apalis poll interval is capped at 1 s (100 ms initial, 1 s max backoff) to meet the hedge-placement latency SLO.

## Testing

New unit tests cover:
- Checkpoint does not advance until both streams have been observed.
- Checkpoint advances to `min(clear_high, take_high) - 1` once both streams are observed.
[...]

PR #646, opened 2026-05-07
refactor: offchain order orchestration

0xgleb:
Closes RAI-82.

## What

Replaces the cron-driven `OrderStatusPoller` sweep with three durable, per-order apalis job types: `PollOrderStatus`, `ReconcileOrderFill`, and `HandleOrderRejection`. Removes the `apalis-cron` dependency entirely.

## Why

The previous implementation polled all submitted orders in a single cron-driven sweep. This meant a transient broker failure for one order blocked polling of every other order in the same cycle. It also had a crash-safety gap: if the bot restarted between `OffchainOrderCommand::Place` succeeding and the poll job being enqueued, the order would sit in `Submitted` forever with no trigger to resume polling.

## How

Each submitted order now gets its own `PollOrderStatus` job instance in the apalis SQLite queue. When the broker reports the order is still pending, the job self-reschedules with a configurable delay via `push_with_delay` rather than burning a retry. When the broker reports a fill or rejection, the job enqueues a `ReconcileOrderFill` or `HandleOrderRejection` job respectively, isolating the CQRS write into its own retryable unit so a transient DB failure does not force a redundant broker call.

On startup, `recover_submitted_offchain_orders` scans the offchain order projection for any orders still in `Submitted` or `PartiallyFilled` state and re-enqueues a `PollOrderStatus` job for each, closing the crash-recovery gap. `PlaceHedge` and `place_offchain_order` also enqueue a `PollOrderStatus` job immediately after a successful broker submission.

`OrderJobsCtx` holds the shared executor, stores, and all three queues, and is passed as the apalis worker context for all three job types. The apalis monitor wiring is extracted into a `MonitorWiring` struct and a `spawn_apalis_monitor` function to keep `builder.rs` readable.

## Testing

- Unit tests in `order_jobs.rs` cover each job type: filled broker state enqueues `ReconcileOrderFill`, pending broker state self-reschedules with a future `run_at`, failed broker state enqueues `HandleOrderRejection`, and a `FailFirstExecutor` fixture verifies that a broker error on one order's poll does not prevent a second order's poll from succeeding.
- Integration tests in `arbitrage.rs` are updated to use a new `poll_submitted_orders` helper that replicates the job chain's CQRS commands directly, removing the dependency on `OrderStatusPoller`.
- The crash-recovery e2e test tolerance is widened symmetrically (`±4` jobs) to account for the fact that crash recovery can now produce slightly fewer jobs than the reference run when a trade processed before the crash skips re-enqueuing poll jobs on the second pass.

## Anything else

The `OffchainOrder::executor()` accessor and the `FixedInterval` cron schedule type are removed as they are no longer needed. The `max_jitter` field is removed from `OrderPollerCtx` since per-order scheduling makes inter-order jitter unnecessary.

---

<picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture>
<sup>Need help on this PR? Tag `@codesmith` with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

## Summary by CodeRabbit

* **New Features**
  * Job-queue based offchain order processing with robust polling, reconciliation and recovery on startup.
  * Offchain order lifecycle jobs for rejection handling, status polling, and fill reconciliation.

* **Bug Fixes**
  * Trade direction parsing accepts common case formats; DTO now serializes directions as "buy"/"sell".
  * Improved handling of offchain placement/retry edge cases to avoid stuck pending states.

* **Refactor**
  * Reorganized offchain order subsystem and replaced cron pollers with explicit CQRS job workflows.


0xgleb:
* **#645** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/645?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3  dependent PRs ([#647](https://github.com/ST0x-Technology/st0x.liquidity/pull/647) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/647?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#650](https://github.com/ST0x-Technology/st0x.liquidity/pull/650) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/650?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#655](https://github.com/ST0x-Technology/st0x.liquidity/pull/655) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/655?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#646** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/646?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/646?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#643** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/643?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#632** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/632?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#636](https://github.com/ST0x-Technology/st0x.liquidity/pull/636) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/636?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#644](https://github.com/ST0x-Technology/st0x.liquidity/pull/644) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/644?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#649** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/649?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=646).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor (approved):

src/order_jobs.rs
```diff
@@ -0,0 +1,1224 @@
...
+                "ReconcileOrderFill: order not found, skipping"
+            );
+            return Ok(());
+        };
+
+        let symbol = order.symbol().clone();
+        let shares_filled = order.shares();
+        let direction = order.direction();
+
+        complete_offchain_order_fill(&ctx.offchain_order, self.offchain_order_id, self.price)
```
JuaniRios: critical: this two-step write is not retry-safe. If `complete_offchain_order_fill` succeeds but the position update below fails transiently, apalis retries the job. On retry, `CompleteFill` against the now-`Filled` aggregate returns `Err(AlreadyCompleted)`, which propagates via `?` before `complete_position_offchain_order` runs. Every subsequent retry fails the same way, and the position is permanently stuck with `pending_offchain_order_id` set, blocking all future hedges for that symbol.

Fix: at the top of `perform`, check the loaded order's state. If it's already `Filled`, skip `complete_offchain_order_fill` and go straight to the position update. Alternatively, catch `AlreadyCompleted` from the offchain order step and treat it as a no-op.

src/order_jobs.rs
```diff
@@ -0,0 +1,1224 @@
...
+            warn!(
+                offchain_order_id = %self.offchain_order_id,
+                "HandleOrderRejection: order not found, skipping"
+            );
+            return Ok(());
+        };
+
+        let symbol = order.symbol().clone();
+
+        mark_offchain_order_failed(
```
JuaniRios: critical: same retry-safety issue as `ReconcileOrderFill`. If `mark_offchain_order_failed` succeeds but `fail_position_offchain_order` fails, retries hit `AlreadyCompleted` on the already-`Failed` aggregate and never reach the position cleanup. Same fix: check loaded state first and skip the offchain order command if already terminal.

src/trading/offchain/hedge.rs
```diff
@@ -111,18 +113,31 @@ impl Job<HedgeCtx> for PlaceHedge {
...
+                        &self.symbol,
+                        PositionCommand::FailOffChainOrder {
+                            offchain_order_id: self.offchain_order_id,
+                            error,
+                        },
+                    )
+                    .await?;
+            }
+
+            Some(Submitted { .. } | PartiallyFilled { .. }) => {
```
JuaniRios, line 174: should fix: if the `poll_status_queue.push` below fails transiently, the job returns `Err` and apalis retries it. On retry, `PlaceOffChainOrder` returns `PendingExecution` (line ~95), which triggers the early `return Ok(())` before reaching this code path. No `PollOrderStatus` is enqueued, and the order sits in `Submitted` until the next bot restart.

When `PendingExecution` fires, the handler could check if the pending order is `Submitted`/`PartiallyFilled` and re-enqueue `PollOrderStatus` instead of returning immediately. A duplicate poll is harmless because `dispatch_for_order_state` drops terminal-state polls.

src/offchain/order/poll_status.rs
```diff
@@ -0,0 +1,1224 @@
...
+enum PollAction {
+    /// Already terminal or unrecoverable -- discard this poll job.
+    Drop,
+    /// Not yet submitted to the broker. Re-queue and wait.
+    Reschedule,
+    /// Submitted (or partially filled). Ask the broker for status.
+    Query { executor_order_id: ExecutorOrderId },
+}
+
+fn dispatch_for_order_state(
```
JuaniRios, line 234: should fix: the old poller filtered submitted orders by `order.executor() == executor.to_supported_executor()`, but this path sends every submitted order's ID to the currently configured executor without checking whether the order was placed through that executor. If the bot restarts with a different executor config, the wrong broker could receive IDs it doesn't recognize. Worth preserving the executor-type check here and in `recover_submitted_offchain_orders`.

src/conductor.rs
```diff
@@ -1530,20 +1530,99 @@ async fn place_offchain_order(
...
+    dispatch_post_place_state(loaded, execution, cqrs, offchain_order_id).await
+}
+
+/// Routes the freshly-loaded `OffchainOrder` post-`Place` to either the
+/// rejection-cleanup or poll-enqueue path. Returns the order id wrapped in
+/// `Some` when the order is real and recorded, `None` when no usable order
+/// exists (Place silently failed or persisted state is a lifecycle bug), or
+/// propagates infrastructure errors so DB / queue failures cannot be
+/// silently swallowed.
+async fn dispatch_post_place_state(
```
JuaniRios, line 1524: minor: the `None` and `Some(Pending | Filled)` arms return `Ok(None)` without rolling back the position's pending state. After `execute_place_offchain_order` set `pending_offchain_order_id`, these paths leave it dangling. Sending `PositionCommand::FailOffChainOrder` here (like the `Failed` arm does) would prevent the position from getting stuck.

src/offchain/order/poll_status.rs
```diff
@@ -0,0 +1,1224 @@
...
+/// Re-enqueue a [`PollOrderStatus`] job for every offchain order still in a
+/// non-terminal post-submission state. Idempotent: workers that pick up a
+/// duplicate poll job for an already-reconciled order observe the terminal
+/// state and exit cleanly.
+///
+/// Closes the gap between [`OffchainOrderCommand::Place`] succeeding and the
+/// in-process push of the poll job that follows it -- if the bot crashes
+/// between those two writes, the order would otherwise sit in `Submitted`
+/// forever waiting for a poll that never comes.
+pub(crate) async fn recover_submitted_offchain_orders(
```
JuaniRios, line 307: minor: this enqueues a `PollOrderStatus` for every submitted/partially-filled order without checking whether a poll job already exists in the apalis queue from before the crash. If both run concurrently for a still-pending order, both query the broker and both enqueue `ReconcileOrderFill`. Not a problem if the reconcile jobs are made idempotent (see earlier comments), but worth a note in the doc comment that duplicate polls are expected.

tests/e2e/hedging/mod.rs
```diff
@@ -656,24 +656,25 @@ async fn crash_recovery_eventual_consistency() -> anyhow::Result<()> {
...
+    // enqueuing on the second pass. Bound the delta in both directions
+    // to catch runaway re-enqueueing while tolerating these legitimate
+    // differences.
+    const MAX_JOB_DELTA: i64 = 4;
 
     assert!(
-        crash_jobs >= ref_jobs && crash_jobs <= ref_jobs + MAX_DUPLICATE_JOBS,
-        "expected {ref_jobs}..={} jobs but got {crash_jobs}",
-        ref_jobs + MAX_DUPLICATE_JOBS,
+        (crash_jobs - ref_jobs).abs() <= MAX_JOB_DELTA,
```
JuaniRios, line 684: minor: the old assertion enforced `crash_jobs >= ref_jobs` (recovery never produces fewer jobs than a clean run). The new bidirectional bound allows up to 4 fewer jobs, which could mask a bug where recovery fails to re-enqueue a required poll job. Consider adding an assertion that no offchain orders remain in `Submitted`/`PartiallyFilled` after recovery. That directly tests the invariant instead of relying on job counts as a proxy.

src/conductor.rs
```diff
@@ -305,6 +314,16 @@ impl Conductor {
         };
 
         let hedge_queue = crate::conductor::job::JobQueue::new(&pool);
+        let poll_status_queue = PollOrderStatusJobQueue::new(&pool);
+        let reconcile_queue = ReconcileOrderFillJobQueue::new(&pool);
+        let rejection_queue = HandleOrderRejectionJobQueue::new(&pool);
+
+        recover_submitted_offchain_orders(
+            &frameworks.offchain_order_projection,
+            &mut poll_status_queue.clone(),
```
JuaniRios: nit: this passes `&mut poll_status_queue.clone()` instead of `&mut poll_status_queue`. Currently works because `SqliteStorage` shares the pool, but the temporary clone is misleading and would silently lose pushes if the sink ever buffers in memory.

JuaniRios (changes requested):

src/conductor.rs
0xgleb: fixed

src/trading/offchain/hedge.rs
0xgleb: should re-enqueue now

crates/dto/src/trade.rs
```diff
@@ -47,13 +95,31 @@ mod tests {
...
+        assert_eq!(Direction::from_str("buy").unwrap(), Direction::Buy);
+        assert_eq!(Direction::from_str("sell").unwrap(), Direction::Sell);
+    }
+
+    #[test]
+    fn direction_from_str_rejects_unknown_input() {
+        use std::str::FromStr;
+
+        let error = Direction::from_str("hold").unwrap_err();
+        assert_eq!(error.direction_provided, "hold");
```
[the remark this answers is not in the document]
0xgleb: stale

crates/event-sorcery/src/lib.rs
```diff
@@ -317,12 +317,36 @@ impl<Entity: EventSourced> Store<Entity> {
...
+                tracing::error!(
+                    aggregate = Entity::AGGREGATE_TYPE,
+                    %id,
+                    command = %command_repr,
+                    %error,
+                    "Command failed"
+                );
+                Err(error)
+            }
+        }
```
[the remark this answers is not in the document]
0xgleb: this is the command handler

src/offchain/order/poll_status.rs
0xgleb: brought back the logic

tests/e2e/hedging/mod.rs
0xgleb: fixed

JuaniRios (approved):

PR #685, opened 2026-05-22
fix: normalize absent Alpaca positions to zero

JuaniRios:
## What

[RAI-648](https://linear.app/makeitrain/issue/RAI-648/normalize-absent-alpaca-positions-to-zero-for-configured-assets)

Normalizes absent Alpaca equity positions to explicit zero balances for configured active assets, and keeps the live inventory view consistent with the persisted snapshot when a symbol drops to zero.

- `InventoryPollingService` now emits zero-share offchain equity positions for configured symbols omitted from Alpaca's positions response.
- Broker-only positions for unconfigured symbols are ignored and warned about (once per contiguous appearance) instead of entering inventory state. This also resolves a prod worker crash loop (2026-05-22): unconfigured Alpaca symbols (CEG/DRAM/TSM) were flowing into the offchain inventory snapshot and reaching the equity rebalance check, which crashed on a vault-registry lookup for the unknown token (`token not in vault registry: CEG`) and put the worker into a restart loop. Filtering them out at the snapshot prevents them from ever reaching that check.
- `InventoryView` now treats an equity snapshot as the complete picture of a venue: symbols tracked at that venue but absent from the snapshot are zeroed, so the live view no longer retains stale balances.
- Rebalancing suppresses equity triggers while an offchain hedge order is pending for the same symbol, recovers that pending state on restart, and re-checks immediately before dispatch.

## Why

Alpaca omits zero-share positions from its positions API. Without normalization, inventory could retain stale offchain equity balances for configured assets that had been reduced to zero, because no new position row arrived to clear the old value.

Configured active assets should always have an explicit offchain balance in inventory snapshots, and symbols outside the bot's configured active universe should not affect inventory or rebalance decisions. The persisted `InventorySnapshot` aggregate already fully replaces its balance map on each poll, but the live in-memory `InventoryView` applied snapshots per-symbol and never cleared omitted symbols -- so a balance reduced to zero (or an unconfigured symbol hydrated from older state) could survive in the live view that the dashboard and rebalancer read. This change makes the two views consistent.

## How

- `SPEC.md` documents both the offchain polling normalization (active configured equities missing from broker positions are treated as zero; broker-only unconfigured symbols are ignored) and the `InventoryView` complete-snapshot semantics for both venues, including the registry-completeness invariant onchain zeroing relies on.
[...]

PR #689, opened 2026-05-26
feat: enable production USDC rebalancing

JuaniRios:
## What

Related: [RAI-685](https://linear.app/makeitrain/issue/RAI-685/decide-production-usdc-rebalancing-policy)
Follow-up: [RAI-687](https://linear.app/makeitrain/issue/RAI-687/use-alpaca-portfolio-value-for-usdc-reserve-calculation)

Enables production USDC rebalancing and updates the reserve semantics so `reserved` cash no longer affects the onchain/offchain ratio calculation. The prod config now rebalances USDC through vault `0xfab4` with a 50% target, 20% deviation, 1,000 USDC per-operation limit, and 25,000 USD offchain reserve.

## Why

The previous implementation subtracted `reserved` cash before calculating the USDC inventory ratio. That made the system treat reserved Alpaca cash as if it did not exist, which could make inventory look artificially onchain-heavy and even flip rebalance direction in some snapshots.

The intended behavior is:

- ratio calculation uses gross venue inventory
- `reserved` protects offchain cash only when moving funds Alpaca -> Base
- Base -> Alpaca transfers are not constrained by Alpaca withdrawable cash or reserve

## How

- Updates `config/prod/st0x-hedge.toml`:
  - enables `[rebalancing.usdc]`
  - sets `target = 0.5` and `deviation = 0.2`
  - switches `[assets.cash].vault_id` to `0xfab4`
  - sets `operational_limit = 1000`
  - sets `reserved = 25000`

- Adds `InventoryView::check_usdc_imbalance_with_gross_offchain`, which uses `offchain_gross_usd_cents` when available instead of the reserve-adjusted offchain venue balance.

- Adds `InventoryView::alpaca_to_base_usdc_capacity`, which computes Alpaca outbound capacity from `withdrawable_cash_cents - reserved`.

- Updates the USDC trigger path so:
  - `TooMuchOffchain` / Alpaca -> Base is capped by operational limit and reserve-adjusted withdrawable capacity
  - missing withdrawable cash skips Alpaca -> Base because capacity cannot be proven reserve-safe
  - `TooMuchOnchain` / Base -> Alpaca ignores reserved and withdrawable cash

- Updates USDC trigger and integration test fixtures to model withdrawable cash explicitly.

## Testing

Added and updated coverage for:

- gross offchain cash being used in the ratio even when `reserved` is configured
- Alpaca -> Base skipping when withdrawable cash is missing
- Alpaca -> Base capping by `withdrawable_cash - reserved`
[...]

PR #690, opened 2026-05-26
fix: clear offchain hedge gate even when inventory update fails

JuaniRios:
## What

Fixes a deadlock in the equity rebalancing gate introduced by #685.

When an `OffChainOrderFilled` event arrives, the rebalancer updates an in-memory `InventoryView` mirror and removes the symbol from `pending_offchain_order_symbols` (the gate that suppresses duplicate equity rebalance triggers while a hedge order is in flight). Previously, the gate-clearing was deferred until *after* the inventory update succeeded — fail-closed. If the inventory update errored, the gate stayed set on the assumption that "the next terminal event for this symbol, or a restart, will clear it."

That assumption is wrong. The only path that can produce another `OffChainOrderFilled`/`OffChainOrderFailed` for the symbol is itself gated by `pending_offchain_order_symbols`, so once the flag sticks it stays stuck until the bot is restarted.

## Why

Hit in prod today (2026-05-26) on **MSTR**, and earlier the same day on **SPYM**. The pattern:

1. The inventory poller calls Alpaca's `/positions` endpoint at `T+0`. The broker had just filled a 5.143-share sell, so the response shows `MSTR = 4.698` (post-fill).
2. The mirror is overwritten with `4.698`.
3. ~0.1s later the bot's order-status poll receives the fill notification.
4. The `OffChainOrderFilled` reactor fires, tries to subtract `5.143` from the mirror, gets `InsufficientAvailable { requested: 5.143, available: 4.698 }`.
5. The fail-closed branch leaves MSTR in `pending_offchain_order_symbols`.
6. Every subsequent `EquityRebalancingCheck(MSTR)` job logs `Skipped equity trigger: offchain hedge order pending symbol=MSTR` and does nothing.

MSTR equity rebalancing was silently suppressed for ~15 minutes until the operator restarted the bot. The onchain vault held 14.76 MSTR while the broker held 4.70 — drift would have kept growing.

## How

Always clear the gate on a `Filled` event, even if the inventory update fails. If it fails, log a warning at `target: "rebalance"` and rely on the next polling snapshot (~60s) to reconcile the mirror.

Trade-off: at most one stale rebalance trigger window (~60s) before polling overwrites the in-memory value, versus an indefinite deadlock requiring manual restart. The drift is bounded to wasted rebalances, not lost capital — the polling snapshot is the source of truth.

Implementation detail: restructured the inventory-apply block in `src/rebalancing/trigger/mod.rs` to capture the result without `?`, so the gate-clear branch can run regardless of inventory outcome and early-return. The `OnChainOrderFilled` path keeps the original `?` propagation since it has no gate to deadlock.

## Testing

New unit test `offchain_order_filled_releases_gate_even_when_inventory_update_fails` (`src/rebalancing/trigger/mod.rs`) sets up an inventory where offchain shares are smaller than the incoming fill, fires `OffChainOrderPlaced` then `OffChainOrderFilled`, and asserts:

- the reactor does not propagate the inventory underflow (returns `Ok`)
- the symbol is no longer in `pending_offchain_order_symbols`

The test fails against the previous code with `Inventory(Equity(InsufficientAvailable { requested: FractionalShares(10), available: FractionalShares(4) }))` (confirmed locally before the fix).

Existing tests around the gate lifecycle (`offchain_order_filled_releases_equity_rebalancing_block`, `offchain_order_failed_enqueues_equity_recheck`, `offchain_order_filled_enqueues_equity_recheck`, `position_offchain_order_lifecycle_blocks_then_releases_equity_check`, `recover_pending_offchain_order_symbols_restores_block_from_projection`) still pass — the only behavior change is on the inventory-error path.

`cargo check -p st0x-hedge --tests` passes locally. Full test suite left to CI (workspace nextest build is slow on this machine).

## Screenshots

N/A.

## Anything else

This patches the deadlock symptom, not the underlying race. The root cause is that the in-memory `InventoryView` has two independent writers — the polling loop (overwrites from broker truth every ~60s) and the fill reactor (subtracts on order events) — with no coordination, so the same fill can be applied twice. The `InsufficientAvailable` error is the second writer noticing the first writer already absorbed the fill.

Follow-up tracked in [RAI-682](https://linear.app/makeitrain/issue/RAI-682/design-proper-reconciliation-between-inventory-poller-and-offchain-order) — designing a proper fix (track inflight orders so the snapshot can be adjusted, use broker timestamps to detect "snapshot already includes this fill," or eliminate one of the writers entirely). The deadlock fix here is independently shippable and not blocked on that design.

Related PRs:
- Introduced by #685 (`fix: normalize absent Alpaca positions to zero`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented off-chain filled orders from becoming stuck when inventory updates fail: pending state is now cleared, a warning is logged when clearing on failure, an equity recheck is always queued, and a cash/USDC recheck is queued only if the inventory update succeeded.

* **Tests**
  * Added a regression test to ensure an off-chain filled-event underflow does not leave orders stuck.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ST0x-Technology/st0x.liquidity/pull/690?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/rebalancing/trigger/mod.rs
+++ b/src/rebalancing/trigger/mod.rs
@@ -1457,13 +1457,15 @@ impl Reactor for RebalancingService {
                         price,
                         ..
                     } => {
-                        // Defer clearing the suppression until the fill has been
-                        // applied to inventory below. If that update fails, the
-                        // gate deliberately stays closed: we never resume equity
-                        // rebalancing on inventory we failed to record the fill
-                        // into. It self-heals when the next terminal offchain
-                        // order event for this symbol applies cleanly, or on
-                        // restart via recover_pending_offchain_order_symbols.
+                        // Always clear the gate on a Filled event, even if the
+                        // inventory update below fails. A fail-closed gate would
+                        // deadlock equity rebalancing for the symbol until the
+                        // next bot restart, because the only event that could
+                        // re-clear it (a later terminal offchain event) is
+                        // itself gated. The polling cycle is the source of
+                        // truth for the broker balance, so any local
+                        // bookkeeping miss self-heals within ~60s and is bounded
+                        // to wasted rebalances, not lost capital.
                         clear_pending_offchain_order = true;
                         let equity_op: Operator = (*direction).into();
                         let quantity: Float = shares_filled.inner().into();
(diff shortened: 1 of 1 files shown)


JuaniRios:
* **#689** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/689?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#690** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/690?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/690?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=690).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor (approved):

0xgleb (approved):

PR #691, opened 2026-05-26
Feat/pnl dashboard prototype

data4rtist:
## What

  Adds a PnL tab to the dashboard.

  The tab reads persisted fill data through the deployed SQL JSON endpoint and displays:

  - counter-trade PnL
  - on-chain netting PnL
  - directional exposure PnL
  - total PnL
  - current replay exposure
  - unmatched counter-trade exposure
  - per-symbol breakdowns
  - time-bucketed chart views
  - expandable debug notes for replay/reconciliation limitations

  Also adds local dev configuration for connecting the dashboard to a production backend
  and SQL JSON endpoint.

  ## Why

  We need a dashboard view over the PnL behavior of the currently deployed system using
  persisted production data.

  The deployed persistence model does not currently store explicit hedge parentage or
  historical portfolio state plus price/NAV snapshots. This PR makes the available
  realized replay view usable while surfacing those persistence limits directly in the
  dashboard.

  ## How

  The SQL PnL source builds a deterministic replay from persisted `Position` events:

  - on-chain fills open or net inventory lots
  - offchain fills are allocated against open opposite-side onchain lots by FIFO replay
  - prompt offchain matches are attributed to counter-trade PnL
  - late matched offchain fills are attributed to directional exposure
  - unmatched offchain fills are carried as current unmatched exposure
  - replayed open lots are reconciled against `position_view`
  - debug notes summarize allocation/reconciliation limitations

  Realized PnL remains date-filtered for the selected dashboard window. Current replay
  exposure is preserved separately so open and unmatched exposure does not disappear when
  there are no closed lots in the selected period.

  Percentage return display is intentionally omitted because historical portfolio state
  and price/NAV snapshots are not currently persisted. Baseline drift is reported as zero
  for the same reason.

  ## Testing

  Tested locally after rebasing onto latest `origin/master`.

  Commands run:

  ```bash
  npx eslint src/lib/components/pnl-panel.svelte src/lib/pnl/api.ts src/lib/pnl/
  api.test.ts src/lib/pnl/sql-source.ts src/lib/pnl/sql-source.test.ts src/lib/pnl/
  synthetic.ts src/lib/pnl/synthetic.test.ts src/lib/pnl/types.ts src/routes/+page.svelte
  src/lib/env.ts src/lib/env.test.ts vite.config.ts src/lib/components/header-bar.svelte

  npm run test:run -- src/lib/pnl/sql-source.test.ts src/lib/pnl/api.test.ts src/lib/pnl/
  synthetic.test.ts src/lib/env.test.ts
  ```
  Result: 29 tests passed.

  Also smoke-tested local dev against production data:

  - /health proxy to production backend
  - /__pnl_sql proxy to production SQL JSON endpoint

  ## Screenshots
  
image
image
image
Screenshot 2026-05-27 at 16 52 22

  ## Anything else

  Known current limitations:

  - explicit offchain_order_id -> onchain_trade_ids parentage is not persisted
  - historical portfolio state and price/NAV snapshots are not persisted
  - baseline drift is therefore reported as zero
  - replayed open lots may differ from position_view when the persisted fill history does
    not fully reconstruct projected state

---
<a href="https://app.blacksmith.sh/ST0x-Technology/codesmith/st0x.liquidity/pr/691"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782425034&installation_id=125569124&pr_number=691&repository=ST0x-Technology%2Fst0x.liquidity&return_to=https%3A%2F%2Fgithub.com%2FST0x-Technology%2Fst0x.liquidity%2Fpull%2F691&signature=e165648bd7b812b6b54595f016579a6efc1856ad85d9ef105a6e033b7205f634"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

* **New Features**
  * PnL dashboard tab with interactive charts, stacked/cumulative views, asset multiselect, date-range presets/custom dates, day-type filtering, summary cards and detailed tables.
  * Local mock mode to use synthetic PnL data and avoid backend/WebSocket setup for development.
  * Optional SQL-backed PnL source for querying real backend data.

* **Documentation**
  * Expanded dashboard README with setup/config examples for mock mode, SQL PnL endpoint, and local proxying.

### DIFF
--- a/dashboard/src/lib/components/pnl-panel.svelte
+++ b/dashboard/src/lib/components/pnl-panel.svelte
@@ -1,0 +1,1846 @@
+<script lang="ts">
+  import { onMount } from 'svelte'
+  import * as Card from '$lib/components/ui/card'
+  import * as Table from '$lib/components/ui/table'
+  import MultiSelect from '$lib/components/multi-select.svelte'
+  import { decimalCompare, formatDecimal } from '$lib/decimal'
+  import { getPnlSqlApiUrl, isDashboardMockMode } from '$lib/env'
+  import { reactive } from '$lib/frp.svelte'
+  import { formatUtc } from '$lib/time'
+  import { fetchPnlReport } from '$lib/pnl/api'
+  import { STREAM_KEYS } from '$lib/pnl/report'
+  import { syntheticPnlDashboard } from '$lib/pnl/synthetic'
+  import type {
+    PnlEntry,
+    PnlResponse,
+    PnlStreamKey,
+    PnlSummary,
+    PnlSymbolSummary,
+    PnlWindow,
+    PnlWindowSymbol
+  } from '$lib/pnl/report'
+
+  type RangePreset = '1w' | '1m' | 'ytd' | '1y' | 'all' | 'custom'
+
+  const PNL_ENTRY_LIMIT = Number.MAX_SAFE_INTEGER
+  const POLL_INTERVAL_MS = 10_000
+  const MS_PER_DAY = 24 * 60 * 60 * 1_000
+  const mockMode = isDashboardMockMode()
+  const hasSqlPnlSource = getPnlSqlApiUrl() !== null
+
+  const dateFromIso = (value: string): Date => new Date(`${value}T00:00:00.000Z`)
+
+  const isoDate = (date: Date): string => date.toISOString().slice(0, 10)
+
+  const shiftDate = (value: string, days: number): string => {
+    const date = dateFromIso(value)
+    date.setUTCDate(date.getUTCDate() + days)
+    return isoDate(date)
+  }
+
+  const syntheticAvailableStartDate = syntheticPnlDashboard.windows[0]?.startAt.slice(0, 10) ?? ''
+  const syntheticAvailableEndDate = syntheticPnlDashboard.windows.at(-1)?.startAt.slice(0, 10) ?? ''
+  const initialAvailableEndDate =
+    !mockMode && hasSqlPnlSource ? new Date().toISOString().slice(0, 10) : syntheticAvailableEndDate
+  const initialAvailableStartDate =
+    !mockMode && hasSqlPnlSource
+      ? shiftDate(initialAvailableEndDate, -6)
+      : syntheticAvailableStartDate
+  const defaultChartSymbols = syntheticPnlDashboard.report.symbols.map((row) => row.symbol).sort()
+
+  const clampDateForRange = (value: string, start: string, end: string): string => {
+    if (!start || !end) return value
+    if (value < start) return start
+    if (value > end) return end
+    return value
+  }
+
+  const startOfYear = (value: string): string => `${value.slice(0, 4)}-01-01`
+  const presetStartForRange = (preset: RangePreset, start: string, end: string): string => {
+    switch (preset) {
+      case '1w':
+        return clampDateForRange(shiftDate(end, -6), start, end)
+      case '1m':
+        return clampDateForRange(shiftDate(end, -29), start, end)
+      case 'ytd':
+        return clampDateForRange(startOfYear(end), start, end)
+      case '1y':
+        return clampDateForRange(shiftDate(end, -364), start, end)
+      case 'all':
+        return start
+      case 'custom':
+        return clampDateForRange(shiftDate(end, -6), start, end)
+    }
+  }
+
+  const report = reactive<PnlResponse | null>(null)
+  const entries = reactive<PnlEntry[]>([])
+  const loading = reactive(false)
+  const error = reactive<string | null>(null)
+  const total = reactive(0)
+  const selectedSymbols = reactive<Set<string>>(new Set())
+  const selectedAssetChartSymbols = reactive<Set<string>>(new Set(defaultChartSymbols))
+  const selectedAssetChartStreams = reactive<Set<PnlStreamKey>>(new Set(STREAM_KEYS))
+  const selectedMethodChartSymbols = reactive<Set<string>>(new Set(defaultChartSymbols))
+  const selectedMethodChartStreams = reactive<Set<PnlStreamKey>>(new Set(STREAM_KEYS))
+  const dayFilter = reactive<'all' | 'weekday' | 'weekend'>('all')
+  const dataMode = reactive<'synthetic' | 'backend'>(
+    !mockMode && hasSqlPnlSource ? 'backend' : 'synthetic'
+  )
+  const availableStartDate = $derived.by(() => {
+    if (dataMode.current === 'synthetic') return syntheticAvailableStartDate
+    return (
+      report.current?.sampleStats?.firstAt?.slice(0, 10) ??
+      report.current?.windows?.[0]?.startAt.slice(0, 10) ??
+      initialAvailableStartDate
+    )
+  })
+  const availableEndDate = $derived.by(() => {
+    if (dataMode.current === 'synthetic') return syntheticAvailableEndDate
+    return (
+      report.current?.sampleStats?.lastAt?.slice(0, 10) ??
+      report.current?.windows?.at(-1)?.startAt.slice(0, 10) ??
+      initialAvailableEndDate
+    )
+  })
+  const clampDate = (value: string): string =>
+    clampDateForRange(value, availableStartDate, availableEndDate)
+  const presetStart = (preset: RangePreset): string =>
+    presetStartForRange(preset, availableStartDate, availableEndDate)
+  const allSymbols = reactive<string[]>([])
+  const selectedPreset = reactive<RangePreset>('1w')
+  const fromDate = reactive(
+    presetStartForRange('1w', initialAvailableStartDate, initialAvailableEndDate)
+  )
+  const toDate = reactive(initialAvailableEndDate)
+  const dateRangeKey = (from: string, to: string): string => `${from}:${to}`
+  let lastRequestedDateRangeKey: string | null = null
+  let fetchSeq = 0
+
+  const streamLabels: Record<PnlStreamKey, string> = {
+    counterTradePnlUsd: 'Counter-trade',
+    onchainNettingPnlUsd: 'On-chain netting',
+    directionalInventoryBaselinePnlUsd: 'Baseline drift',
+    directionalImbalanceExcessPnlUsd: 'Directional realized'
+  }
+  const streamColors: Record<PnlStreamKey, string> = {
+    counterTradePnlUsd: '#38bdf8',
+    onchainNettingPnlUsd: '#22c55e',
+    directionalInventoryBaselinePnlUsd: '#f59e0b',
+    directionalImbalanceExcessPnlUsd: '#f43f5e'
+  }
+  const assetColors = ['#38bdf8', '#f59e0b', '#22c55e', '#f43f5e', '#a78bfa']
+  const formulaRows = [
+    {
+      label: 'Counter-trade PnL',
+      formula: 'sum((broker_price - onchain_price) * matched_qty * side_sign)',
+      note: 'Only broker fills inside the counter-trade threshold are attributed here. side_sign is +1 for long inventory opened by the on-chain fill and -1 for short inventory.'
+    },
+    {
+      label: 'On-chain netting PnL',
+      formula: 'sum((closing_onchain_price - opening_onchain_price) * matched_qty * side_sign)',
+      note: 'Used when a later opposite on-chain fill closes inventory before or instead of a broker counter-trade.'
+    },
+    {
+      label: 'Baseline drift PnL',
+      formula: 'sum(frozen_start_qty * (bucket_end_mark - bucket_start_mark))',
+      note: 'Requires historical portfolio and mark snapshots; currently reported as zero.'
+    },
+    {
+      label: 'Directional realized PnL',
+      formula: 'sum((close_price - open_price) * matched_qty * side_sign)',
+      note: 'Realized closed-lot PnL for delayed or offchain-origin exposure.'
+    },
+    {
+      label: 'Directional realized total',
+      formula: 'baseline_drift_pnl + directional_excess_pnl',
+      note: 'Baseline drift is zero until historical snapshots are persisted.'
+    },
+    {
+      label: 'Realized gross PnL',
+      formula:
+        'counter_trade_pnl + onchain_netting_pnl + baseline_drift_pnl + directional_excess_pnl',
+      note: 'Gross realized replay by close date; costs, financing, and unrealized NAV drift are not included.'
+    }
+  ]
+
+  const streamOptions = STREAM_KEYS.map((stream) => ({
+    value: stream,
+    label: streamLabels[stream]
+  }))
+
+  const symbolOptions = $derived(
+    allSymbols.current.map((symbol) => ({
+      value: symbol,
+      label: symbol
+    }))
+  )
+
+  const isPnlStreamKey = (value: string): value is PnlStreamKey =>
+    (STREAM_KEYS as readonly string[]).includes(value)
+
+  const isWeekendTimestamp = (timestamp: string): boolean =>
+    [0, 6].includes(new Date(timestamp).getUTCDay())
+
+  const matchesDayFilter = (isWeekend: boolean): boolean => {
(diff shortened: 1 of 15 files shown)


dashboard/src/lib/pnl/sql-source.ts
```diff
@@ -0,0 +1,1221 @@
...
+
+export const buildSqlApiUrl = (baseUrl: string, sql: string): string => {
+  const url = new URL(baseUrl, globalOrigin())
+  url.searchParams.set('sql', sql)
+  url.searchParams.set('_shape', 'array')
+  url.searchParams.set('_size', 'max')
+  return url.toString()
+}
+
+const fetchSqlRows = async <Row>(baseUrl: string, sql: string): Promise<Row[]> => {
```
JuaniRios, line 208: critical: this issues one GET with `_size=max` and treats the array as complete, but Datasette clamps that to `max_returned_rows` (default 1000) and under `_shape=array` the truncation indicator is stripped. The position-events query has no LIMIT and no pagination, so once the events table passes the cap the FIFO replay silently runs on a prefix of history and the displayed PnL is wrong with no warning. Even as a prototype this needs to fail closed: at minimum, request `_shape=objects` and surface a hard warning (and refuse to render PnL) when the returned row count equals the page cap. Pagination is the right fix but the warning is a one-liner.

dashboard/src/lib/pnl/sql-source.ts
```diff
@@ -0,0 +1,1221 @@
...
+  if (!Array.isArray(rows)) {
+    throw new Error('SQL endpoint returned a non-array response')
+  }
+
+  return rows as Row[]
+}
+
+const positionEventsSql = (symbols: Set<string>): string => `
+SELECT rowid, aggregate_id AS symbol, event_type, payload
+FROM events NOT INDEXED
```
JuaniRios: should fix: `NOT INDEXED` forces a full table scan against an append-only CQRS log, and the client wraps it in a 5s `AbortSignal.timeout`. For any production-sized DB this exceeds 5s and the dashboard surfaces "The operation was aborted" with no diagnosis. Drop the `NOT INDEXED` hint (or add an index on `(aggregate_type, event_type)`) and use a longer timeout for the SQL path than for normal API calls.

dashboard/src/lib/pnl/sql-source.ts
```diff
@@ -0,0 +1,1221 @@
...
+  unmatchedOffchainShortLots: [],
+  seenOnchainFillIds: new Set(),
+  seenOffchainPlacementIds: new Set(),
+  seenOffchainFillIds: new Set(),
+  originalOnchainShares: new Map(),
+  matchedOnchainShares: new Map(),
+  summary: emptySummary()
+})
+
+const sqlString = (value: string): string => `'${value.replaceAll("'", "''")}'`
```
JuaniRios, line 170: minor: defense in depth on the raw-SQL endpoint. `sqlString` only escapes single quotes; symbols flow from the SQL response back into the next query with no allow-list, and the `column` parameter is interpolated raw (safe today because all call sites pass literals, but a latent regression). Validate symbols against something like `/^[A-Za-z0-9._-]+$/` and drop non-matching entries with a warning; pin `column` to a typed enum. Cheap, and the SQL endpoint exists today regardless of when backend PnL lands.

dashboard/src/lib/components/pnl-panel.svelte
```diff
@@ -0,0 +1,1848 @@
...
+      let negativeCursor = 0
+
+      return {
+        key: row.key,
+        label: row.label,
+        segments: row.segments.map((segment) => {
+          const cumulative = (cumulativeByKey.get(segment.key) ?? 0) + segment.value
+          cumulativeByKey.set(segment.key, cumulative)
+
+          if (cumulative >= 0) {
```
JuaniRios: should fix: stacking-by-sign per bucket produces invalid polygons. `positiveCursor`/`negativeCursor` reset per row, and each key's cumulative is placed in the positive or negative stack based on its sign at that bucket. The same key can land in opposite stacks across adjacent buckets, so y-coordinates jump discontinuously across zero and the polygon (built by walking `y1` forward and `y0` backward) self-intersects. Once any layer's cumulative crosses zero, which happens over any meaningful window, the chart misrepresents the data. Stack layers by stable insertion order with signed offsets, or render each layer as a line of `cumulative` without sign-based stacking.

dashboard/src/lib/components/pnl-panel.svelte
```diff
@@ -0,0 +1,1848 @@
...
+        ? clampDate(fromDate.current)
+        : presetStart(selectedPreset.current)
+    const nextTo =
+      selectedPreset.current === 'custom' ? clampDate(toDate.current) : availableEndDate
+
+    if (nextFrom === fromDate.current && nextTo === toDate.current) return
+
+    fromDate.update(() => nextFrom)
+    toDate.update(() => nextTo)
+    void fetchPnl()
```
JuaniRios: should fix: this effect self-amplifies. It reads `availableEndDate` (derived from the most recent fetch's `sampleStats`), reconciles `toDate`, and calls `fetchPnl()`. Every successful poll that returns a different `lastAt` re-runs the effect, mutates `toDate`, and fires another fetch. Combined with the missing request cancellation that CodeRabbit flagged on line 440, one 10s tick can produce two overlapping SQL fetches racing each other. It also silently snaps the visible date range under the user when the preset isn't `custom`. Skip the `fetchPnl` call if the resolved dates match those of the most recent fetch, and only snap when the date actually changed.

dashboard/src/lib/components/pnl-panel.svelte
```diff
@@ -0,0 +1,1848 @@
...
+      report.update(() => data)
+      total.update(() => data.total)
+      entries.update(() => data.entries)
+
+      syncAvailableSymbols(data.symbolUniverse ?? data.symbols.map((summary) => summary.symbol))
+    } catch (fetchError) {
+      error.update(() => (fetchError instanceof Error ? fetchError.message : 'Unknown error'))
+    } finally {
+      loading.update(() => false)
+    }
```
JuaniRios, line 456: agree with this finding. The race is real and easy to hit because date pickers fire on every keystroke and the 10s poll overlaps user input. An `AbortController` (or monotonic request token) tracked in a `reactive` is the smallest fix.

JuaniRios (changes requested):
I get this is a prototype to surface some PnL data while the backend doesn't have proper tracking, and I'm fine merging it as-is (after comments are addressed) to start gathering some profitability data. The inline comments cover small fixes worth doing regardless. But before one of the engineers picks up the backend side, we need to align on the model, because the trade-pairing approach this PR takes can't be made correct here *or* in the backend, and the worst outcome is investing weeks of Rust work into the same fundamental mismatch.

## The model mismatch

The dashboard is computing PnL trade-by-trade, pairing specific onchain fills with specific broker fills via FIFO. But the bot doesn't work that way. It accumulates onchain fills into a per-symbol `net` position and fires a single Alpaca hedge for the full net when `|net|` crosses an `ExecutionThreshold`. That means one Alpaca order routinely covers many onchain fills that accumulated since the last hedge, and an onchain fill that arrives mid-hedge ends up split: part of `net` gets absorbed by the current cycle, the rest by the next one. There's no canonical mapping between an Alpaca order and the onchain fills it hedged, anywhere in the system. The relationship is many-to-many and emergent from timing.

That's why the replay needs the "unmatched offchain" pool that never resolves, the hardcoded 300-second threshold to guess "prompt vs late", and a separate reconciliation against `position_view` that ends up disagreeing with the replay. These aren't bugs we can fix in the dashboard, and they aren't fixable in the backend either: the parentage is something we'd have to invent at hedge time, and there's no canonical way to do it for a position-rebalancing system.

## What we can compute correctly from position data

Treat each symbol as a single signed inventory number that goes up and down with every event. The `events` stream already has everything we need: every `OnChainOrderFilled` and `OffChainOrderFilled` changes inventory by `±shares` based on direction, and the running net is what `position_view.net_position` already snapshots.

The accounting rule is plain inventory FIFO. When inventory moves *away* from zero, we opened a lot at that event's price. When inventory moves *toward* zero, we're closing exposure: match against the oldest open lot and record realized PnL = `shares_closed × (open_price − close_price)` with the sign matching long vs short.

Worked example for tAAPL:

- onchain user buys 10 tAAPL @ $103 → net = −10, open short lot 10 @ $103
- bot buys 5 AAPL from Alpaca @ $100 → net = −5, close 5 of the short lot, **realized = +$15**
- bot buys 5 more AAPL from Alpaca @ $102 → net = 0, close the rest, **realized = +$5**
- onchain user buys 8 tAAPL @ $104 → net = −8, open short lot 8 @ $104
- bot buys 3 AAPL from Alpaca @ $104 → net = −5, **realized = $0**

End state: net = −5, one open lot of 5 short @ $104, **realized = $20**.

Then add the unrealized side: if the current Alpaca mark for AAPL is $106, we're short 5 shares at basis $104, so closing now would cost us $10. **Unrealized = −$10. Total PnL = $10.**

That's the whole calculation. No threshold, no parentage, no "matched vs unmatched offchain", no two-stream interleaving. And the end-state inventory the replay produces should equal `position_view.net_position` exactly. If it doesn't, that's a real bug to chase, because both are deterministic functions of the same event stream.

## The buckets become quality metrics, not PnL

The counter-trade / directional / netting split is genuinely useful operationally ("are we capturing the spread cleanly, or eating drift while waiting to hedge"), but it's not PnL attribution. It's two separate quality questions, each answerable from the data we already have:

- **Spread captured**: for each onchain fill, compare its price to the nearest broker price for the same symbol in time. Compute the per-share gap (sign-adjusted for direction), aggregate as volume-weighted average per time window, report in basis points. Tells us what gross spread the bot is earning per round-trip without pretending to pair specific trades. Drift cost during the hedging lag is captured by the next metric, not this one.
- **Directional risk**: integrate `|net_position|` over time per symbol (share-hours of exposure carried), and report the median time from an onchain fill until `|net|` falls back below `ExecutionThreshold`. Tells us how much exposure the bot is sitting on while it waits for the threshold trigger to fire.

These belong next to the PnL numbers, not inside them.

## Concrete question

Do the three PnL buckets matter to whoever will look at this, or were they framed that way because per-trade attribution felt more granular than position-level realized + unrealized? If the latter, simplifying is pure upside: half of `sql-source.ts` deletes, the eventual backend `/pnl` endpoint becomes obvious (we already emit position events and have marks in `position_view`), and the SPEC.md write-up is half a page instead of a multi-page attribution model.

data4rtist:
> I get this is a prototype to surface some PnL data while the backend doesn't have proper tracking, and I'm fine merging it as-is (after comments are addressed) to start gathering some profitability data. The inline comments cover small fixes worth doing regardless. But before one of the engineers picks up the backend side, we need to align on the model, because the trade-pairing approach this PR takes can't be made correct here _or_ in the backend, and the worst outcome is investing weeks of Rust work into the same fundamental mismatch.
> 
> ## The model mismatch
> The dashboard is computing PnL trade-by-trade, pairing specific onchain fills with specific broker fills via FIFO. But the bot doesn't work that way. It accumulates onchain fills into a per-symbol `net` position and fires a single Alpaca hedge for the full net when `|net|` crosses an `ExecutionThreshold`. That means one Alpaca order routinely covers many onchain fills that accumulated since the last hedge, and an onchain fill that arrives mid-hedge ends up split: part of `net` gets absorbed by the current cycle, the rest by the next one. There's no canonical mapping between an Alpaca order and the onchain fills it hedged, anywhere in the system. The relationship is many-to-many and emergent from timing.
> 
> That's why the replay needs the "unmatched offchain" pool that never resolves, the hardcoded 300-second threshold to guess "prompt vs late", and a separate reconciliation against `position_view` that ends up disagreeing with the replay. These aren't bugs we can fix in the dashboard, and they aren't fixable in the backend either: the parentage is something we'd have to invent at hedge time, and there's no canonical way to do it for a position-rebalancing system.
> 
> ## What we can compute correctly from position data
> Treat each symbol as a single signed inventory number that goes up and down with every event. The `events` stream already has everything we need: every `OnChainOrderFilled` and `OffChainOrderFilled` changes inventory by `±shares` based on direction, and the running net is what `position_view.net_position` already snapshots.
> 
> The accounting rule is plain inventory FIFO. When inventory moves _away_ from zero, we opened a lot at that event's price. When inventory moves _toward_ zero, we're closing exposure: match against the oldest open lot and record realized PnL = `shares_closed × (open_price − close_price)` with the sign matching long vs short.
> 
> Worked example for tAAPL:
> 
> * onchain user buys 10 tAAPL @ $103 → net = −10, open short lot 10 @ $103
> * bot buys 5 AAPL from Alpaca @ $100 → net = −5, close 5 of the short lot, **realized = +$15**
> * bot buys 5 more AAPL from Alpaca @ $102 → net = 0, close the rest, **realized = +$5**
> * onchain user buys 8 tAAPL @ $104 → net = −8, open short lot 8 @ $104
> * bot buys 3 AAPL from Alpaca @ $104 → net = −5, **realized = $0**
> 
> End state: net = −5, one open lot of 5 short @ $104, **realized = $20**.
> 
> Then add the unrealized side: if the current Alpaca mark for AAPL is $106, we're short 5 shares at basis $104, so closing now would cost us $10. **Unrealized = −$10. Total PnL = $10.**
> 
> That's the whole calculation. No threshold, no parentage, no "matched vs unmatched offchain", no two-stream interleaving. And the end-state inventory the replay produces should equal `position_view.net_position` exactly. If it doesn't, that's a real bug to chase, because both are deterministic functions of the same event stream.
> 
> ## The buckets become quality metrics, not PnL
> The counter-trade / directional / netting split is genuinely useful operationally ("are we capturing the spread cleanly, or eating drift while waiting to hedge"), but it's not PnL attribution. It's two separate quality questions, each answerable from the data we already have:
> 
> * **Spread captured**: for each onchain fill, compare its price to the nearest broker price for the same symbol in time. Compute the per-share gap (sign-adjusted for direction), aggregate as volume-weighted average per time window, report in basis points. Tells us what gross spread the bot is earning per round-trip without pretending to pair specific trades. Drift cost during the hedging lag is captured by the next metric, not this one.
> * **Directional risk**: integrate `|net_position|` over time per symbol (share-hours of exposure carried), and report the median time from an onchain fill until `|net|` falls back below `ExecutionThreshold`. Tells us how much exposure the bot is sitting on while it waits for the threshold trigger to fire.
> 
> These belong next to the PnL numbers, not inside them.
> 
> ## Concrete question
> Do the three PnL buckets matter to whoever will look at this, or were they framed that way because per-trade attribution felt more granular than position-level realized + unrealized? If the latter, simplifying is pure upside: half of `sql-source.ts` deletes, the eventual backend `/pnl` endpoint becomes obvious (we already emit position events and have marks in `position_view`), and the SPEC.md write-up is half a page instead of a multi-page attribution model.


Thank you for the review.

Imo there is no fundamental “model mismatch” here, and the current model is compatible with how the bot operates.

The bot trades against aggregate per-symbol position, while the PnL calculation replays the same fill stream through a per-symbol FIFO inventory ledger. The lot mapping is an accounting/attribution overlay, not a claim that the live bot persists causal parentage between specific on-chain fills and broker fills.

I think the disagreement is mostly around framing, apart from the real issue that off-chain residuals / overshoots need to remain part of the PnL accounting if they occur. Apologies for the wording in the PR description that may have contributed to the misunderstanding.

The implementation issues highlighted are real and should now be addressed.

JuaniRios (approved):

0xgleb (approved):

PR #704, opened 2026-05-28
ci: cache nix source closure before matrix

JuaniRios:
Closes [RAI-707](https://linear.app/makeitrain/issue/RAI-707)

## What

Adds a `nix source cache` warmup job to CI and makes backend, dashboard, and hooks jobs depend on it before running their Nix-based validation steps.

## Why

CI has been repeatedly hanging during Nix/Git source downloads for Solidity/Nix inputs. The failures are stochastic: rerunning only the failed job usually passes, but rerunning the whole workflow often causes a different matrix leg to stall while fetching another source repository.

## How

The new warmup job fetches and realizes the shared Nix source/ABI closure once, then stores it in an Actions-backed local Nix binary cache under `.tmp/nix-ci-cache`.

Downstream jobs restore that cache and import the paths before entering `nix develop`, so the matrix no longer fans out simultaneous live Git fetches for the same flake inputs and ABI derivations.

This keeps generated ABI JSON dynamic and uncommitted. It also configures Git low-speed limits so genuinely stalled downloads fail fast instead of hanging indefinitely.

## Testing

- CI passed on this PR:
  - `nix source cache`
  - backend check
  - backend test 1-4
  - backend clippy
  - dashboard
  - hooks
  - Graphite mergeability check

## Screenshots

Not applicable.

## Anything else

The cache key is tied to runner OS/arch plus `flake.lock`, `flake.nix`, and `nix/*.nix`, so cache invalidation follows the Nix source/ABI configuration.

---
[...]

PR #705, opened 2026-05-29
fix: restore offchain inventory polling

JuaniRios:
## What

Restores offchain inventory polling by tracking Alpaca-held USDC from the broker positions endpoint instead of the Alpaca wallet endpoint.

This PR:
- reads Alpaca `USDCUSD` crypto positions from the broker positions endpoint and stores them as `Inventory.alpaca_usdc`, using the position's total `qty` (not `qty_available`)
- emits `InventorySnapshotCommand::AlpacaUsdc` from `poll_offchain()`
- excludes `USDCUSD` from equity inventory by symbol, and filters other non-equity Alpaca positions out before symbol parsing
- removes wallet-balance polling from inventory; Alpaca wallet APIs remain for deposit/withdrawal flow plumbing
- makes inflight equity polling a hard precondition in `poll_and_record()`, while onchain, wallet, and offchain polling each fail independently (log and continue)

## Why

Prod offchain inventory polling was failing because the bot tried to parse Alpaca wallet endpoint responses as account balances, but that endpoint does not return the balance shape the poller expected. As a result, offchain inventory stopped updating.

The USDC value we need is the Alpaca-held `USDCUSD` position created after USD-to-USDC conversion and before withdrawal to Ethereum, not the wallet endpoint response. We read its total `qty` rather than `qty_available` so a USDC amount locked by an in-flight withdrawal is still counted as held at Alpaca until the withdrawal settles.

Inventory polling also needs partial freshness: an onchain or wallet polling failure should not prevent broker/offchain inventory from updating during the same tick. Inflight equity is the exception -- it must be polled first because balance snapshots drive `check_and_trigger_equity`, whose dedup guard is `has_inflight()`. If inflight data is stale, a balance snapshot could trigger a duplicate operation for a request the provider already has pending, so an inflight poll failure aborts the whole tick.

## How

`st0x_execution::Inventory` now includes optional `alpaca_usdc` (`None` when the executor does not model an Alpaca USDC venue; `Some(Usdc::ZERO)` for a reporting executor with no balance).

`fetch_inventory()` in the Alpaca broker implementation:
- captures the `USDCUSD` position's total `qty` as Alpaca-held USDC, and fails fast with `MissingPositionQuantity` if a `USDCUSD` position omits `qty` (rather than silently reporting zero)
- excludes `USDCUSD` from equity by symbol in `is_non_equity()`, so it never reaches equity/vault reconciliation even if Alpaca omits `asset_class`/`exchange` metadata
- ignores other non-equity Alpaca positions instead of parsing them as equity symbols
- returns zero USDC when no `USDCUSD` position is present

`InventoryPollingService::poll_offchain()` emits the Alpaca USDC snapshot event alongside offchain equity, USD, buying power, and withdrawable cash.

`poll_and_record()` polls inflight equity first; on failure it logs and aborts the tick before emitting any balance snapshot. The onchain, wallet, and offchain groups then each warn-and-continue on failure so one venue outage does not stall the others.

The Alpaca wallet `get_asset_balance()` path was removed because wallet endpoints are not the source of this inventory value.

## Testing

Added/updated tests for:
- parsing the real Alpaca `USDCUSD` position shape
- reading total `qty` (not `qty_available`) for USDC, and erroring when `qty` is absent
- reporting zero for a fully-locked `USDCUSD` position
- excluding `USDCUSD` from equity even when `asset_class`/`exchange` are absent
- ignoring other Alpaca crypto positions / non-equity asset classes without breaking equity inventory
- emitting / skipping Alpaca USDC based on executor inventory
- aborting the tick (no balance snapshots) when inflight polling fails
- continuing to emit offchain inventory after onchain or wallet polling failure
- preserving lower-level error tests by calling `poll_onchain()` / `poll_wallets()` directly

Verified locally (`nix develop`): `cargo check`, targeted `cargo nextest run` for the execution `fetch_inventory*` and hedge `poll_*` tests (19 + 35 passing), and `cargo clippy`/`cargo fmt` on the changed crates.

## Screenshots

Not applicable.

## Anything else

Stacked on #704 (`ci-cache-nix-source-closure`).

Follow-up deferred to [RAI-703](https://linear.app/makeitrain/issue/RAI-703): during a degraded tick (onchain/wallet poll fails, offchain succeeds), fresh offchain balance snapshots can still trigger rebalancing checks that read stale counterpart-venue data. Tracked separately as it concerns the deliberate partial-freshness design.

### DIFF
--- a/crates/execution/src/alpaca_broker_api/mock_api.rs
+++ b/crates/execution/src/alpaca_broker_api/mock_api.rs
@@ -749,9 +749,9 @@ fn register_positions_endpoint(server: &MockServer, state: &Arc<Mutex<MockState>
         when.method(GET)
             .path(format!("/v1/trading/accounts/{TEST_ACCOUNT_ID}/positions"));
         then.respond_with(move |_request: &HttpMockRequest| {
-            let result: Result<Vec<Value>, rain_math_float::FloatError> = {
+            let result = (|| -> Result<Vec<Value>, rain_math_float::FloatError> {
                 let state = lock(&state);
-                state
+                let mut positions = state
                     .account
                     .positions
                     .values()
@@ -761,7 +761,7 @@ fn register_positions_endpoint(server: &MockServer, state: &Arc<Mutex<MockState>
                             .lt(Float::from_raw(alloy::primitives::B256::ZERO))?;
                         let side = if is_neg { "short" } else { "long" };
                         let abs_qty = pos.quantity.abs()?;
-                        Ok(json!({
+                        Ok::<Value, rain_math_float::FloatError>(json!({
                             "symbol": pos.symbol,
                             "qty_available": format_float_with_fallback(&abs_qty),
                             "market_value": format_float_with_fallback(&pos.market_value),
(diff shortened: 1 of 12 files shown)


JuaniRios:
* **#705** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/705?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/705?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#704** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/704?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=705).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor (approved):

alastairong1 (approved):

0xgleb (approved):

PR #712, opened 2026-05-29
feat: TransferUsdcToHedging apalis job (Base->Alpaca)

0xgleb:
## What

Closes [RAI-388](https://linear.app/makeitrain/issue/RAI-388/convert-base-to-alpaca-usdc-transfer-to-transferusdctohedging-apalis). Converts `CrossVenueCashTransfer::execute_base_to_alpaca` from a one-off `tokio::spawn` driven by `Rebalancer` into a state-resumable apalis job. `check_and_trigger_usdc` enqueues `TransferUsdcToHedging { id, amount }` directly; the Rebalancer mpsc channel is no longer the path for Base->Alpaca.

This is the **implementation** that satisfies the failing test in the parent PR #714. Per [docs/ttdd.md](https://github.com/ST0x-Technology/st0x.liquidity/blob/master/docs/ttdd.md), the test sits downstack of this branch so reviewers can confirm the hypothesis was failing on #714 alone and that the diff in this PR is what turns it green.

## Why

A crash mid-transfer (after Alpaca withdrawal, before CCTP mint) used to orphan in-flight capital -- there was no persistent record of "this `UsdcRebalanceId` is in flight" and nothing to resume the lifecycle on restart. The apalis queue persists the job row, and the same `perform()` that starts a new transfer also resumes a stalled one. The end-to-end proof of that durability claim is the `interrupted_usdc_base_to_alpaca_resumes_after_restart` test in #714, which is expected to fail on #714 in isolation and pass once this PR lands on top.

## How

- Extract `resume_base_to_alpaca(id, amount)` on `CrossVenueCashTransfer`. Loads the `UsdcRebalance` aggregate via `Store::load` and dispatches on state: `None` -> start fresh; `Withdrawing` -> `ConfirmWithdrawal` + continue; `WithdrawalComplete` -> CCTP burn; `Bridging`/`Attested` -> re-poll attestation + mint; `Bridged` -> initiate Alpaca deposit; `DepositInitiated` -> poll Alpaca; `DepositConfirmed` (BaseToAlpaca) -> USDC->USD conversion; `Converting` -> emit `FailConversion` (broker order ID isn't persisted, so re-issue is unsafe); terminal failure states -> propagate `PreviouslyFailedAggregate`.
- New `TransferUsdcToHedging` apalis job + `ResumeBaseToAlpaca` trait so the conductor's ctx erases the wallet `Chain` generic.
- `RebalancerServices::spawn` now returns a `SpawnedRebalancer` struct (handle, shutdown token, trait-erased resume handle) instead of an anonymous tuple.
- The `usdc_in_progress` guard is cleared event-driven only when the aggregate reaches a terminal state (success or a recorded failure). An indeterminate mid-flight failure (e.g. stalled at `WithdrawalSubmitting`/`BridgingSubmitting`) deliberately keeps the guard latched so automation does not re-arm a fresh transfer on top of a partial one -- the apalis retry re-enters the scan path rather than re-issuing.
- **Crash-safe resume of in-flight rows.** apalis's `fetch_next` only picks `Pending`/retryable-`Failed` rows, and on a quick restart its orphan sweep never fires (the deterministic worker name keeps the dead worker's heartbeat fresh), so a row left `Running` when the process died is never re-driven -- and the enqueue dedup keyed off that row then suppresses all future Base->Alpaca transfers. `Conductor::start` now resets orphaned `Running`/`Queued` `TransferUsdcToHedging` rows to `Pending` at startup, before the monitor spawns (where every such row is by definition orphaned), so the worker re-drives them. `Failed` rows (retries exhausted) are left latched for operator reconciliation.
- **Per-attempt timeout.** `perform()` bounds `resume_base_to_alpaca` with a configurable wall-clock timeout, so a hung RPC fails the attempt (and retries) instead of wedging the single-concurrency worker forever.
- The recovery scan helpers (`find_recent_burn`/`find_recent_withdrawal`) live on the `Bridge` trait, not the concrete `CctpBridge`, so callers depend only on the bridge abstraction.

## Testing

- The e2e durability claim is asserted in #714 (`interrupted_usdc_base_to_alpaca_resumes_after_restart`): it crashes the bot mid-bridge, restarts, and asserts the transfer resumes to `ConversionConfirmed` without re-issuing the vault withdrawal. Fails on #714 alone, passes on this branch.
- `requeue_orphaned` resets only this queue's `Running`/`Queued` rows (and clears their lock), leaving `Pending`/`Failed`/`Done` and other job types untouched.
- `perform()` returns the `Timeout` error when the resume future exceeds the configured per-attempt bound.
- Job-level: failure clears `usdc_in_progress`, success preserves it, payload `id`/`amount` reach the transfer.
- Resume routing: terminal failure -> `PreviouslyFailedAggregate`; `ConversionComplete` -> no-op (no service mocks needed); `Bridged` -> reaches `ConversionComplete` without re-burn (no CCTP mocks; would surface as anvil error if exercised).
- Trigger and integration tests updated: `check_and_trigger_usdc` enqueues a `TransferUsdcToHedging` row instead of sending `UsdcBaseToAlpaca` via mpsc.

## Anything else

- New required config field `rebalancing.transfer_attempt_timeout_secs` (per-attempt hang backstop, distinct from the whole-transfer `transfer_timeout_secs` stall reaper). Set to `3600` in the example/staging/prod configs -- a coarse bound that exceeds the worst-case healthy attempt (the Alpaca deposit poll alone is bounded at 1800s), so it only trips on a true hang.
- The mpsc path still carries `UsdcAlpacaToBase` until #713 (RAI-390) lands. The unused `TriggeredOperation::UsdcBaseToAlpaca` variant is kept for now to keep the diff focused; removing it depends on #713 also dropping its sibling.

## Summary by CodeRabbit

* **New Features**
  * Durable, persisted TransferUsdcToHedging job and optional worker to perform Base→Alpaca USDC transfers.

* **Reliability Improvements**
  * Crash-safe resume for Base→Alpaca flows with helpers to detect/adopt prior on-chain burns/withdrawals and per-attempt timeouts.
  * Queue sweep to reset orphaned in-flight jobs on startup.

* **Behavior Changes**
  * USDC triggers now enqueue persisted jobs instead of using an in-memory channel.

* **Configuration**
  * New transfer_attempt_timeout_secs setting (default 3600s).

* **Documentation**
  * Added guidance on orphaned job behavior and restart implications.

* **Tests**
  * Updated unit and E2E tests for persisted-job and resume semantics.

### DIFF
--- a/src/conductor/job.rs
+++ b/src/conductor/job.rs
@@ -170,6 +170,36 @@ impl<Task: Serialize + DeserializeOwned + Send + Sync + Unpin + 'static> JobQueu
             );
         }
     }
+
+    /// Resets this queue's in-flight rows (`Running`/`Queued`) back to
+    /// `Pending` so the apalis monitor re-drives them, and returns the number
+    /// of rows reset.
+    ///
+    /// Must only be called at startup, BEFORE the apalis monitor spawns: at
+    /// that point no worker is alive to legitimately own a `Running` row, so
+    /// every such row is necessarily orphaned by a previous process that died
+    /// mid-job. apalis's own orphan recovery cannot rescue these on a quick
+    /// restart -- it re-enqueues a locked row only once the owning worker's
+    /// heartbeat ages past `reenqueue_orphaned_after` (5 min default), but the
+    /// worker name is deterministic across restarts, so a fresh process
+    /// re-registers the same worker id and keeps refreshing its heartbeat,
+    /// and the orphan is never aged out. Resetting the row here closes that
+    /// gap. `Failed` rows (retries exhausted) are deliberately left untouched
+    /// so a latched job awaiting operator reconciliation is not re-driven on
+    /// every restart. `attempts` is preserved because a crash is not a failed
+    /// attempt against the retry budget.
+    pub(crate) async fn requeue_orphaned(&self) -> Result<u64, sqlx::Error> {
+        let job_type = std::any::type_name::<Task>();
+        let result = sqlx::query(
+            "UPDATE Jobs SET status = 'Pending', lock_by = NULL, lock_at = NULL \
+             WHERE job_type = ? AND status IN ('Running', 'Queued')",
+        )
+        .bind(job_type)
+        .execute(self.pool())
+        .await?;
+
+        Ok(result.rows_affected())
+    }
 }
 
 /// A persistent, retryable unit of work backed by apalis storage.
@@ -312,6 +342,7 @@ pub enum JobKind {
     SeedVaultRegistry,
     WrappedEquityRecovery,
     CheckPositions,
+    TransferUsdcToHedging,
 }
 
 /// Job execution error. Wraps the concrete `Job::Error` type at
@@ -347,6 +378,7 @@ pub struct FailureInjector {
     seed_vault_registry: Arc<Mutex<InjectionState>>,
     wrapped_equity_recovery: Arc<Mutex<InjectionState>>,
     check_positions: Arc<Mutex<InjectionState>>,
+    transfer_usdc_to_hedging: Arc<Mutex<InjectionState>>,
 }
 
 #[cfg(any(test, feature = "test-support"))]
@@ -380,6 +412,7 @@ impl FailureInjector {
             seed_vault_registry: Arc::new(Mutex::new(InjectionState::Idle)),
             wrapped_equity_recovery: Arc::new(Mutex::new(InjectionState::Idle)),
             check_positions: Arc::new(Mutex::new(InjectionState::Idle)),
+            transfer_usdc_to_hedging: Arc::new(Mutex::new(InjectionState::Idle)),
         }
     }
 
@@ -425,6 +458,7 @@ impl FailureInjector {
             JobKind::SeedVaultRegistry => &self.seed_vault_registry,
             JobKind::WrappedEquityRecovery => &self.wrapped_equity_recovery,
             JobKind::CheckPositions => &self.check_positions,
+            JobKind::TransferUsdcToHedging => &self.transfer_usdc_to_hedging,
         };
 
         match mutex.lock() {
@@ -796,6 +830,97 @@ mod tests {
                 "pending".to_string(),
                 "running".to_string()
             ]
+        );
+    }
+
+    async fn insert_locked_job(
+        pool: &SqlitePool,
+        id: &str,
+        job_type: &str,
+        status: &str,
+        lock_by: Option<&str>,
+    ) {
+        sqlx::query(
+            "INSERT INTO Jobs \
+             (job, id, job_type, status, attempts, max_attempts, run_at, priority, lock_by, lock_at) \
+             VALUES (?, ?, ?, ?, 0, 25, 0, 0, ?, ?)",
+        )
+        .bind(vec![0_u8])
+        .bind(id)
+        .bind(job_type)
+        .bind(status)
+        .bind(lock_by)
+        .bind(lock_by.map(|_| 0_i64))
+        .execute(pool)
+        .await
+        .unwrap();
+    }
+
+    async fn status_of(pool: &SqlitePool, id: &str) -> String {
+        sqlx::query_scalar::<_, String>("SELECT status FROM Jobs WHERE id = ?")
+            .bind(id)
+            .fetch_one(pool)
+            .await
+            .unwrap()
+    }
+
+    async fn lock_by_of(pool: &SqlitePool, id: &str) -> Option<String> {
+        sqlx::query_scalar::<_, Option<String>>("SELECT lock_by FROM Jobs WHERE id = ?")
+            .bind(id)
+            .fetch_one(pool)
+            .await
+            .unwrap()
+    }
+
+    async fn insert_worker(pool: &SqlitePool, id: &str) {
+        sqlx::query(
+            "INSERT INTO Workers (id, worker_type, storage_name) VALUES (?, 'test', 'test')",
+        )
+        .bind(id)
+        .execute(pool)
+        .await
+        .unwrap();
+    }
+
+    #[tokio::test]
+    async fn requeue_orphaned_resets_only_in_flight_rows_of_this_queue() {
+        let pool = setup_test_db().await;
+        let job_type = std::any::type_name::<TestJob>();
+
+        // A previous process died holding the lock on these in-flight rows.
+        insert_worker(&pool, "dead-worker").await;
+        insert_locked_job(&pool, "running", job_type, "Running", Some("dead-worker")).await;
+        insert_locked_job(&pool, "queued", job_type, "Queued", Some("dead-worker")).await;
+        insert_locked_job(&pool, "pending", job_type, "Pending", None).await;
+        insert_locked_job(&pool, "failed", job_type, "Failed", Some("dead-worker")).await;
+        insert_locked_job(&pool, "done", job_type, "Done", Some("dead-worker")).await;
+        // A Running row of a different queue's job type must be left alone.
+        insert_locked_job(&pool, "other", "other::Job", "Running", Some("dead-worker")).await;
+
+        let queue = JobQueue::<TestJob>::new(&pool);
+        let reset = queue.requeue_orphaned().await.unwrap();
+
+        assert_eq!(reset, 2, "only this queue's Running + Queued rows reset");
+        assert_eq!(status_of(&pool, "running").await, "Pending");
+        assert_eq!(status_of(&pool, "queued").await, "Pending");
+        assert_eq!(
+            lock_by_of(&pool, "running").await,
+            None,
+            "lock cleared so the apalis monitor re-picks the row",
+        );
+        assert_eq!(lock_by_of(&pool, "queued").await, None);
+
+        assert_eq!(status_of(&pool, "pending").await, "Pending");
+        assert_eq!(
+            status_of(&pool, "failed").await,
+            "Failed",
+            "a latched failure awaiting operator reconciliation is not re-driven",
+        );
+        assert_eq!(status_of(&pool, "done").await, "Done");
+        assert_eq!(
+            status_of(&pool, "other").await,
+            "Running",
+            "another queue's in-flight row is untouched",
         );
     }
 }
--- a/src/rebalancing/trigger/mod.rs
+++ b/src/rebalancing/trigger/mod.rs
@@ -13,6 +13,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::Duration;
 use tokio::sync::{Mutex, RwLock, mpsc};
 use tracing::{debug, error, trace, warn};
+use uuid::Uuid;
 
 use rain_math_float::Float;
 use st0x_config::{AssetsConfig, OperationMode};
@@ -22,6 +23,7 @@ use st0x_event_sorcery::{
 use st0x_execution::{FractionalShares, Positive, SharesBlockchain, SharesConversionError, Symbol};
 use st0x_finance::{HasZero, Usd, Usdc};
 
+use crate::conductor::job::QueuePushError;
 use crate::equity_redemption::{
     EquityRedemption, EquityRedemptionCommand, EquityRedemptionEvent, RedemptionAggregateId,
 };
@@ -32,6 +34,7 @@ use crate::inventory::{
     Operator, PendingRequestOwnership, PendingRequestOwnershipSnapshot, TransferOp, Venue,
 };
 use crate::position::{Position, PositionEvent};
+use crate::rebalancing::usdc::{TransferUsdcToHedging, TransferUsdcToHedgingJobQueue};
 use crate::tokenized_equity_mint::{
     IssuerRequestId, TokenizedEquityMint, TokenizedEquityMintCommand, TokenizedEquityMintEvent,
 };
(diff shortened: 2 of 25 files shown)


0xgleb:
* **#723** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/723?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#724](https://github.com/ST0x-Technology/st0x.liquidity/pull/724) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/724?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#725](https://github.com/ST0x-Technology/st0x.liquidity/pull/725) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/725?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#712** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/712?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/712?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=712).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/rebalancing/usdc/manager.rs
```diff
@@ -601,6 +601,229 @@ impl<Chain: Wallet> CrossVenueCashTransfer<Chain> {
...
+        let state = self.cqrs.load(id).await?;
+
+        info!(
+            target: "rebalance",
+            ?state,
+            "Resuming Base->Alpaca transfer from aggregate state",
+        );
+
+        match state {
+            None => self.execute_base_to_alpaca(id, amount).await,
```
JuaniRios, line 641: critical: this resume path can double-spend. Each phase performs its irreversible on-chain action before persisting the event: withdraw_from_vault calls raindex.withdraw_usdc then sends Initiate, and execute_cctp_burn_on_base burns then sends InitiateBridging. So a crash after withdraw_usdc lands the aggregate at None and this arm re-withdraws, and a crash after burn lands it at WithdrawalComplete and that arm re-burns. withdraw_usdc has no idempotency guard, so this drains the vault twice or burns USDC twice with at most one mint, which is worse than the orphaned-capital case this is meant to fix. Persist intent (Initiate/InitiateBridging with the tx hash) before the on-chain call so resume can detect an already-submitted tx, the way the conversion phase records InitiatePostDepositConversion before placing the order.

src/rebalancing/usdc/manager.rs
```diff
@@ -601,6 +601,229 @@ impl<Chain: Wallet> CrossVenueCashTransfer<Chain> {
...
+
+            Some(UsdcRebalance::WithdrawalComplete {
+                direction, amount, ..
+            }) => {
+                Self::require_base_to_alpaca(id, &direction)?;
+                self.continue_from_withdrawal_complete(id, amount).await
+            }
+
+            Some(
+                UsdcRebalance::Bridging {
```
JuaniRios: should fix: resuming from Attested is permanently unrecoverable. This arm routes Attested through continue_from_bridging, which calls poll_attestation_for_base_burn, and that unconditionally sends ReceiveAttestation. The aggregate's transition_receive_attestation accepts that command only from Bridging and returns InvalidCommand from Attested, so a crash after the attestation is persisted but before the mint confirms fails every retry and fail-stops with the USDC already burned. The doc comment blames the Circle API, but the failure is the rejected command. Split the Attested arm out: re-poll Circle for the AttestationResponse without re-emitting ReceiveAttestation, then go straight to mint.

src/rebalancing/usdc/job.rs
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    async fn perform(&self, ctx: &TransferUsdcToHedgingCtx) -> Result<Self::Output, Self::Error> {
+        let result = ctx
+            .transfer
+            .resume_base_to_alpaca(&self.id, self.amount)
+            .await;
+
+        if let Err(error) = &result {
+            ctx.usdc_in_progress.store(false, Ordering::SeqCst);
```
JuaniRios: should fix: clearing usdc_in_progress on every Err is too aggressive given the worker has RetryPolicy::retries(3). A transient failure drops the guard while the job is still scheduled to retry, and once retries are exhausted the row is terminal-Failed, the flag is already false, and the dedup query (which only counts Pending/Queued/Running) no longer matches, so the next imbalance check enqueues a fresh job with a new UsdcRebalanceId on top of a partially-completed transfer. Clear the flag only on terminal failure (thread the apalis Attempt in, or let the terminal-failure handler do it), and keep it latched for indeterminate states so automation does not re-arm without an operator.

src/rebalancing/usdc/manager.rs
```diff
@@ -601,6 +601,229 @@ impl<Chain: Wallet> CrossVenueCashTransfer<Chain> {
...
+                warn!(
+                    target: "rebalance",
+                    %id,
+                    "resume_base_to_alpaca called on AlpacaToBase DepositConfirmed; \
+                     treating as no-op terminal",
+                );
+                Ok(())
+            }
+
+            Some(UsdcRebalance::Converting { .. }) => {
```
JuaniRios, line 756: should fix: resuming from Converting unconditionally emits FailConversion, but the order is placed after InitiatePostDepositConversion, so if the crash happened after the Alpaca order filled this records ConversionFailed for a conversion that actually succeeded. Combined with the usdc_in_progress clear, a fresh end-to-end transfer can then re-arm on top of it, and there is no broker order ID surfaced for the manual reconciliation the error asks for. Persist the Alpaca order ID in ConversionInitiated so resume can poll order status and pick ConfirmConversion vs FailConversion deterministically.

src/rebalancing/trigger/mod.rs
```diff
@@ -1942,14 +1953,101 @@ impl RebalancingService {
...
+    /// event would let the next imbalance check enqueue a second job for
+    /// the same imbalance.
+    async fn enqueue_transfer_usdc_to_hedging(&self, amount: Usdc) -> bool {
+        let queue = self.transfer_usdc_to_hedging_queue.clone();
+        let job_type = std::any::type_name::<TransferUsdcToHedging>();
+
+        let pending_count: Result<(i64,), _> = sqlx::query_as(
+            "SELECT COUNT(*) FROM Jobs \
+             WHERE job_type = ? \
+             AND status IN ('Pending', 'Queued', 'Running')",
```
JuaniRios: should fix: this dedup check blocks a new enqueue whenever any TransferUsdcToHedging row is Pending/Queued/Running, so a single job stuck retrying transient errors silently suppresses every future Base->Alpaca enqueue for any imbalance, with only a debug log and no staleness bound. Since the comment names this query as the sole cross-restart guard, a false-positive block can starve the hedge side of capital indefinitely. Log at warn with the row id and age, and add a staleness threshold (or scope dedup to the specific imbalance).

src/rebalancing/trigger/mod.rs
```diff
@@ -1942,14 +1953,101 @@ impl RebalancingService {
             return;
         };
 
-        if !self.try_send_operation(&operation, "usdc") {
+        let dispatched = match &operation {
+            TriggeredOperation::UsdcBaseToAlpaca { amount } => {
+                self.enqueue_transfer_usdc_to_hedging(*amount).await
+            }
+            _ => self.try_send_operation(&operation, "usdc"),
```
JuaniRios: should fix: this wildcard _ arm violates the project's exhaustive-match convention. A future TriggeredOperation variant that should also go through the job queue would silently route through mpsc here. The enum has only four variants, so enumerate the non-queue ones explicitly (Mint | Redemption | UsdcAlpacaToBase) so adding a variant forces a compile error.

src/rebalancing/usdc/manager.rs
```diff
@@ -2409,4 +2641,236 @@ mod tests {
...
+                fee_collected: usdc("0.01"),
+            },
+        )
+        .await
+        .unwrap();
+        (burn_tx, mint_tx)
+    }
+
+    #[tokio::test]
+    async fn resume_base_to_alpaca_from_terminal_failure_returns_previously_failed_error() {
```
JuaniRios, line 2928: should fix: the resume dispatch has about ten arms but the new tests only cover ConversionComplete, terminal failure, and Bridged. The arms that re-execute irreversible side-effects (None-after-withdrawal, Withdrawing, WithdrawalComplete, Bridging/Attested, DepositInitiated, Converting) and the direction guard are untested, and AGENTS.md mandates coverage for new logic. A test staging WithdrawalComplete or Attested and asserting via call-count mocks that the burn/attestation is invoked zero times would catch the double-burn and the Attested-unrecoverable bugs above.

src/rebalancing/usdc/job.rs
```diff
@@ -0,0 +1,254 @@
...
+        make_job().perform(&ctx).await.unwrap();
+        assert!(
+            usdc_in_progress.load(Ordering::SeqCst),
+            "usdc_in_progress should remain set on success; the next event clears it"
+        );
+        assert_eq!(*stub.call_count.lock().unwrap(), 1);
+    }
+
+    #[tokio::test]
+    async fn perform_forwards_id_and_amount_to_transfer() {
```
JuaniRios: minor: this test asserts that the id/amount captured by the stub equal the job's own fields, which just exercises perform's two-line plumbing that the compiler already guarantees. It catches no bug. Consider deleting it; the flag-clearing tests already cover perform's actual behavior (and their call_count == 1 assertions are trivially true and can go too).

src/rebalancing/usdc/manager.rs
```diff
@@ -601,6 +601,229 @@ impl<Chain: Wallet> CrossVenueCashTransfer<Chain> {
...
+        let mint_receipt = self
+            .execute_cctp_mint_on_ethereum(id, attestation_response)
+            .await?;
+        self.continue_from_bridged(id, u256_to_usdc(mint_receipt.amount)?, mint_receipt.tx)
+            .await
+    }
+
+    /// Drives the transfer from `Bridged` through to terminal: initiate
+    /// Alpaca deposit, poll, then USDC->USD conversion.
+    async fn continue_from_bridged(
```
JuaniRios, line 875: minor: resume from Bridged re-sends InitiateDeposit and is safe only because the CCTP mint deposits directly to the Alpaca address, so re-sending records intent without moving funds. That invariant is load-bearing for crash safety but only stated in prose, and unlike the DepositInitiated arm there is no guard here. Worth a test pinning that resume from Bridged makes no fund-moving deposit call, plus a comment documenting the mint-deposits-directly assumption.

findolor (approved):

JuaniRios (changes requested):

src/rebalancing/trigger/mod.rs
0xgleb: added a warn + staleness threshold

src/rebalancing/trigger/mod.rs
0xgleb: fixed

src/rebalancing/trigger/usdc.rs
```diff
@@ -122,7 +122,11 @@ impl UsdcRebalanceStage {
             Bridged { .. } => Some(Self::Bridged),
             DepositInitiated { .. } => Some(Self::DepositInitiated),
             DepositConfirmed { .. } => Some(Self::DepositConfirmed),
-            ConversionConfirmed { .. }
+            // Transient intent markers: immediately superseded by Initiated /
+            // BridgingInitiated, so they emit no distinct progress stage.
+            WithdrawalSubmitting { .. }
+            | BridgingSubmitting { .. }
+            | ConversionConfirmed { .. }
```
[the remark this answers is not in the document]
0xgleb: fixed

src/rebalancing/usdc/job.rs
0xgleb: fixed, clearing is event driven now

src/rebalancing/usdc/manager.rs
0xgleb: intents persisted ✅

src/conductor/job.rs
```diff
@@ -798,4 +832,95 @@ mod tests {
...
+    }
+
+    #[tokio::test]
+    async fn requeue_orphaned_resets_only_in_flight_rows_of_this_queue() {
+        let pool = setup_test_db().await;
+        let job_type = std::any::type_name::<TestJob>();
+
+        // A previous process died holding the lock on these in-flight rows.
+        insert_worker(&pool, "dead-worker").await;
+        insert_locked_job(&pool, "running", job_type, "Running", Some("dead-worker")).await;
```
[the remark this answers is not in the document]
0xgleb: setup_test_db already does that

src/rebalancing/usdc/manager.rs
0xgleb: fixed upstack in #725

src/rebalancing/usdc/manager.rs
0xgleb: fixed upstack in #732

JuaniRios (approved):

PR #339, opened 2026-02-27
feat: benchmark Alpaca tokenization pipeline latency

0xgleb:
## Motivation

Closes [RAI-60](https://linear.app/makeitrain/issue/RAI-60/benchmark-alpaca-tokenization-pipeline-latency-for-external-partners)
Closes [#338](https://github.com/ST0x-Technology/st0x.liquidity/issues/338)

External partners need performance data for the Alpaca tokenization pipeline before committing to market maker conversations. We need to measure end-to-end latency for both minting (USD -> tSTOCK) and redemption (tSTOCK -> USD) directions, including fee data, across multiple round trips to capture variance.

### New `alpaca-benchmark` CLI command

Adds a CLI command that runs N round trips of **buy -> mint -> redeem -> sell** for a given equity symbol, measuring tokenization and detokenization latency at each step.

**Usage:**
```
cargo run --bin cli -- alpaca-benchmark -s AAPL -n 10 -q 1 --output report.tsv
```

Each round trip:
1. Buys shares via Alpaca Broker API
2. Mints tokenized equity (tSTOCK) and measures latency
3. Redeems tokenized equity back and measures latency
4. Sells shares to return to neutral position
5. On trip failure, attempts a cleanup sell to avoid accumulating position

**Outputs:**
- **TSV report file** -- one row per phase (mint/redeem) with columns: trip, phase, started_at, completed_at, duration_secs, fees, status. Streamed incrementally so partial results survive crashes.
- **CLI summary table** -- min/max/avg/median statistics for mint, redeem, and full round trip durations.

### Supporting changes

- **`Usd` newtype** (`src/threshold.rs`) -- typed USD amounts for fee data, replacing raw `Decimal` in `TokenizationRequest.fees`.
- **`TokenizedSymbol::from_base()`** (`src/onchain/io.rs`) -- construct a tokenized symbol from its underlying base symbol, used for equity config lookup.
- **`TokenizationRequestStatus::Display`** -- lowercase display impl for report serialization.
- **CLI refactor** -- removed the `SimpleCommand`/`ProviderCommand` classification layer that duplicated every command variant. The match now dispatches directly, connecting WebSocket providers lazily only for commands that need them.

### Module structure

```
src/cli/benchmark.rs            -- orchestration (round trip loop, order polling, mint/redeem measurement)
src/cli/benchmark/measurement.rs -- Measurement struct, RoundTripPhase enum, serde serialization
[...]

PR #713, opened 2026-05-29
feat: TransferUsdcToMarketMaking apalis job (Alpaca->Base)

0xgleb:
## What

Closes [RAI-390](https://linear.app/makeitrain/issue/RAI-390/convert-alpaca-to-base-usdc-transfer-to-transferusdctomarketmaking). Adds the Alpaca->Base counterpart to #712's `TransferUsdcToHedging`: a state-resumable `TransferUsdcToMarketMaking` apalis job, and switches `check_and_trigger_usdc` to enqueue both USDC directions via apalis instead of routing through the Rebalancer mpsc channel.

This PR is purely additive. The now-dormant `CrossVenueTransfer` impls on `CrossVenueCashTransfer`, the `Rebalancer` USDC fields, the `TriggeredOperation::UsdcAlpacaToBase`/`UsdcBaseToAlpaca` variants, and the `MockUsdcRebalance` test mock all stay in place. PR #714 removes them in a follow-up.

## Why

A crash mid-transfer in the Alpaca->Base direction used to orphan in-flight capital the same way the Base->Alpaca direction did before #712 (Alpaca withdrawal initiated, USDC sitting in the bot's EOA wallet, no driver to bridge or deposit it). With this PR both directions are driven by persistent apalis jobs that load the `UsdcRebalance` aggregate on retry and resume from the next unfinished phase.

This crash-safe resume covers the post-conversion phases (withdrawal, bridging, deposit). The `Converting` phase is the one exception: the broker order id is not persisted, so a crash after the conversion fills but before `ConfirmConversion` still needs manual reconciliation -- the same residual window accepted for the Base->Alpaca side in #712.

## How

- Extract `resume_alpaca_to_base(id, amount)` on `CrossVenueCashTransfer`, mirroring `resume_base_to_alpaca`. Loads aggregate via `Store::load` and dispatches on the AlpacaToBase-direction state arms.
- Add `TransferUsdcToMarketMaking` apalis job + `ResumeAlpacaToBase` trait + ctx, wired through `RebalancingSchedulers`, `RebalancingService`, `SpawnedRebalancer`, and the conductor builder (worker registration, ctx wiring).
- `check_and_trigger_usdc` now enqueues `TransferUsdcToHedging` for `UsdcBaseToAlpaca` and `TransferUsdcToMarketMaking` for `UsdcAlpacaToBase`. The Rebalancer mpsc path for USDC becomes dead in production.

## Testing

- Job-level: failure clears `usdc_in_progress`, success preserves it, payload `id`/`amount` reach the transfer.
- Resume routing: terminal failure -> `PreviouslyFailedAggregate`; `DepositConfirmed` (AlpacaToBase) -> no-op; `Bridged` -> reaches `DepositConfirmed` without re-burn.
- Trigger and integration tests updated: `check_and_trigger_usdc` now enqueues to the apalis queue. Tests assert on the queue rather than the mpsc receiver.

## Anything else

The Rebalancer still has the dead USDC routing branches and `TriggeredOperation::UsdcAlpacaToBase`/`UsdcBaseToAlpaca` are still produced by `check_imbalance_and_build_operation`. #714 removes the now-unused code and switches `check_imbalance_and_build_operation` to a dedicated `UsdcRebalanceOperation` enum.

## Summary by CodeRabbit

* **New Features**
  * Support for Alpaca→Base (market-making) USDC transfer jobs with dedicated enqueue/worker path.

* **Improvements**
  * USDC rebalancing now enqueues and tracks transfer work persistently for both directions.
  * Enhanced resume/recovery and clearer error handling for USDC transfer flows.

* **Tests**
  * Integration tests updated to assert outcomes via persisted job records rather than internal channels.

* **Chores**
  * Onchain mock and raindex confirmation behavior updated to better simulate deposit/withdraw scenarios.

### DIFF
--- a/src/rebalancing/trigger/mod.rs
+++ b/src/rebalancing/trigger/mod.rs
@@ -34,7 +34,10 @@ use crate::inventory::{
     Operator, PendingRequestOwnership, PendingRequestOwnershipSnapshot, TransferOp, Venue,
 };
 use crate::position::{Position, PositionEvent};
-use crate::rebalancing::usdc::{TransferUsdcToHedging, TransferUsdcToHedgingJobQueue};
+use crate::rebalancing::usdc::{
+    TransferUsdcToHedging, TransferUsdcToHedgingJobQueue, TransferUsdcToMarketMaking,
+    TransferUsdcToMarketMakingJobQueue,
+};
 use crate::tokenized_equity_mint::{
     IssuerRequestId, TokenizedEquityMint, TokenizedEquityMintCommand, TokenizedEquityMintEvent,
 };
@@ -57,6 +60,7 @@ pub(crate) struct RebalancingSchedulers {
     pub(crate) usdc: UsdcRebalancingCheckScheduler,
     pub(crate) wrapped_equity_recovery: WrappedEquityRecoveryJobQueue,
     pub(crate) transfer_usdc_to_hedging: TransferUsdcToHedgingJobQueue,
+    pub(crate) transfer_usdc_to_market_making: TransferUsdcToMarketMakingJobQueue,
 }
 
 impl RebalancingSchedulers {
@@ -66,6 +70,7 @@ impl RebalancingSchedulers {
             usdc: UsdcRebalancingCheckScheduler::new(pool),
             wrapped_equity_recovery: WrappedEquityRecoveryJobQueue::new(pool),
             transfer_usdc_to_hedging: TransferUsdcToHedgingJobQueue::new(pool),
+            transfer_usdc_to_market_making: TransferUsdcToMarketMakingJobQueue::new(pool),
         }
     }
 }
@@ -399,7 +404,9 @@ fn mint_event_tokenization_request_id(
     }
 }
 
-/// Operations triggered by inventory imbalances.
+/// Equity rebalancing operations dispatched through the `Rebalancer` mpsc
+/// channel. USDC rebalances are enqueued directly as apalis jobs from
+/// [`RebalancingService::check_and_trigger_usdc`].
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) enum TriggeredOperation {
     /// Mint tokenized equity (too much offchain).
@@ -439,10 +446,11 @@ pub(crate) struct RebalancingService {
     pub(super) equity_scheduler: EquityRebalancingCheckScheduler,
     pub(super) usdc_scheduler: UsdcRebalancingCheckScheduler,
     pub(super) wrapped_equity_recovery_queue: WrappedEquityRecoveryJobQueue,
-    /// Queue for `TransferUsdcToHedging` apalis jobs that drive Base->Alpaca
-    /// USDC transfers. Enqueued by `check_and_trigger_usdc` instead of being
-    /// dispatched through the `Rebalancer` mpsc channel.
+    /// Queues for the USDC transfer apalis jobs that drive each rebalancing
+    /// direction. `check_and_trigger_usdc` enqueues into one of these
+    /// directly rather than dispatching through the `Rebalancer` mpsc channel.
     pub(super) transfer_usdc_to_hedging_queue: TransferUsdcToHedgingJobQueue,
+    pub(super) transfer_usdc_to_market_making_queue: TransferUsdcToMarketMakingJobQueue,
     /// Tracks symbol/quantity for in-flight mints. The initial `MintRequested`
     /// event carries this data; follow-up events don't.
     mint_tracking: Arc<RwLock<HashMap<IssuerRequestId, MintTracking>>>,
@@ -494,6 +502,7 @@ impl RebalancingService {
             usdc: usdc_scheduler,
             wrapped_equity_recovery: wrapped_equity_recovery_queue,
             transfer_usdc_to_hedging: transfer_usdc_to_hedging_queue,
+            transfer_usdc_to_market_making: transfer_usdc_to_market_making_queue,
         } = schedulers;
         Self {
             config,
@@ -510,6 +519,7 @@ impl RebalancingService {
             usdc_scheduler,
             wrapped_equity_recovery_queue,
             transfer_usdc_to_hedging_queue,
+            transfer_usdc_to_market_making_queue,
             mint_tracking: Arc::new(RwLock::new(HashMap::new())),
             redemption_tracking: Arc::new(RwLock::new(HashMap::new())),
             usdc_tracking: Arc::new(RwLock::new(HashMap::new())),
@@ -2021,9 +2031,10 @@ impl RebalancingService {
             TriggeredOperation::UsdcBaseToAlpaca { amount } => {
                 self.enqueue_transfer_usdc_to_hedging(*amount).await
             }
-            TriggeredOperation::Mint { .. }
-            | TriggeredOperation::Redemption { .. }
-            | TriggeredOperation::UsdcAlpacaToBase { .. } => {
+            TriggeredOperation::UsdcAlpacaToBase { amount } => {
+                self.enqueue_transfer_usdc_to_market_making(*amount).await
+            }
+            TriggeredOperation::Mint { .. } | TriggeredOperation::Redemption { .. } => {
                 self.try_send_operation(&operation, "usdc")
             }
         };
@@ -2036,17 +2047,44 @@ impl RebalancingService {
         guard.defuse();
     }
 
+    /// Returns the oldest non-terminal USDC transfer job in *either* direction
+    /// (the row id and its age in seconds), if any.
+    ///
+    /// The dedupe gate is deliberately direction-independent. Both directions
+    /// move funds through the same vault and market-maker wallet, and the
+    /// in-memory `usdc_in_progress` guard resets on restart, so an in-flight
+    /// transfer in one direction must also suppress enqueuing a transfer in the
+    /// *other* direction. Querying only one job type would, after a restart, let
+    /// an opposite-direction transfer run concurrently against the same funds --
+    /// churning capital and paying CCTP/withdrawal fees twice.
+    async fn in_flight_usdc_transfer(
+        pool: &sqlx::SqlitePool,
+    ) -> Result<Option<(String, i64)>, sqlx::Error> {
+        sqlx::query_as(
+            "SELECT id, CAST(strftime('%s', 'now') AS INTEGER) - run_at AS age_secs \
+             FROM Jobs \
+             WHERE job_type IN (?, ?) \
+             AND status IN ('Pending', 'Queued', 'Running') \
+             ORDER BY run_at ASC \
+             LIMIT 1",
+        )
+        .bind(std::any::type_name::<TransferUsdcToHedging>())
+        .bind(std::any::type_name::<TransferUsdcToMarketMaking>())
+        .fetch_optional(pool)
+        .await
+    }
+
     /// Enqueues a [`TransferUsdcToHedging`] apalis job for a Base->Alpaca
     /// transfer. Generates a fresh `UsdcRebalanceId` at push time so apalis
     /// retries (and bot restarts that re-pick the job row) hit the same
     /// aggregate. Returns `true` on successful enqueue.
     ///
-    /// Before enqueueing, queries the apalis Jobs table for any
-    /// non-terminal `TransferUsdcToHedging` row. The in-memory
-    /// `usdc_in_progress` guard resets on restart, so without this check a
-    /// crash between `queue.push` and the first persisted `UsdcRebalance`
-    /// event would let the next imbalance check enqueue a second job for
-    /// the same imbalance.
+    /// Before enqueueing, [`Self::in_flight_usdc_transfer`] checks the apalis
+    /// Jobs table for any non-terminal USDC transfer in either direction. The
+    /// in-memory `usdc_in_progress` guard resets on restart, so without this
+    /// check a crash between `queue.push` and the first persisted
+    /// `UsdcRebalance` event would let the next imbalance check enqueue a second
+    /// job for the same imbalance.
     async fn enqueue_transfer_usdc_to_hedging(&self, amount: Usdc) -> bool {
         // A non-terminal row in flight longer than this is treated as likely
         // stuck: the suppression is logged at warn (with the row id and age) so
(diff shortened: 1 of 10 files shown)


0xgleb:
* **#719** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/719?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 4  dependent PRs ([#339](https://github.com/ST0x-Technology/st0x.liquidity/pull/339) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/339?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#721](https://github.com/ST0x-Technology/st0x.liquidity/pull/721) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/721?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#722](https://github.com/ST0x-Technology/st0x.liquidity/pull/722) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/722?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 1 other)
* **#713** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/713?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/713?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#718** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/718?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=713).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor (approved):

src/rebalancing/usdc/manager.rs
```diff
@@ -250,6 +250,255 @@ impl<Chain: Wallet> CrossVenueCashTransfer<Chain> {
...
+            }
+
+            Some(
+                UsdcRebalance::Bridging {
+                    direction: RebalanceDirection::AlpacaToBase,
+                    amount,
+                    burn_tx_hash,
+                    ..
+                }
+                | UsdcRebalance::Attested {
```
JuaniRios: critical: resuming from `Attested` always fails. Both the `Bridging` and `Attested` arms route through `continue_alpaca_to_base_from_bridging`, which calls `poll_attestation`, and that helper unconditionally re-sends `ReceiveAttestation`. The aggregate rejects `ReceiveAttestation` from `Attested` (usdc_rebalance.rs:1312 returns `InvalidCommand`), so a crash during the mint leaves the aggregate in `Attested` and every resume errors instead of minting, stranding the burned USDC. Suggest splitting the two arms so that for `Attested` you reconstruct the attestation response without re-emitting the command and go straight to mint. The merged Base->Alpaca side (manager.rs:900) has the same shape, so a fix likely wants to cover both.

src/rebalancing/usdc/manager.rs
```diff
@@ -250,6 +250,255 @@ impl<Chain: Wallet> CrossVenueCashTransfer<Chain> {
...
+    /// Drives an Alpaca->Base transfer from `Bridged` to terminal: vault
+    /// deposit + `ConfirmDeposit`.
+    async fn continue_alpaca_to_base_from_bridged(
+        &self,
+        id: &UsdcRebalanceId,
+        amount_received: Usdc,
+    ) -> Result<(), UsdcTransferError> {
+        let amount_u256 = usdc_to_u256(amount_received)?;
+
+        self.deposit_to_vault(id, amount_u256).await?;
```
JuaniRios, line 643: critical: this can double-deposit on resume. `deposit_usdc` is atomic submit plus confirm (raindex.rs:344), so it lands the on-chain deposit before `InitiateDeposit` is persisted inside `deposit_to_vault`. If the process crashes in that window the aggregate stays in `Bridged`, and resuming re-enters here and submits a second vault deposit for the same funds. The adjacent `DepositInitiated` arm is already guarded with a `confirm_tx` re-check, but this submission window is not. Consider using the existing `submit_deposit4_to_vault` to persist `InitiateDeposit` before confirming, so a crash always lands in `DepositInitiated` where the guard applies.

src/rebalancing/trigger/mod.rs
```diff
@@ -2048,6 +2063,83 @@ impl RebalancingService {
...
+    /// a crash between `queue.push` and the first persisted
+    /// `UsdcRebalance` event would let the next imbalance check enqueue a
+    /// second job for the same rebalance.
+    async fn enqueue_transfer_usdc_to_market_making(&self, amount: Usdc) -> bool {
+        let queue = self.transfer_usdc_to_market_making_queue.clone();
+        let job_type = std::any::type_name::<TransferUsdcToMarketMaking>();
+
+        let pending_count: Result<(i64,), _> = sqlx::query_as(
+            "SELECT COUNT(*) FROM Jobs \
+             WHERE job_type = ? \
```
JuaniRios: critical: this dedupe only checks `TransferUsdcToMarketMaking` rows, but the in-memory `usdc_in_progress` guard resets on restart, so after a restart an in-flight `TransferUsdcToHedging` job will not block enqueuing an opposite-direction transfer here (and vice versa). If inventory has flipped in the meantime, both directions run concurrently against separate aggregates, churning capital and paying fees twice. Suggest querying both job types (`job_type IN (hedging, market_making)`) or using a single direction-independent USDC transfer lock, and adding a regression test with a pending opposite-direction row.

src/rebalancing/usdc/manager.rs
```diff
@@ -250,6 +250,255 @@ impl<Chain: Wallet> CrossVenueCashTransfer<Chain> {
...
+                // records the deposit hash at `InitiateDeposit` time,
+                // but a process restart between that event and the
+                // confirmation can land on a chain where the tx has
+                // been reorged out, dropped, or never mined. Trusting
+                // the persisted state alone would mark the rebalance
+                // complete while the funds aren't actually deposited.
+                let TransferRef::OnchainTx(deposit_tx) = deposit_ref else {
+                    return Err(UsdcTransferError::DepositRefMustBeOnchain { id: id.clone() });
+                };
+                self.raindex.confirm_tx(deposit_tx).await?;
```
JuaniRios: should fix: this arm propagates the `confirm_tx` error without recording a terminal failure, unlike the Base->Alpaca side where `poll_alpaca_deposit_and_confirm` emits `FailDeposit` on failure (manager.rs:1253-1276). A transient not-yet-confirmed response is indistinguishable from a dropped tx here, so a permanently dropped deposit retries forever until the breaker trips, leaving a stuck `DepositInitiated` with funds un-deposited and no terminal state an operator can recognize. Distinguish pending from absent, and on confirmed-absent either re-submit or emit `FailDeposit`.

src/rebalancing/usdc/job.rs
```diff
@@ -116,6 +143,66 @@ impl Job<TransferUsdcToHedgingCtx> for TransferUsdcToHedging {
...
+
+    #[cfg(any(test, feature = "test-support"))]
+    const JOB_KIND: crate::conductor::job::JobKind =
+        crate::conductor::job::JobKind::TransferUsdcToMarketMaking;
+
+    fn label(&self) -> Label {
+        Label::new(format!("TransferUsdcToMarketMaking:{}", self.id))
+    }
+
+    async fn perform(
```
JuaniRios, line 318: should fix: this `perform` has no unit tests, while the `TransferUsdcToHedging` sibling has three (clear-on-failure, preserve-on-success, forward-id-and-amount at job.rs:262/285/302). The clear-on-failure invariant here is operationally critical because if it regresses `usdc_in_progress` sticks true and freezes all USDC rebalancing in both directions. Add the three mirror tests with a small stub implementing `ResumeAlpacaToBase`. AGENTS.md requires test coverage for all new logic.

src/rebalancing/usdc/manager.rs
```diff
@@ -250,6 +250,255 @@ impl<Chain: Wallet> CrossVenueCashTransfer<Chain> {
     /// On errors, sends appropriate `Fail*` command to transition
     /// aggregate to failed state.
     #[instrument(target = "rebalance", skip(self), fields(%id, %amount), level = tracing::Level::DEBUG)]
+    pub(crate) async fn resume_alpaca_to_base(
```
JuaniRios, line 270: should fix: of the ~12 arms in this function, only `DepositInitiated` is covered (the reverify test at manager.rs:3192). The `Attested` and `Bridged` arms are exactly where the two highest-risk bugs sit, and arm tests mirroring the `resume_base_to_alpaca` suite would likely have caught them. Worth adding coverage for `Bridged` (burn/mint skipped, only deposit runs), `Attested` (mint reached), `ConversionComplete`, the `DepositConfirmed` no-op, and the direction-mismatch guard.

src/rebalancing/usdc/manager.rs
```diff
@@ -250,6 +250,255 @@ impl<Chain: Wallet> CrossVenueCashTransfer<Chain> {
...
+        info!(
+            target: "rebalance",
+            ?state,
+            "Resuming Alpaca->Base transfer from aggregate state",
+        );
+
+        match state {
+            None => self.execute_alpaca_to_base(id, amount).await,
+
+            Some(UsdcRebalance::Converting {
```
JuaniRios: minor: this `Converting` arm sends `FailConversion` and returns `ResumeIndeterminateConversion` because the broker order id is not persisted, so a crash after the fill but before `ConfirmConversion` strands the converted USDC in Alpaca needing manual reconciliation. That is a real residual orphan window, which reads as contradicting the PR description's claim that a mid-transfer crash can no longer orphan in-flight capital. Suggest scoping that guarantee to post-conversion phases in the description, since this mirrors accepted #712 behavior and probably needs no code change.

src/rebalancing/usdc/manager.rs
```diff
@@ -2873,4 +3178,71 @@ mod tests {
...
+        let amount = usdc("100");
+        let deposit_tx =
+            fixed_bytes!("0xdddd000000000000000000000000000000000000000000000000000000000001");
+
+        advance_to_deposit_initiated_alpaca_to_base(&cqrs, &id, amount, deposit_tx).await;
+
+        let result = manager.resume_alpaca_to_base(&id, amount).await;
+
+        assert!(
+            result.is_err(),
```
JuaniRios: minor: `assert!(result.is_err())` is the pattern AGENTS.md forbids for tests. Unwrap and match the exact variant instead, e.g. `let err = result.unwrap_err(); assert!(matches!(err, UsdcTransferError::Vault(_)), "got {err:?}");`, so the test cannot pass on an unrelated setup error and actually pins the `confirm_tx` failure path.

src/rebalancing/trigger/mod.rs
```diff
@@ -2048,6 +2063,83 @@ impl RebalancingService {
...
 
+    /// Sibling of [`Self::enqueue_transfer_usdc_to_hedging`] for the
+    /// Alpaca->Base direction.
+    ///
+    /// Mirrors the same persistent dedupe: an in-memory `usdc_in_progress`
+    /// guard resets on restart, so without querying the apalis Jobs table
+    /// a crash between `queue.push` and the first persisted
+    /// `UsdcRebalance` event would let the next imbalance check enqueue a
+    /// second job for the same rebalance.
+    async fn enqueue_transfer_usdc_to_market_making(&self, amount: Usdc) -> bool {
```
JuaniRios, line 2175: nit: this is a near-verbatim copy of `enqueue_transfer_usdc_to_hedging`, differing only in job type and log strings, kept in lockstep by a comment. A single generic helper parameterized over the job type would remove the drift risk, and the dedupe gap flagged above is exactly the kind of bug that now lives in both copies. Fine to defer to #714 if that is where the cleanup lands.

JuaniRios (changes requested):

src/rebalancing/trigger/mod.rs
0xgleb: Fixed -- the guard queries WHERE job_type IN (?, ?) binding both TransferUsdcToHedging and TransferUsdcToMarketMaking, so an in-flight transfer in either direction blocks the opposite enqueue across restarts

src/rebalancing/trigger/mod.rs
0xgleb: deferred to 714

src/rebalancing/usdc/job.rs
0xgleb: fixed

src/rebalancing/usdc/manager.rs
0xgleb: fixed, now Attested has its own arm routing through continue_alpaca_to_base_from_attested, which reconstructs the mint (adopt an existing one or re-poll the attestation) without re-emitting ReceiveAttestation

src/rebalancing/usdc/manager.rs
0xgleb: Fixed -- the deposit-confirm arm records FailDeposit for operator reconciliation on a confirmed-absent tx instead of retrying forever

src/rebalancing/usdc/manager.rs
0xgleb: scoped the #713 description

src/rebalancing/usdc/manager.rs
0xgleb: fixed

src/rebalancing/usdc/manager.rs
0xgleb: added resume_alpaca_to_base_from_{deposit_initiated_reverifies_deposit_tx, deposit_confirmed_is_noop, attested_does_not_re_emit_receive_attestation, bridged_deposit_submission_failure_stays_bridged} & the ResumeDirectionMismatch guard arms

src/rebalancing/usdc/manager.rs
0xgleb: Fixed -- on an unconfirmed/crashed deposit deposit_to_vault deliberately leaves the aggregate in Bridged (no FailDeposit) so resume re-checks rather than submitting a second deposit; covered by resume_alpaca_to_base_from_bridged_deposit_submission_failure_stays_bridged

JuaniRios (approved):

PR #717, opened 2026-05-29
fix: read-after-write tip check on backfill getLogs

0xgleb:
Closes [RAI-420](https://linear.app/makeitrain/issue/RAI-420/chaos-transient-rpc-failure-injection).

## What

Adds a read-after-write tip check to `enqueue_batch_events` in `src/onchain/backfill.rs`. Every `eth_getLogs` call now follows up with `eth_blockNumber` on the same provider; if the responding node's reported tip is below the requested `to_block`, the call returns the new `OnChainError::NodeLaggingBehindRequest` variant so the existing `backon` retry loop reissues the request.

## Why

This is the implementation half of the failing-test pair stacked beneath it (#715 + #716). #715 fires a `TakeOrderV3` *before* the bot starts and arms the chaos proxy to return empty `eth_getLogs` results for the bot's initial backfill batch. On master the bot trusts the empty response, advances its checkpoint past the take's block, and `OffchainOrderEvent::Filled` never fires -- the test times out. With this PR applied, the bot detects the lagging node, retries, eventually pulls the real logs once the load balancer routes to a caught-up node, and the test passes.

## How

`fetch_logs_with_tip_check` wraps `provider.get_logs` + `provider.get_block_number` into a single fallible call. Both retried `getLogs` closures in `enqueue_batch_events` go through it. The retry loop already lives in `backon`, so the new error variant slots in without changing the retry strategy. The extra `eth_blockNumber` adds one RPC per `getLogs`; cheap.

## Testing

`cargo check --workspace --all-features --all-targets` clean.

`#715`'s `transient_empty_get_logs_during_backfill_does_not_drop_events` is the consumer test. Without this PR it fails at the `OffchainOrderEvent::Filled` poll; with this PR it passes. The existing backfill unit tests continue to use mocked providers that respond with consistent tip + logs, so they keep passing.

## Anything else

Stacked on #716. The new error variant intentionally only surfaces when the responding node lies about its indexing depth -- a healthy node and load balancer will never trigger this path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved blockchain event fetching reliability by detecting when RPC nodes are lagging behind the requested block range and automatically retrying against fresher nodes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ST0x-Technology/st0x.liquidity/pull/717?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #716, opened 2026-05-29
feat: chaos WS proxy implementation

0xgleb:
[RAI-75](https://linear.app/makeitrain/issue/RAI-75/chaos-adversarial-external-behavior-in-e2e-harness) -- harness implementation, fills the `todo!()` left in #715.

## What

Replaces `ChaosProxy::start`'s `todo!()` with a real WebSocket man-in-the-middle:

- Binds a `TcpListener` on `127.0.0.1:0`, exposes the assigned port as `ChaosProxy::endpoint`.
- Per accepted connection: upgrades the inbound socket via `tokio_tungstenite::accept_async`, opens a parallel WS to the upstream Anvil endpoint via `connect_async`, splits both into sink/stream halves, and runs two forwarding tasks under a `tokio::select!`.
- Tracks JSON-RPC `id -> method` mapping on the client->upstream path so the response interceptor knows which method's chaos config to consult.
- On the upstream->client path: parses each text frame as JSON-RPC, looks up the pending method, consults the active `ChaosConfig`, and applies the configured `ChaosBehaviour` to the response body before forwarding. Only `EmptyResult` is wired here; other variants land with the sub-issues that need them.
- `ChaosProxy` keeps the accept loop's `AbortHandle` and aborts it on drop so the proxy dies with the test.

## Why

#715 lands the harness API surface and a failing test that panics in setup because the proxy is `todo!()`. With this PR stacked on top, the proxy actually runs; the chaos test's panic moves from setup into the system-behavior assertion (`OffchainOrderEvent::Filled` poll), which is what the next upstack PR has to address.

## How

- All new code lives in `tests/e2e/chaos.rs`, behind the existing `pub(crate)` surface. No production paths touched.
- Concurrency: per-connection `PendingMethods` is a `tokio::sync::Mutex<HashMap<id, method>>`. The shared `ChaosConfig` is the same `Arc<Mutex<...>>` that #715 introduced and that test code mutates via `ChaosProxy::empty_get_logs`.
- Non-text frames (Ping/Pong/Binary/Close) pass through untouched.
- A response whose corresponding request wasn't seen (id not in the map -- e.g. a subscription notification) passes through untouched.

## Testing

`cargo check --workspace --all-features --all-targets` clean.

[...]

PR #715, opened 2026-05-29
test: chaos harness scaffold + failing RPC empty-getLogs e2e 

0xgleb:
[RAI-75](https://linear.app/makeitrain/issue/RAI-75/chaos-adversarial-external-behavior-in-e2e-harness) (harness foundation) + [RAI-420](https://linear.app/makeitrain/issue/RAI-420/chaos-transient-rpc-failure-injection) (load-balancer-inconsistency scenario).

## Failing test, on purpose

Per [docs/ttdd.md](https://github.com/ST0x-Technology/st0x.liquidity/blob/master/docs/ttdd.md), the test goes in first. This PR alone is **expected to fail CI**: it lands the harness API surface (`ChaosProxy`, `ChaosConfig`, `ChaosBehaviour::EmptyResult`) and the failing-test scaffolding, but `ChaosProxy::start` is `todo!()`. The test panics in setup, not in the system-behavior assertion, so the next upstack PR turns it green by:

1. Implementing the chaos WebSocket proxy.
2. Adding the bot-side recovery logic that the proxy's empty `eth_getLogs` responses currently expose as a missing-events bug.

The stack is: this PR (test + types) -> proxy implementation PR -> recovery-logic PR. Each upstack PR turns red->green for a different reason and the test surface stays the same.

## What

- `tests/e2e/chaos.rs`: `ChaosProxy` + `ChaosConfig` + `ChaosBehaviour` public API. Only the `EmptyResult` behaviour is exposed here -- additional variants (drop, error, stall) land alongside the chaos sub-issues that need them.
- `tests/e2e/chaos_rpc/mod.rs`: `transient_empty_get_logs_during_backfill_does_not_drop_events`. Models RAI-420's load-balancer inconsistency case: spawns the bot through the proxy, fires a SellEquity take, configures the next three `eth_getLogs` responses to come back empty, and asserts the bot still reaches `OffchainOrderEvent::Filled`.
- `tests/e2e/hedging/assertions.rs`: `build_ctx` grows a `ws_rpc_url_override` parameter so chaos tests can route the bot through the proxy without forking the builder.
- `tests/e2e/main.rs`: wires the two new modules.

## Why

RAI-75 calls out that the e2e layer assumes external services always behave correctly. RAI-420 specifically wants validation of load-balancer inconsistency, where one node in the upstream pool is behind on indexing and answers `eth_getLogs` with no events while the chain has actually advanced. The bot's backfill checkpoint must not silently advance past blocks it never inspected.

## How

[...]

PR #718, opened 2026-05-29
feat: Alpaca placement idempotency via client_order_id + chaos test for transient 5xx

0xgleb:
[RAI-75](https://linear.app/makeitrain/issue/RAI-75/chaos-adversarial-external-behavior-in-e2e-harness) (harness extension) + [RAI-421](https://linear.app/makeitrain/issue/RAI-421/chaos-exchange-api-fault-injection-alpaca) (Alpaca double-submit scenario).

## Failing test, on purpose

Per [docs/ttdd.md](https://github.com/ST0x-Technology/st0x.liquidity/blob/master/docs/ttdd.md), the failing test goes in first. This PR alone is **expected to fail CI**: it adds the `AlpacaBrokerMock` knob for transient placement failures and the e2e test that uses it. The assertion `orders.len() == 1` fails on the current bot because `OrderRequest` has no `client_order_id` and the apalis retry path creates a second order on the mock after the 5xx-after-record. The upstack PR adds end-to-end idempotency-key plumbing and turns this green.

## What

- `crates/execution/src/alpaca_broker_api/mock_api.rs`:
  - New `MockState::transient_placement_failures_remaining: usize`.
  - New `AlpacaBrokerMock::set_transient_placement_failures(count)`.
  - `place_order` handler now records the order in state *then* returns 503 while the counter is positive, decrementing per request. Models "broker processed it but the response was lost in flight."
- `tests/e2e/chaos_alpaca/mod.rs`: new `transient_alpaca_5xx_after_record_does_not_double_submit` e2e test. Arms one transient failure, fires a SellEquity take, polls for `OffchainOrderEvent::Filled`, and asserts the broker ended with exactly one order.

## Why

RAI-421 calls out 4xx/5xx fault injection on the Alpaca side and explicitly wants the bot to *not* double-submit and to reconcile pending orders. Today the bot retries failed `place_order` calls without an idempotency key, so any 5xx that occurs after the broker processed the request leaves the broker with two orders for the same intent. The test makes that observable.

## How

The 5xx is returned *after* the order has been inserted into `MockState::orders`, mirroring the real-world "request reached upstream, response was lost." That single difference is what creates the duplicate when apalis retries the `PlaceHedge` job: the second placement creates a second order in the mock because there is no key the broker can use to deduplicate.

## Testing

`cargo check --workspace --all-features --all-targets` clean.

CI's e2e job is expected to fail this test on this branch alone -- the assertion `orders.len() == 1` finds 2. Once the upstack PR plumbs `client_order_id` end-to-end (`OrderRequest`, the bot's hedge job, the mock's dedup table), the same test passes.

## Anything else

Stacked on #717 to share the chaos harness conventions established by #715 / #716. The bot-side idempotency-key fix is the next PR upstack of this branch.

## Release Notes

* **New Features**
  * Added order idempotency support with unique client identifiers to prevent duplicate submissions on retry.
  * Automatic broker-side deduplication of retried orders for more reliable order placement.
  * Graceful recovery from transient broker failures with improved retry handling.

* **Tests**
  * Added chaos testing to validate system fault tolerance under broker failures.

### DIFF
--- a/crates/execution/src/alpaca_broker_api/mock_api.rs
+++ b/crates/execution/src/alpaca_broker_api/mock_api.rs
@@ -102,6 +102,7 @@ struct MockOrder {
     status: OrderStatus,
     poll_count: usize,
     filled_price: Option<Float>,
+    client_order_id: Option<String>,
 }
 
 /// A single calendar entry controlling market open/close times.
@@ -144,6 +145,13 @@ struct MockState {
     wallet_balances: HashMap<String, Float>,
     /// Whitelisted withdrawal addresses.
     whitelisted_addresses: Vec<WhitelistEntry>,
+    /// Number of upcoming `place_order` requests that should be answered
+    /// with a 5xx *after* the order has already been recorded in
+    /// [`MockState::orders`]. Models the adversarial case where the
+    /// broker processes the request but the response is lost in flight:
+    /// any caller-side retry creates a second order even though one
+    /// already exists. Decrements per request until zero.
+    transient_placement_failures_remaining: usize,
 }
 
 /// Status of a whitelisted address.
@@ -305,6 +313,7 @@ impl AlpacaBrokerMock {
             alpaca_deposit_address: String::new(),
             wallet_balances: HashMap::new(),
             whitelisted_addresses: Vec::new(),
+            transient_placement_failures_remaining: 0,
         }));
 
         let server = MockServer::start_async().await;
@@ -336,6 +345,16 @@ impl AlpacaBrokerMock {
     /// Changes the mock mode for subsequent requests.
     pub fn set_mode(&self, mode: MockMode) {
         lock(&self.state).mode = mode;
+    }
+
+    /// Arms the mock to answer the next `count` `place_order` requests
+    /// with a 5xx *after* the order has already been recorded in the
+    /// mock's internal state. Models adversarial broker behaviour where
+    /// the request reaches the upstream and is processed but the
+    /// response is lost in flight; any caller-side retry creates a
+    /// second order with the same intent.
+    pub fn set_transient_placement_failures(&self, count: usize) {
+        lock(&self.state).transient_placement_failures_remaining = count;
     }
 
     /// Sets a per-symbol fill delay: the order stays "new" for
@@ -694,6 +713,7 @@ fn register_endpoints(server: &MockServer, state: &Arc<Mutex<MockState>>) {
     register_latest_trade_endpoint(server, state);
     register_asset_endpoint(server);
     register_order_placement_endpoint(server, state);
+    register_order_by_client_order_id_endpoint(server, state);
     register_order_status_endpoint(server, state);
 }
 
@@ -907,6 +927,7 @@ fn register_order_placement_endpoint(server: &MockServer, state: &Arc<Mutex<Mock
                     );
                 }
             };
+            let client_order_id = body["client_order_id"].as_str().map(str::to_owned);
             let order_id = Uuid::new_v4().to_string();
 
             let mut state = lock(&state);
@@ -919,6 +940,25 @@ fn register_order_placement_endpoint(server: &MockServer, state: &Arc<Mutex<Mock
                 return handle_crypto_order(&mut state, &order_id, &symbol, quantity, side);
             }
 
+            // Real Alpaca rejects a re-used `client_order_id` on an active order
+            // with a 422 ("client_order_id must be unique") rather than deduping
+            // to a 2xx. Mirror that so the executor's recovery path -- look the
+            // order up by client_order_id and adopt it -- is exercised
+            // end-to-end instead of being papered over by a mock-only 200.
+            if let Some(client_id) = client_order_id.as_deref() {
+                let already_recorded = state
+                    .orders
+                    .values()
+                    .any(|order| order.client_order_id.as_deref() == Some(client_id));
+                if already_recorded {
+                    drop(state);
+                    return json_response(
+                        422,
+                        &json!({"message": "client_order_id must be unique"}),
+                    );
+                }
+            }
+
             state.orders.insert(
                 order_id.clone(),
                 MockOrder {
@@ -928,8 +968,19 @@ fn register_order_placement_endpoint(server: &MockServer, state: &Arc<Mutex<Mock
                     status: OrderStatus::New,
                     poll_count: 0,
                     filled_price: None,
+                    client_order_id: client_order_id.clone(),
                 },
             );
+
+            if state.transient_placement_failures_remaining > 0 {
+                state.transient_placement_failures_remaining -= 1;
+                drop(state);
+                return json_response(
+                    503,
+                    &json!({"message": "transient upstream failure (chaos)"}),
+                );
+            }
+
             drop(state);
             json_response(
                 200,
@@ -940,6 +991,7 @@ fn register_order_placement_endpoint(server: &MockServer, state: &Arc<Mutex<Mock
                     "side": side.to_string(),
                     "status": "new",
                     "filled_avg_price": null,
+                    "client_order_id": client_order_id,
                 }),
             )
         });
@@ -1068,6 +1120,7 @@ fn handle_crypto_order(
             status: OrderStatus::Filled,
             poll_count: 0,
             filled_price: Some(fill_price),
+            client_order_id: None,
         },
     );
 
@@ -1083,6 +1136,49 @@ fn handle_crypto_order(
             "created_at": "2025-01-06T12:00:00Z"
         })
     })
+}
+
+fn register_order_by_client_order_id_endpoint(server: &MockServer, state: &Arc<Mutex<MockState>>) {
+    let state = Arc::clone(state);
+    let path = format!("/v1/trading/accounts/{TEST_ACCOUNT_ID}/orders:by_client_order_id");
+
+    server.mock(|when, then| {
+        when.method(GET).path(path.clone());
+        then.respond_with(move |request: &HttpMockRequest| {
+            let Some(client_order_id) = request.uri().query().and_then(|query| {
+                query
+                    .split('&')
+                    .find_map(|pair| pair.strip_prefix("client_order_id=").map(str::to_owned))
+            }) else {
+                return json_response(
+                    400,
+                    &json!({"message": "missing client_order_id query param"}),
+                );
+            };
+
+            let found = lock(&state).orders.iter().find_map(|(id, order)| {
+                (order.client_order_id.as_deref() == Some(client_order_id.as_str())).then(|| {
+                    json!({
+                        "id": id,
+                        "symbol": order.symbol.to_string(),
+                        "qty": format_float_with_fallback(&order.quantity),
+                        "side": order.side.to_string(),
+                        "status": order.status.to_string(),
+                        "filled_avg_price": order
+                            .filled_price
+                            .as_ref()
+                            .map(format_float_with_fallback),
+                        "client_order_id": client_order_id,
+                    })
+                })
+            });
+
+            found.map_or_else(
+                || json_response(404, &json!({"message": "order not found"})),
+                |payload| json_response(200, &payload),
+            )
+        });
+    });
 }
 
 fn register_order_status_endpoint(server: &MockServer, state: &Arc<Mutex<MockState>>) {
@@ -1695,6 +1791,7 @@ mod tests {
             alpaca_deposit_address: format!("{:#x}", Address::ZERO),
             wallet_balances: HashMap::new(),
             whitelisted_addresses: vec![],
+            transient_placement_failures_remaining: 0,
         };
 
         let response = handle_crypto_order(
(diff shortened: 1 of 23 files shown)


0xgleb:
* **#719** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/719?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 4  dependent PRs ([#339](https://github.com/ST0x-Technology/st0x.liquidity/pull/339) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/339?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#721](https://github.com/ST0x-Technology/st0x.liquidity/pull/721) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/721?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#722](https://github.com/ST0x-Technology/st0x.liquidity/pull/722) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/722?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 1 other)
* **#713** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/713?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#718** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/718?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/718?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=718).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor (approved):

crates/execution/src/alpaca_broker_api/mock_api.rs
```diff
@@ -970,6 +990,34 @@ fn register_order_placement_endpoint(server: &MockServer, state: &Arc<Mutex<Mock
...
+                        "id": existing_id,
+                        "symbol": existing_order.symbol.to_string(),
+                        "qty": format_float_with_fallback(&existing_order.quantity),
+                        "side": existing_order.side.to_string(),
+                        "status": "new",
+                        "filled_avg_price": null,
+                        "client_order_id": client_id,
+                    });
+                    drop(state);
+                    return json_response(200, &payload);
```
JuaniRios: critical: this dedup branch returns 200 with the existing order, and the comment says it matches Alpaca's native dedupe behaviour, but real Alpaca rejects a duplicate client_order_id on an active order with a 422 ("client_order_id must be unique"), not a 200. The executor has no 422-recovery and never calls GET .../orders:by_client_order_id, so in production the retry gets a 422, fails again, and the bot never reconciles the order Alpaca actually accepted, leaving the position silently un-hedged. We should make the mock return 422 on a duplicate and add a by-client-order-id lookup that reconciles the accepted order. The learn page documenting this currently 502s, but its cached text says "a duplicate client_order_id was used for another active order ... 'client_order_id must be unique'" (https://alpaca.markets/learn/how-to-fix-common-trading-api-errors-at-alpaca); lookup endpoint: https://docs.alpaca.markets/reference/getorderbyclientorderid . Worth a quick paper-API confirmation before acting.

src/position.rs
```diff
@@ -165,8 +185,15 @@ impl EventSourced for Position {
...
-            OffChainOrderFailed { failed_at, .. } => Ok(Some(Self {
+            OffChainOrderFailed {
+                offchain_order_id,
+                failed_at,
+                ..
+            } => Ok(Some(Self {
                 pending_offchain_order_id: None,
+                // Stash the failed OID so the next placement attempt can
+                // reuse it as `client_order_id` and let the broker dedupe.
+                last_failed_offchain_order_id: Some(*offchain_order_id),
```
JuaniRios: should fix: this overwrites the anchor unconditionally on every failure. If attempt A is recorded at the broker then its retry (using key A) also loses its response, pending is now B, so this stashes B and the next attempt places with key B while the broker only holds A, which double-submits. Preserve the first anchor instead, e.g. entity.last_failed_offchain_order_id.or(Some(*offchain_order_id)). It is still cleared on fill.

src/trading/offchain/hedge.rs
```diff
@@ -146,6 +156,15 @@ impl Job<HedgeCtx> for PlaceHedge {
             Err(error) => return Err(error.into()),
         }
 
+        // Prefer the position's last-failed-attempt id so a retry across
+        // a fresh `PlaceHedge` enqueue (after a transient 5xx) shares the
+        // same broker-side `client_order_id` as the original attempt.
+        // Fall back to the current aggregate id for the first attempt.
+        let client_order_id_source = self
```
JuaniRios: should fix: the key is pinned to the original failed attempt, but shares and direction are recomputed from current net on each scan, so a retry after intervening onchain takes carries the old key with a larger or different intent. If the broker dedupes on that key it returns the original smaller order and the extra exposure never gets hedged. The key should identify the exact (symbol, direction, shares) intent, or the anchor should be cleared when the recomputed intent differs.

src/position_check.rs
```diff
@@ -149,6 +149,7 @@ where
             executor: ready.executor,
             threshold: self.ctx.execution_threshold,
             offchain_order_id: OffchainOrderId::new(),
+            last_failed_offchain_order_id: ready.last_failed_offchain_order_id,
```
JuaniRios: should fix: the anchor is serialized into PlaceHedge at enqueue time. If several hedge jobs are queued before the first runs, the first failure updates the position to anchor A but the already-queued stale job still carries None and places with a fresh key, double-submitting. Better to derive client_order_id from the live Position aggregate after PlaceOffChainOrder succeeds rather than capturing it at enqueue.

crates/execution/src/order/mod.rs
```diff
@@ -12,6 +13,67 @@ pub mod status;
...
+    ///
+    /// # Errors
+    /// Returns an error if the input is empty.
+    pub fn new(value: impl Into<String>) -> Result<Self, ClientOrderIdError> {
+        let raw: String = value.into();
+        if raw.is_empty() {
+            return Err(ClientOrderIdError::Empty);
+        }
+        let truncated = if raw.len() > 128 {
+            raw[..128].to_owned()
```
JuaniRios: should fix: raw[..128] slices by byte, so a non-ASCII input whose 128th byte lands mid-char panics, and silent truncation can also collapse two distinct keys into the same broker key, which is precision-capping on a dedupe identifier. Returning a typed TooLong error instead of truncating removes both. Also note the limit: this bot uses the Broker API, where client_order_id is documented as <= 48 chars, not the 128 in the comment (https://github.com/alpacahq/alpaca-docs/blob/master/content/api-references/broker-api/trading/orders.md).

src/position.rs
```diff
@@ -145,6 +161,9 @@ impl EventSourced for Position {
             } => Ok(Some(Self {
                 net: (entity.net + shares_filled.inner())?,
                 pending_offchain_order_id: None,
+                // Successful fill: drop the failed-attempt anchor so the
```
JuaniRios, line 164: should fix: the stash-on-failure and clear-on-fill transitions, plus the PlaceHedge routing, are the core of this change but have no unit coverage. offchain_failed_clears_pending only checks pending is None. Worth adding an assert that last_failed == Some(oid) after failure and a replay test asserting it is None after a fill.

tests/e2e/chaos_alpaca/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+        "OffchainOrderEvent::Filled",
+        1,
+        Duration::from_secs(120),
+    )
+    .await;
+
+    let orders = infra.broker_service.orders();
+    let order_count = orders.len();
+    assert_eq!(
+        order_count, 1,
```
JuaniRios, line 108: minor: this asserts the outcome but not that the recovery path ran. If a fill landed before the rescan re-enqueued, it could pass without touching the idempotency code. Consider also asserting the Failed event and that a second PlaceHedge was enqueued.

src/cli/trading.rs
```diff
@@ -789,16 +795,23 @@ mod tests {
...
                 .path("/v1/trading/accounts/904837e3-3b76-47ec-b432-046db621571b/orders")
-                .json_body(json!({
-                    "symbol": symbol,
-                    "qty": quantity,
-                    "side": side,
-                    "type": "market",
-                    "time_in_force": "day",
-                    "extended_hours": false
-                }));
+                .json_body_includes(
```
JuaniRios, line 828: minor: switching from exact json_body to json_body_includes drops any assertion that client_order_id is present in the request at all, and the chaos tests use the broker mock rather than this httpmock so they do not cover it. Worth asserting the body contains the client_order_id key.

crates/execution/src/alpaca_broker_api/order.rs
```diff
@@ -120,6 +120,7 @@ pub(super) struct OrderRequest {
     pub order_type: &'static str,
     pub time_in_force: &'static str,
     pub extended_hours: bool,
+    pub client_order_id: String,
```
JuaniRios: nit: OrderRequest is a private wire struct and ClientOrderId is serde(transparent), so this could be ClientOrderId directly and drop the .as_str().to_owned() at the call site.

tests/e2e/chaos_alpaca/mod.rs
```diff
@@ -0,0 +1,85 @@
+//! Chaos tests for Alpaca broker fault tolerance.
+//!
+//! Drives the bot through scenarios where the Alpaca broker mock returns
```
JuaniRios, line 3: nit: the docstring frames this as apalis retrying the job, but the actual recovery is: the 503 is handled, the job returns Ok, the position stashes the anchor, and the CheckPositions scan enqueues a fresh PlaceHedge. Worth correcting so the mechanism is clear.

src/onchain/accumulator.rs
```diff
@@ -15,6 +15,12 @@ pub(crate) struct ExecutionCtx {
     pub(crate) direction: Direction,
     pub(crate) shares: Positive<FractionalShares>,
     pub(crate) executor: SupportedExecutor,
+    /// If the previous placement attempt for this position failed, the
+    /// position aggregate stashes its `OffchainOrderId` here so we can
+    /// reuse it as the broker-side `client_order_id` and let the broker
+    /// dedupe a duplicate submission whose first response was lost in
+    /// flight. `None` means this is a fresh attempt with no prior anchor.
+    pub(crate) last_failed_offchain_order_id: Option<crate::offchain::order::OffchainOrderId>,
```
JuaniRios: nit: inline full path here, imported everywhere else in this change. Add OffchainOrderId to the use crate::... group per the import convention.

JuaniRios (changes requested):

crates/execution/src/alpaca_broker_api/mock_api.rs
0xgleb: the mock now returns 422 "client_order_id must be unique" on a duplicate and exposes the by_client_order_id lookup; the executor classifies that 422 via is_duplicate_client_order_id and reconciles the accepted order with get_order_by_client_order_id (order.rs:351-372), covered by place_market_order_reconciles_duplicate_client_order_id

crates/execution/src/order/mod.rs
0xgleb: fixed. ClientOrderId is now a UUID-based typed enum (Automated/Cli) with an InvalidUuid error; no byte-slicing or truncation, and the rendered id is well under the Broker API's 48-char limit

tests/e2e/chaos_alpaca/mod.rs
0xgleb: the test now asserts the OffchainOrderEvent::Failed event and a second OffchainOrder aggregate (the scan re-enqueue), not just the outcome

tests/e2e/chaos_alpaca/mod.rs
0xgleb: the docstring now describes the real mechanism (503 handled -> job returns Ok -> position stashes the anchor -> CheckPositions re-enqueues a fresh PlaceHedge -> broker 422 reconciled via lookup)

src/trading/offchain/hedge.rs
0xgleb: The key is derived from the live Position anchor after PlaceOffChainOrder claims it (hedge.rs:151-167); a larger recomputed intent reconciles the original order, fills it (clearing the anchor), and the next CheckPositions scan hedges the remaining delta with a fresh key -- so the extra exposure isn't dropped

src/cli/trading.rs
0xgleb: added a regex-based check

src/position_check.rs
0xgleb: the client_order_id is no longer captured at enqueue; PlaceHedge derives it from the live Position aggregate after PlaceOffChainOrder claims the order

src/position.rs
0xgleb: fixed

src/position.rs
0xgleb: offchain_failed_clears_pending now also asserts last_failed == Some(oid), and the new offchain_failed_preserves_first_failed_anchor / offchain_filled_clears_failed_anchor cover the chain-preserve and clear-on-fill invariants

JuaniRios (approved):

PR #719, opened 2026-05-29
refactor: drop dead USDC routing from Rebalancer + TriggeredOperation

0xgleb:
## What

Closes [RAI-859](https://linear.app/makeitrain/issue/RAI-859).

Removes the dormant code left behind by #713: `Rebalancer` no longer holds or routes USDC transfers, `TriggeredOperation::Usdc*` variants are gone, `CrossVenueTransfer` impls on `CrossVenueCashTransfer` are dropped, the CLI uses `execute_alpaca_to_base`/`execute_base_to_alpaca` directly, and `MockUsdcRebalance` (only used to mock the now-deleted impls) is deleted.

## Why

After #713, the Rebalancer's USDC fields, the two `TriggeredOperation` USDC variants, and the `CrossVenueTransfer<...> for CrossVenueCashTransfer` impls were never reached at runtime -- USDC dispatch had already moved to apalis. Keeping the dead code around forces every reviewer to spend cycles confirming "yes, this is dead". This PR is the deletion pass.

## How

- Drop `usdc_to_mm`, `usdc_to_hedging`, `usdc_in_progress` fields + the corresponding constructor params + match arms + `clear_usdc_in_progress` from `Rebalancer`.
- Drop the `TriggeredOperation::UsdcAlpacaToBase` and `UsdcBaseToAlpaca` variants; introduce `UsdcRebalanceOperation` inside the USDC trigger module and change `check_imbalance_and_build_operation` to return it.
- Drop the two `CrossVenueTransfer` impls on `CrossVenueCashTransfer`. CLI `transfer-usdc` now calls `execute_alpaca_to_base` / `execute_base_to_alpaca` directly with a fresh `UsdcRebalanceId`.
- Delete `MockUsdcRebalance` and the `mock_*` tests in `manager.rs` (they tested only the deleted trait impls). Other tests that previously passed the mock to `Rebalancer::new` drop those args.

## Testing

- `cargo nextest run --lib`: green.
- `cargo clippy --workspace --all-targets --all-features`: green.
- No behavior change at runtime -- all production dispatch already went through apalis after #713.

## Summary by CodeRabbit

* **Refactor**
  * Rebalanced handling split: the background rebalancer now processes equity-only operations; USDC transfers are enqueued and executed via dedicated transfer jobs.
  * Improved in-flight detection for USDC transfers to avoid premature timeout/cleanup across both directions.
* **Tests**
  * Test suites updated to reflect the new USDC job model and simplified equity-only rebalancer; USDC test mocks removed.

PR #721, opened 2026-06-01
feat: base unwrapped equity recovery mechanism

0xgleb:
## What

[RAI-88](https://linear.app/makeitrain/issue/RAI-88/recover-from-detected-unwrapped-equity-tokens-on-base)

Adds end-to-end automatic recovery for unwrapped equity tokens (tSTOCK) found on the bot wallet on Base. Mirrors `WrappedEquityRecovery` but extends the orphan path to four on-chain steps (wrap + deposit) because the tokens haven't been wrapped yet, and uses `InFlightEquityLocation::BaseWalletUnwrapped` for detection.

- `SPEC.md` -- new "Unwrapped Equity Recovery" subsection under Equity Rebalancing covering the three dispatch paths (active mint / active redemption / orphan), the four-step orphan command sequence, idempotency model, conflict handling, and the aggregate's state diagram.
- `src/unwrapped_equity_recovery/aggregate.rs` -- event-sourced state machine. Eight commands (`Detect`, `DispatchToMint`, `DispatchToRedemption`, `SubmitOrphanWrap`, `ConfirmOrphanWrap`, `SubmitOrphanDeposit`, `ConfirmOrphanDeposit`, `FailRecovery`); dispatch-success states (`DispatchedToMint`, `DispatchedToRedemption`, `OrphanDeposited`) are terminal. `Services = UnwrappedEquityRecoveryServices { raindex, wrapper, transfer, wallet }`: each dispatch command's handler calls the relevant service inside the handler and emits the success event iff the side effect actually completed. Service failures are recorded as `RecoveryFailed` events.
- `src/unwrapped_equity_recovery/job.rs` -- pure orchestration. `perform` claims the per-symbol `equity_in_progress` guard, reads inventory, validates the active aggregate's quantity matches the snapshot, picks a `DispatchDecision`, and sends the command sequence. The orphan path sends four commands in order with terminal-state short-circuits between each so a crash between any two steps resumes at the right place.
- `src/conductor.rs`, `src/conductor/{builder,job}.rs`, `src/rebalancing/trigger/mod.rs`, `src/lib.rs` -- registers the apalis worker (gated on rebalancing being enabled; `debug!` log when disabled), threads the queue + ctx through the builder, and wires the `BaseWalletUnwrappedEquity` inventory event to push `UnwrappedEquityRecoveryJob` per symbol with a positive balance (with the same per-symbol-existing-row dedupe as the wrapped path).
- `tests/e2e/rebalancing/mod.rs` -- new `unwrapped_equity_in_bot_wallet_recovers_into_raindex` e2e that seeds the wallet with tSTOCK via vault deployment, runs the bot with recovery enabled, asserts both unwrapped and wrapped balances reach zero, the configured Raindex vault is topped up by at least the recovered amount, and the recovery aggregate's event stream records the full orphan sequence `Detected → OrphanWrapSubmitted → OrphanWrapped → OrphanDepositSubmitted → OrphanDeposited`.

## Why

A tSTOCK balance on the bot wallet means equity was minted or unwrapped but never made it to its market-making venue -- the result of a mint stalled after `TokensReceived` but before wrapping, a redemption stalled after `TokensUnwrapped` but before sending to the broker, or tokens arriving at the wallet without any in-flight transfer (external deposit, prior-process crash, or a false-positive `TransactionDropped` from a laggy RPC backend that left the bot believing a wrap attempt failed). Without automatic recovery these balances sit indefinitely, idle from the system's perspective and invisible to the rebalancer's normal triggers.

This PR mirrors the wrapped recovery design so detection and orchestration follow a single pattern: the inventory poller emits `BaseWalletUnwrappedEquity`, the trigger reactor pushes a job per symbol, the job picks a dispatch path from `InventoryView`, and the aggregate's command handlers run the actual on-chain work inside their `Services`. The four-step orphan path persists each transaction independently so the bot can resume from any mid-sequence crash without double-wrapping.

## How

- The aggregate's orphan path is four commands (`SubmitOrphanWrap → ConfirmOrphanWrap → SubmitOrphanDeposit → ConfirmOrphanDeposit`) rather than the wrapped path's two. `OrphanWrapped` carries the actual wrapped amount returned by `confirm_wrap`, which `SubmitOrphanDeposit` then uses for the Raindex deposit -- the underlying `shares` from `Detected` aren't reused because wrapping at the configured ratio may produce a different output amount.
- `UnwrappedEquityRecoveryServices` adds a `wallet: Address` field over the wrapped recovery's `Services` because `wrapper.submit_wrap` needs an explicit receiver. The wallet is the bot's Base owner address, the same address that funds the subsequent deposit.
- Job-level idempotency: the apalis payload carries the `UnwrappedEquityRecoveryId` so apalis retries re-target the same aggregate. Between each orphan command the job reloads the aggregate and short-circuits if it's already terminal, so a retry that lands after `OrphanDeposited` is recorded doesn't emit further commands.
- Trigger dispatch reuses the `equity_in_progress` per-symbol guard the wrapped path uses, and the per-symbol `wrapped_equity_recovery` config flag also gates the unwrapped path (the same intent -- "auto-recover misplaced equity" -- covers both).
- The recovery store is built alongside the wrapped recovery store in `conductor.rs` after `recovery_transfer` exists. Both ctxs flow through the same builder plumbing.

## Testing

- `cargo nextest run -p st0x-hedge --all-features unwrapped_equity_recovery` -- 5 aggregate state-machine tests pass (detect-initializes, detect-on-live-rejected, confirm-wrap-requires-submit, submit-deposit-requires-confirm, terminal-state-rejects).
- New e2e `unwrapped_equity_in_bot_wallet_recovers_into_raindex` exercises the full orphan path end-to-end against the live test infra.
- `cargo check --workspace --all-targets --all-features` -- clean.
- `cargo clippy --workspace --all-targets --all-features` -- clean.

## Anything else

The `wallet` field on `UnwrappedEquityRecoveryServices` is the only structural difference from `WrappedEquityRecoveryServices`. If wrapped recovery ever needs to wrap (which it doesn't today -- its tokens are already wrapped), that field becomes shared and the two services can collapse.


0xgleb:
* **#721** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/721?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/721?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#719** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/719?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3 other dependent PRs ([#339](https://github.com/ST0x-Technology/st0x.liquidity/pull/339) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/339?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#722](https://github.com/ST0x-Technology/st0x.liquidity/pull/722) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/722?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#740](https://github.com/ST0x-Technology/st0x.liquidity/pull/740) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/740?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#713** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/713?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#718** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/718?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=721).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/unwrapped_equity_recovery/job.rs
```diff
@@ -0,0 +1,692 @@
...
+        let symbol = self.symbol.clone();
+
+        let Some(guard) = claim_guard(&ctx.equity_in_progress, &symbol) else {
+            debug!(
+                target: "rebalance",
+                %symbol,
+                recovery_id = %self.recovery_id,
+                "Unwrapped equity recovery: equity_in_progress held; returning retryable error",
+            );
+            return Err(UnwrappedEquityRecoveryJobError::EquityInProgressGuardHeld { symbol });
```
JuaniRios: critical: returning a retryable error on guard contention can halt the whole bot, not just this worker. A normal rebalance claims equity_in_progress and defuses the guard so it stays held for the mint's entire lifecycle, and the poller reads the raw on-chain balanceOf, so during a mint's TokensReceived to wrapped window this job gets enqueued, fails to claim the guard, exhausts its 3 retries in ~7s, and the terminal failure fires the shared failure_notify that stops the entire apalis monitor (hedging, order fill, rebalancing, everything). The wrapped recovery path returns Ok(()) on this exact race precisely to avoid it. Suggest matching it (return Ok and rely on the next poll), or rescheduling with push_with_delay so contention never becomes a terminal failure.

src/rebalancing/trigger/mod.rs
```diff
@@ -1307,14 +1316,62 @@ impl RebalancingService {
...
+                    if !self.is_wrapped_equity_recovery_enabled(symbol) {
+                        continue;
+                    }
+                    let mut queue = self.unwrapped_equity_recovery_queue.clone();
+                    let recovery_id = UnwrappedEquityRecoveryId(uuid::Uuid::new_v4());
+                    // The apalis Jobs table's `UNIQUE INDEX(job_type,
+                    // idempotency_key)` makes this insert atomic: two
+                    // concurrent enqueues for the same symbol can't both
+                    // succeed, so we don't need a separate COUNT pre-check.
+                    let idempotency_key = format!("UnwrappedEquityRecovery:{symbol}");
```
JuaniRios: should fix: this dedup doesn't behave the way the comment claims. apalis sink.sql uses ON CONFLICT(job_type, idempotency_key) DO NOTHING, so a duplicate insert returns Ok with 0 rows affected and never raises a unique-violation error, which makes QueuePushError::Duplicate and the debug skip-log below dead code. Separately, the unique index spans rows of all statuses and ack only marks finished rows Done (actual deletion is the hourly cleanup), so once a recovery completes its Done row keeps this key occupied and new tSTOCK for the same symbol can't be enqueued for up to an hour, with no log explaining why. The wrapped path's status-scoped COUNT pre-check avoids both, so consider reusing it, or add a partial unique index on non-terminal statuses and detect the no-op via rows_affected.

src/unwrapped_equity_recovery/job.rs
```diff
@@ -0,0 +1,692 @@
...
+        ?snapshot.dispatch,
+        "Unwrapped equity recovery: dispatched detection",
+    );
+
+    dispatch_from_detected(ctx, recovery_id, symbol, snapshot.dispatch).await
+}
+
+/// Resume from `Detected`: a previous attempt persisted `Detect` but crashed
+/// before any dispatch command landed. Re-read inventory and dispatch.
+async fn resume_from_detected(
```
JuaniRios, line 270: should fix: resume_from_detected skips validate_active_aggregate_quantity, but start_from_detect runs it before dispatching. A crash that persisted Detect and then retries enters here and dispatches without the stale-aggregate check that the fresh path enforces. Suggest calling validate_active_aggregate_quantity here too, or folding it into dispatch_from_detected so every dispatch (fresh or resumed) goes through it.

SPEC.md
```diff
@@ -2136,6 +2136,164 @@ returning and the `OrphanDepositSubmitted` event persisting will still re-submit
...
+- Terminal states: `DispatchedToMint`, `DispatchedToRedemption`,
+  `OrphanDeposited`, `Failed`.
+
+###### Job-level Idempotency
+
+Same model as `WrappedEquityRecoveryJob`: the apalis payload carries the
+`UnwrappedEquityRecoveryId` so retries re-target the same aggregate and pick up
+from the next non-terminal command. A crash between `wrap()` returning and
+`OrphanWrapSubmitted` persisting still re-submits on retry; the
+`client_order_id`-style idempotency the wrapper layer enforces (or a nonce-based
```
JuaniRios, line 2293: should fix: this describes an idempotency mechanism that doesn't exist. Wrapper::submit_wrap takes (wrapped_token, underlying_amount, wallet) with no client_order_id or nonce, and the handler emits OrphanWrapSubmitted only after submit_wrap returns, so a crash in that window re-submits on retry with no replay protection. The practical result is the aggregate can land in Failed for a wrap that actually succeeded (the wtSTOCK then gets mopped up by the wrapped path). Please correct the SPEC to describe the real behavior, and consider an on-chain balance check before re-submitting so a succeeded-but-unpersisted wrap resumes into OrphanWrapped. Minor: wrapping is on-chain, not a broker call.

src/unwrapped_equity_recovery/job.rs
```diff
@@ -0,0 +1,692 @@
...
+                    },
+                );
+            }
+            Ok(())
+        }
+        DispatchDecision::Orphan | DispatchDecision::Conflict { .. } => Ok(()),
+    }
+}
+
+fn decide_dispatch(view: &InventoryView, symbol: &Symbol) -> DispatchDecision {
```
JuaniRios, line 541: minor: decide_dispatch keys only on the presence of active_mint/active_redemption, not their lifecycle stage. But note perform claims the guard before reaching here, and active_mint/active_redemption are set in inventory exactly while the guard is held, so a successful claim implies no active transfer and this should almost always return Orphan. Can you confirm the ActiveMint/ActiveRedemption/Conflict branches and validate_active_aggregate_quantity are actually reachable outside a narrow guard/inventory desync window? Same structure as the wrapped path so not a blocker, just want to make sure it's intended.

src/unwrapped_equity_recovery/job.rs
```diff
@@ -0,0 +1,692 @@
...
+            }
+            Some(UnwrappedEquityRecovery::OrphanWrapped { .. }) => {
+                resume_from_submit_deposit(ctx, &self.recovery_id).await
+            }
+            Some(UnwrappedEquityRecovery::OrphanDepositSubmitted { .. }) => {
+                resume_from_confirm_deposit(ctx, &self.recovery_id).await
+            }
+            // Terminal states were filtered above; the aggregate's evolve
+            // function never produces other non-terminal variants.
+            Some(_) => Ok(()),
```
JuaniRios, line 210: minor: this Some(_) => Ok(()) leans on the is_terminal() filter above, but a future non-terminal state variant would silently fall here and never be driven forward. Suggest enumerating the terminal variants explicitly so adding a state fails the compile instead of becoming a silent no-op.

src/unwrapped_equity_recovery/aggregate.rs
```diff
@@ -0,0 +1,940 @@
...
+        }
+    };
+
+    match services
+        .raindex
+        .submit_deposit(
+            wrapped_token,
+            vault_id,
+            wrapped_amount,
+            TOKENIZED_EQUITY_DECIMALS,
```
JuaniRios, line 738: minor: this deposits the confirm_wrap output scaled by TOKENIZED_EQUITY_DECIMALS (18), but that amount is the wrapped ERC-4626 share token, whose decimals aren't guaranteed to be 18. It's correct as long as the wrapped token is 18 decimals (same assumption as the wrapped path), but since it's a money path it's worth resolving the wrapped token's actual decimals or asserting the invariant.

SPEC.md
```diff
@@ -2136,6 +2136,164 @@ returning and the `OrphanDepositSubmitted` event persisting will still re-submit
...
+without writing an aggregate. The `equity_in_progress` guard prevents
+overlapping runs against the same symbol.
+
+##### UnwrappedEquityRecovery Aggregate
+
+The audit trail AND the actor for recovery side effects, keyed by
+`UnwrappedEquityRecoveryId(Uuid)`. Each detection-and-dispatch attempt is its
+own aggregate instance.
+
+`Services = UnwrappedEquityRecoveryServices { raindex: Arc<dyn Raindex>,
```
JuaniRios, line 2224: nit: the actual struct adds a wallet: Address field (the wrap receiver and deposit source), so this isn't quite the same shape as WrappedEquityRecoveryServices. Worth listing it here since the SPEC is the source of truth.

JuaniRios (changes requested):

0xgleb:
Superseded by the 3-PR split of this work (dead-code constraints made a finer split infeasible, so aggregate+job+wiring ship as one impl PR):

- #727 — spec: base unwrapped equity recovery mechanism (SPEC.md)
- #728 — test: e2e for base unwrapped equity recovery (failing-test-first / TTDD)
- #729 — feat: unwrapped equity recovery aggregate, job, and conductor wiring

All review feedback on this PR has been addressed: the SPEC corrections (wrap-idempotency description + services `wallet` field) land in #727; the code fixes land in #729 — the guard-contention bot-halt fix, the status-scoped enqueue dedup (removing the dead idempotency-key path), the resume-from-Detected active-aggregate validation, the explicit terminal-variant enumeration, and the documented wtSTOCK 18-decimal invariant. Two 'consider…' suggestions (on-chain balance check before re-wrap, runtime decimals validation) are tracked as follow-ups.

PR #723, opened 2026-06-02
test: finality-gate withdrawal scan against RPC node lag

0xgleb:
## What

Closes [RAI-823](https://linear.app/makeitrain/issue/RAI-823/crash-safe-burnwithdrawal-resume-can-re-issue-an-irreversible-action). Hardens the crash-safe burn/withdrawal resume scan against false-empty `eth_getLogs` responses from lagging load-balanced RPC nodes (e.g. dRPC).

## Why

A load-balanced RPC node may return an empty `eth_getLogs` result not because the burn is absent, but because that particular node is behind the chain head. Previously, a single empty scan was enough to conclude no burn existed, which could cause the resume path to re-issue a burn — burning USDC twice with at most one mint.

## How

- `find_recent_burn` now retries up to `SCAN_ATTEMPTS` (5) times with a `SCAN_RETRY_BACKOFF` (150ms) between attempts, so different load-balanced nodes are likely to answer each request.
- An empty scan is only treated as a true absence when the node's reported head is at least `SCAN_FINALITY_MARGIN` (2) blocks past `from_block` **and** all retry attempts agree the burn is absent.
- If the node appears to be lagging (head not yet past `from_block`) after all attempts, the function returns a new retryable `CctpError::ScanInconclusive { from_block }` error instead of `Ok(None)`, ensuring the caller never re-burns based on a stale empty result.
- The burn match criteria are tightened from `(depositor, amount)` to `(depositor, amount, destinationDomain, mintRecipient)`, so an adopted burn is provably this transfer's and not a same-amount burn to a different destination.
- The same `ScanInconclusive` guard is applied to `find_recent_withdrawal` in the Raindex service.
- The `find_recent_burn` public API now requires `recipient` to be passed explicitly, and the `BaseToEthereum` resume path passes `market_maker_wallet` as the recipient.

## Testing

Two new unit tests cover the key safety properties of `find_recent_withdrawal`:

- `find_recent_withdrawal_is_inconclusive_when_node_lags`: mocks every scan attempt returning an empty log set with a head below `from_block`, and asserts that `ScanInconclusive` is returned rather than `None`.
- `find_recent_withdrawal_is_none_when_caught_up_and_absent`: mocks every scan attempt returning an empty log set with a head well past `from_block`, and asserts that `Ok(None)` is correctly returned, allowing a safe re-issue.

## Anything else

[...]

PR #724, opened 2026-06-02
fix: keep usdc guard latched on timeout while transfer job in flight

0xgleb:
## What

Fixes two crash-recovery bugs in the USDC rebalancing pipeline and adds a safety guard in the timeout sweeper to prevent double-spend during job resumption.

## Why

Three distinct failure modes were identified:

1. **`resume_base_to_alpaca` from `Attested` re-emitted `ReceiveAttestation`**, which the aggregate rejects as an invalid command from that state, causing the resume to fail instead of proceeding to the mint.

2. **`resume_base_to_alpaca` from `Bridged` could issue a fund-moving Alpaca deposit call**, even though the CCTP mint already credits USDC directly to the Alpaca-monitored address. Re-issuing a deposit would be a double-spend.

3. **The timeout sweeper could clear the `usdc_in_progress` guard while a `TransferUsdcToHedging` apalis job was still in flight.** A crash mid-withdraw/burn leaves the aggregate at a `*Submitting` state with an irreversible on-chain effect possibly already submitted. Clearing the guard would allow a second transfer to touch the same vault/wallet concurrently.

## How

- **`Attested` resume path**: Routes `Attested` through `continue_from_attested` (poll → mint) rather than re-emitting `ReceiveAttestation`, so the resume correctly proceeds to the mint step.

- **`Bridged` resume path**: `continue_from_bridged` now only re-records intent via `InitiateDeposit` (an idempotent aggregate event) and polls Alpaca read-only via `poll_deposit_by_tx_hash`. No fund-moving Alpaca API call is issued. A doc comment explains the crash-safety invariant and flags the assumption that must hold for this to remain safe.

- **Timeout sweeper guard**: Before cleaning up a timed-out transfer, `check_and_trigger_usdc` now calls `transfer_usdc_to_hedging_in_flight()`, which queries the apalis `Jobs` table for any non-terminal (`Pending`/`Queued`/`Running`) `TransferUsdcToHedging` row. If one exists, cleanup is skipped entirely — the guard, tracking entry, and timed-out map are all left intact so the resuming job can drive the aggregate to terminal. On DB query failure the method defaults to `true` (safe/latched).

## Testing

- **`find_recent_burn_returns_tx_of_real_burn`** and **`find_recent_burn_returns_none_for_wrong_amount_or_recipient`**: Verify that CCTP burn scanning returns the correct transaction for a matching amount+recipient and `None` for mismatches.

[...]

PR #725, opened 2026-06-03
feat: crash-safe resume for Base->Alpaca USDC transfers (Converting/Attested via order lookup)

0xgleb:
## What

Closes [RAI-822](https://linear.app/makeitrain/issue/RAI-822/base-alpaca-usdc-transfer-cannot-crash-resume-from-converting-or). Adds crash-safe resume for the later phases of a Base->Alpaca USDC transfer — resuming from `Converting` (the Alpaca USDC->USD conversion) and from `Attested` (post-burn, pre-mint) — and keeps the `usdc_in_progress` guard latched while a `TransferUsdcToHedging` job is still in flight.

## Why

A crash mid-conversion or mid-attestation previously could not resume deterministically:

- Resuming from `Converting` recorded `ConversionFailed` unconditionally, even if the Alpaca order actually filled — losing the conversion result.
- The inflight-timeout sweep could clear `usdc_in_progress` while the transfer job was still retrying, letting a fresh transfer re-arm on top of a partially-completed one.
- Resuming from `Attested` re-emitted `ReceiveAttestation`, which the aggregate rejects from `Attested`, fail-stopping a transfer whose USDC is already burned.

## How

- **Converting resume**: look up the Alpaca conversion order by its client order id (`get_crypto_order_by_client_order_id` / `find_conversion_order`), classify the broker status (`CryptoOrderOutcome::classify`), and deterministically pick `ConfirmConversion` (filled), `FailConversion` (terminally failed / never reached Alpaca), or retry (still settling).
- **Attested resume**: routes through `continue_from_attested` (re-poll Circle, then mint) without re-emitting `ReceiveAttestation`.
- **Guard latch**: `usdc_in_progress` stays latched while a non-terminal `TransferUsdcToHedging` row exists (`transfer_usdc_to_hedging_in_flight`), so the timeout sweep can't clear it and let automation re-arm a duplicate transfer.
- Adds negative-case coverage for `find_recent_burn` adoption matching, which already keys on `(depositor, amount, destinationDomain, mintRecipient)` — including a burn from a different depositor that must not be adopted. No predicate change.

## Testing

- Resume-arm tests: `resume_base_to_alpaca_from_attested_does_not_re_emit_receive_attestation`; `resume_base_to_alpaca_from_converting_{confirms_from_filled_order, still_settling_retries, fails_on_terminally_failed_order, fails_when_order_never_reached_alpaca}`; `resume_base_to_alpaca_from_bridged_makes_no_fund_moving_deposit_call`.
- `timeout_does_not_clear_guard_while_transfer_job_in_flight`.
- `find_recent_burn` / `find_recent_withdrawal` scan tests (real-burn match, wrong-amount/wrong-recipient/different-depositor rejection, vault/token mismatch).
- `CryptoOrderOutcome::classify` maps every broker status to its outcome.

## Anything else

Stacked on #723 (finality-gate burn/withdrawal scan). Resolves several #712 review threads immediately upstack of #712: Converting resume, Attested re-emit, Bridged no-fund-move test, and the guard-latch-on-timeout.

## Summary by CodeRabbit

* **New Features**
  * Improved crash-safe resumption for USDC conversions by tracking and resuming previously-placed conversion orders (including per-conversion correlation IDs) and polling until terminal outcomes.
* **Bug Fixes**
  * Prevent premature cleanup of in-flight USDC transfers when an associated transfer job is still running; defer cleanup and log a warning.
* **Tests**
  * Expanded tests: burn scanning, withdrawal lookups, conversion resumption, and rebalancing/in-flight transfer behaviors.
* **Documentation**
  * Clarified resume/continuation invariants and error messaging.

### DIFF
--- a/crates/bridge/src/cctp/mod.rs
+++ b/crates/bridge/src/cctp/mod.rs
@@ -1326,7 +1326,7 @@ mod tests {
     /// Full CCTP test infrastructure with two chains (simulating Ethereum and Base)
     struct LocalCctp {
         _ethereum_anvil: AnvilInstance,
-        _base_anvil: AnvilInstance,
+        base_anvil: AnvilInstance,
         ethereum_endpoint: String,
         base_endpoint: String,
         ethereum: DeployedCctpChain,
@@ -1432,7 +1432,7 @@ mod tests {
 
             Ok(Self {
                 _ethereum_anvil: ethereum_anvil,
-                _base_anvil: base_anvil,
+                base_anvil,
                 ethereum_endpoint,
                 base_endpoint,
                 ethereum,
@@ -1638,15 +1638,30 @@ mod tests {
             >,
             Box<dyn std::error::Error>,
         > {
+            self.create_bridge_with_key(&self.deployer_key).await
+        }
+
+        /// Builds a bridge whose wallet is `private_key` rather than the deployer.
+        /// Used to simulate a burn submitted by a different depositor so the
+        /// depositor clause in `find_recent_burn`'s predicate can be exercised.
+        async fn create_bridge_with_key(
+            &self,
+            private_key: &B256,
+        ) -> Result<
+            CctpBridge<
+                RawPrivateKeyWallet<impl Provider + Clone + use<>>,
+                RawPrivateKeyWallet<impl Provider + Clone + use<>>,
+            >,
+            Box<dyn std::error::Error>,
+        > {
             let ethereum_provider = ProviderBuilder::new()
                 .connect(&self.ethereum_endpoint)
                 .await?;
 
             let base_provider = ProviderBuilder::new().connect(&self.base_endpoint).await?;
 
-            let ethereum_wallet =
-                RawPrivateKeyWallet::new(&self.deployer_key, ethereum_provider, 1)?;
-            let base_wallet = RawPrivateKeyWallet::new(&self.deployer_key, base_provider, 1)?;
+            let ethereum_wallet = RawPrivateKeyWallet::new(private_key, ethereum_provider, 1)?;
+            let base_wallet = RawPrivateKeyWallet::new(private_key, base_provider, 1)?;
 
             let ethereum = CctpEndpoint::new(
                 self.ethereum.usdc,
@@ -2387,6 +2402,7 @@ mod tests {
             .unwrap();
 
         let other_recipient = address!("0x000000000000000000000000000000000000dEaD");
+
         assert_eq!(
             bridge
                 .find_recent_mint(BridgeDirection::BaseToEthereum, other_recipient, from_block)
@@ -2412,6 +2428,7 @@ mod tests {
             .destination_block(BridgeDirection::BaseToEthereum)
             .await
             .unwrap();
+
         assert_eq!(
             bridge
                 .find_recent_mint(BridgeDirection::BaseToEthereum, recipient, head_above_mint)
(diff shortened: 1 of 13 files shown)


0xgleb:
* **#716** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/716?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#717](https://github.com/ST0x-Technology/st0x.liquidity/pull/717) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/717?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#718](https://github.com/ST0x-Technology/st0x.liquidity/pull/718) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/718?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#725** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/725?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/725?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=725).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/rebalancing/usdc/manager.rs
```diff
@@ -825,6 +831,74 @@ impl<Chain: Wallet> CrossVenueCashTransfer<Chain> {
...
+                        id,
+                        UsdcRebalanceCommand::ConfirmConversion { filled_amount },
+                    )
+                    .await?;
+                info!(target: "rebalance", order_id = %order.id, %filled_amount, "Resumed conversion confirmed from already-filled order");
+                Ok(())
+            }
+            CryptoOrderOutcome::Pending => {
+                warn!(target: "rebalance", order_id = %order.id, "Resumed conversion order still settling; retrying");
+                Err(UsdcTransferError::ConversionStillSettling { id: id.clone() })
```
JuaniRios: critical: returning ConversionStillSettling as an Err models a healthy-but-slow conversion as a failed job attempt, which spends the finite apalis retry budget (retries(3), 1s/2s/4s). If Alpaca keeps the order in new/pending_new/partially_filled longer than ~7s, the job goes terminal-Failed while the aggregate is still Converting, and the fail-stop circuit breaker halts processing. The timeout sweep can then stop seeing a non-terminal row and clear usdc_in_progress, letting a second transfer start over a conversion that may still fill, exactly the partial-completion clobber this PR is meant to prevent. The fresh path (poll_crypto_order_until_filled) polls indefinitely instead. Either poll the looked-up order until terminal inside the attempt, or reschedule without consuming the retry budget, and add a test where lookup stays pending past the budget then fills.

crates/execution/src/alpaca_broker_api/order.rs
```diff
@@ -231,21 +234,46 @@ pub struct CryptoOrderResponse {
...
     }
+
+    /// Classifies the order's current status into a fill/pending/failed outcome,
+    /// reusing the same terminal mapping as `poll_crypto_order_until_filled`.
+    pub fn classify(&self) -> CryptoOrderOutcome {
+        use BrokerOrderStatus::*;
+
+        match self.status {
+            Filled => CryptoOrderOutcome::Filled,
+            Canceled => CryptoOrderOutcome::Failed(CryptoOrderFailureReason::Canceled),
```
JuaniRios: should fix: classify() maps canceled/expired/rejected straight to Failed without looking at filled_quantity. Alpaca terminal statuses can carry a nonzero partial fill, so a partial-then-cancelled conversion gets recorded as a full FailConversion even though real USDC was already converted, leaving inventory and cash state wrong. This mirrors the pre-existing poll loop, but classify() and the resume path are both new, so worth deciding here. Consider preserving a nonzero fill (a TerminalPartiallyFilled { reason, filled_quantity } outcome) and settling it, or document why partial-then-terminal is impossible for these market orders.

crates/execution/src/alpaca_broker_api/order.rs
```diff
@@ -231,21 +234,46 @@ pub struct CryptoOrderResponse {
...
+            Expired => CryptoOrderOutcome::Failed(CryptoOrderFailureReason::Expired),
+            Rejected => CryptoOrderOutcome::Failed(CryptoOrderFailureReason::Rejected),
+            _ => CryptoOrderOutcome::Pending,
```
JuaniRios: minor: the `_ => Pending` wildcard means a future terminal Alpaca status silently becomes Pending and (combined with the still-settling retry path above) gets retried forever instead of failing the transfer. The project prefers exhaustive matches in production so the compiler flags new variants. Enumerating the remaining non-terminal variants explicitly would force a compile error when Alpaca adds a status, and a test for an unknown status mapping to Pending would pin the wildcard behavior.

src/rebalancing/trigger/mod.rs
```diff
@@ -771,6 +771,22 @@ impl RebalancingService {
         };
 
         for id in timed_out_ids {
+            // A timed-out transfer whose apalis row is still in flight is NOT
+            // abandoned: the job will resume and may have an irreversible on-chain
+            // withdraw/burn already submitted. Skip the cleanup entirely so we
+            // neither clear the in-progress guard / inventory inflight nor mark it
+            // timed-out (which would make the reactor ignore its eventual terminal
+            // event and latch the guard forever). Let apalis drive it to terminal.
+            if self.transfer_usdc_to_hedging_in_flight().await {
```
JuaniRios: minor: transfer_usdc_to_hedging_in_flight() queries for any non-terminal TransferUsdcToHedging row regardless of id, but it runs once per timed-out id and the continue skips cleanup per id (the log even prints aggregate_id = %id). It's correct today only because usdc_in_progress admits a single transfer at a time. If that invariant ever relaxes, one in-flight job would defer cleanup of an unrelated abandoned transfer and latch its guard. Scoping the query to the aggregate id (the payload carries it) would remove the implicit coupling.

crates/execution/src/alpaca_broker_api/client.rs
```diff
@@ -207,6 +207,33 @@ impl AlpacaBrokerApiClient {
...
+
+        debug!(
+            "Fetching crypto order by client_order_id {} from {}",
+            client_order_id, url
+        );
+
+        match self.get::<CryptoOrderResponse>(&url).await {
+            Ok(order) => Ok(Some(order)),
+            Err(AlpacaBrokerApiError::ApiError { status, .. })
+                if status == reqwest::StatusCode::NOT_FOUND =>
```
JuaniRios, line 235: should fix: the whole "order never reached Alpaca -> fail for reconciliation" branch hinges on this endpoint returning exactly 404 for an unknown client_order_id. If Alpaca returns 422/400 or an empty 200 instead, the None branch never fires: a 4xx propagates as an error (retries then fails, aggregate stuck in Converting) and an empty 200 fails deserialization, which is the indeterminate stuck state this PR is trying to remove. Worth confirming Alpaca's documented not-found response and broadening the None mapping plus a test if it can be non-404.

crates/bridge/src/cctp/mod.rs
```diff
@@ -2444,4 +2446,94 @@ mod tests {
...
+
+        assert_eq!(
+            found,
+            Some(receipt.tx),
+            "scan must return the real burn's tx for the matching amount + recipient",
+        );
+    }
+
+    #[tokio::test]
+    async fn find_recent_burn_returns_none_for_wrong_amount_or_recipient() {
```
JuaniRios, line 2501: minor: the PR description says find_recent_burn matching was tightened to include depositor, but the matching predicate already had all four fields at this SHA and this diff only adds tests (no evm.rs change). This negative test varies only amount and recipient, so the depositor clause, the one the description emphasizes, has no negative coverage. A regression dropping the depositor check (adopting another sender's same-amount/same-recipient burn) wouldn't be caught. Adding a different-depositor case asserting None, and splitting the wrong-amount and wrong-recipient asserts so a failure points at the right predicate, would close that.

src/rebalancing/trigger/mod.rs
```diff
@@ -7828,6 +7876,70 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn timeout_does_not_clear_guard_while_transfer_job_in_flight() {
```
JuaniRios, line 7896: minor: this test verifies the guard stays latched while the job is in flight, but nothing tests the complementary property that once the job row is Failed/Done the next sweep proceeds to cleanup and clears the guard. That "eventually released, not latched forever" property is the whole safety argument for skipping cleanup; a regression mislisting terminal statuses would silently starve all future Base->Alpaca rebalances uncaught. A test pushing a row, marking it Failed, sweeping, and asserting the guard clears would cover it. The MissingFilledQuantity path in resume_converting (filled order with filled_qty absent) is also currently untested.

src/rebalancing/usdc/manager.rs
```diff
@@ -825,6 +831,74 @@ impl<Chain: Wallet> CrossVenueCashTransfer<Chain> {
...
+            }
+            CryptoOrderOutcome::Failed(reason) => {
+                warn!(target: "rebalance", order_id = %order.id, ?reason, "Resumed conversion order failed terminally");
+                self.cqrs
+                    .send(
+                        id,
+                        UsdcRebalanceCommand::FailConversion {
+                            reason: format!("conversion order failed: {reason:?}"),
```
JuaniRios: nit (optional, matches existing convention so fine to leave): these FailConversion reasons are built with format!, and this one debug-formats a CryptoOrderFailureReason into the string, so stored ConversionFailed.reason holds debug enum names that consumers can't match on. FailConversion { reason: String } is pre-existing and every existing caller already stringifies, so this is out of scope to fix here. Flagging only in case the team later wants to make reason a typed enum, at which point both format! calls would disappear.

JuaniRios (changes requested):

src/rebalancing/usdc/manager.rs
0xgleb: resume_converting now awaits the order to terminal via poll_conversion_to_terminal instead of erroring

src/rebalancing/trigger/mod.rs
0xgleb: `timeout_clears_guard_once_transfer_job_reaches_terminal_status` and z
`resume_base_to_alpaca_from_converting_errors_when_filled_order_lacks_filled_qty` cover the guard-release and `MissingFilledQuantity` paths

crates/bridge/src/cctp/mod.rs
0xgleb: corrected the description

crates/execution/src/alpaca_broker_api/client.rs
0xgleb: confirmed that it's 404

crates/execution/src/alpaca_broker_api/order.rs
0xgleb: Failed branch now surfaces any nonzero partial fill

crates/execution/src/alpaca_broker_api/order.rs
0xgleb: fixed to be exhaustive

src/rebalancing/trigger/mod.rs
0xgleb: transfer_usdc_to_hedging_in_flight now takes the aggregate id and filters job.id == id

JuaniRios (approved):

alastairong1 (approved):

PR #807, opened 2026-06-11
feat: force-fail stuck redemptions at TokensSent/Pending

JuaniRios:
## What

Implements [RAI-983](https://linear.app/makeitrain/issue/RAI-983) — Step 5 of the recovery-CLI epic [RAI-978](https://linear.app/makeitrain/issue/RAI-978). Lets the operator force-fail a redemption stuck past the transfer stage.

## Why

A redemption stuck at `TokensSent` (provider never detected) or `Pending` (Alpaca detected but never completed) had **no force-fail path**: `fail-transfer --type redemption` bailed telling the operator to use `FailDetection`/`RejectRedemption`, neither of which had a CLI. Same no-exit class of bug as RAI-957.

## How

The CLI now dispatches per state instead of bailing:
- `TokensSent` → `FailDetection { failure: Operator }` (→ `DetectionFailed` → `Failed`)
- `Pending` → `RejectRedemption { reason }` (→ `RedemptionRejected` → `Failed`)

Adds a `DetectionFailure::Operator` variant (unit; keeps `Copy`) so an operator-initiated fail is distinguishable from an automated polling timeout in the event log. All via the CQRS aggregate-command flow.

## Testing

Aggregate test `fail_detection_operator_from_tokens_sent_reaches_failed` (Operator path → `Failed`); existing reject-from-`Pending` test covers the other path. `nextest` + `clippy` + `fmt` green.

## Anything else

The free-text `--reason` is persisted on the `Pending`→`RejectRedemption` path; the `TokensSent`→`FailDetection{Operator}` path records the `Operator` marker only (`DetectionFailure` is `Copy`) and surfaces the reason in CLI output. Part of epic [RAI-978](https://linear.app/makeitrain/issue/RAI-978).

PR #782, opened 2026-06-11
feat: alert on low Base-wallet gas via logs and Telegram

JuaniRios:
## What

- Add low-gas observability for the bot's Base wallet: a supervised `GasMonitor` polls native ETH balance, logs ERROR below a threshold (INFO on recovery), and sends a Telegram alert (dedup + recovery notice).
- New optional `[alerts]` config section + `Notifier`/`TelegramNotifier`.
- Disabled by default: when `[alerts]` (config + secret) is absent, the monitor is not spawned.

## Why

- The Base wallet intermittently runs out of ETH gas and silently stalls equity rebalancing (`-32003 insufficient funds for gas`), with no alerting today.
- Closes [RAI-959](https://linear.app/makeitrain/issue/RAI-959).

## How

- `[alerts]` config (mirrors `crates/config/src/telemetry.rs` optional pattern): `chat_id`, `low_balance_threshold` (decimal-ETH string parsed to wei, fail-fast), `poll_interval`, `realert_interval` (seconds, zero rejected); Telegram `bot_token` in secrets.
- `Notifier` trait + `TelegramNotifier` POSTs to Telegram `sendMessage` (typed `NotifierError`).
- `GasMonitor` (`src/conductor/monitor/gas.rs`) uses a pure `evaluate` state machine: alert on transition into low, re-alert at most every `realert_interval`, recovery notice on return to normal; injectable `BalanceReader` for deterministic tests; an RPC read failure logs and is non-fatal.
- Registered as a supervised `gas-monitor` task in `conductor/builder.rs`, gated on `[alerts]` present.
- Base wallet only; the Ethereum-mainnet wallet (CCTP gas) is a documented follow-up.

## Testing

- 18 new tests: notifier (httpmock success/error), the dedup state machine (below/above/suppress-within-interval/realert/recovery/read-failure), and config/assembly (parse/absent/fail-fast-bad-threshold/zero intervals).

## Release Notes

* **New Features**
  * Added optional gas balance monitoring with Telegram notifications for low balance alerts, including configurable thresholds, polling frequency, and alert re-triggering intervals with intelligent de-duplication.

* **Documentation**
  * Added specification and configuration examples for enabling gas balance alerting via Telegram integration.

* **Chores**
  * Updated internal test utilities to support new alerting configuration structure.

PR #781, opened 2026-06-10
feat: reconcile USDC rebalances stuck in terminal DepositFailed

JuaniRios:
## What

- Add operator CLI `reconcile-usdc-transfer --id <id> [--reason funds-moved-manually|deposit-credited-offline]` that drives a USDC rebalance stuck in terminal `DepositFailed` to a new clearing terminal state `Reconciled`, releasing the `usdc_in_progress` guard.
- New aggregate command `ReconcileStuckRebalance`, event `OperatorReconciled`, and terminal state `Reconciled` (`src/usdc_rebalance.rs`).
- Reactor reconciles source-venue USDC inflight when the reconcile lands (`src/rebalancing/trigger/{mod,usdc}.rs`).

## Why

- A post-burn `DepositFailed` permanently blocks **all** USDC rebalancing: `holds_rebalance_guard()` is true for it, so `recover_usdc_guard` re-latches the guard every startup and the timeout sweep preserves it every cycle (`Skipped USDC trigger: already in progress` forever).
- Nothing transitions out of `DepositFailed` (`InitiateDeposit`/`ConfirmDeposit`/`FailDeposit` all reject it); `RecoverBridging` only applies to `BridgingFailed`; `resume-usdc-transfer` just returns the prior failure.
- Closes [RAI-957](https://linear.app/makeitrain/issue/RAI-957).

## How

- `ReconcileStuckRebalance { reason }` is valid only from `DepositFailed` (else `InvalidCommand`) and emits `OperatorReconciled { direction, reason, reconciled_at }`, evolving to terminal `Reconciled`, for which `holds_rebalance_guard()` returns false.
- The reactor classifies `OperatorReconciled` as a Clear terminal: it removes tracking, clears the guard, and zeroes source-venue USDC inflight via `clear_usdc_inflight` — **post-burn semantics: it does not credit `available`**, because the minted funds physically left source. The venue is derived from the event `direction`, so it works with in-memory tracking absent after a restart.
- `reason` is a typed `ReconcileReason` enum (no opaque String). `SCHEMA_VERSION` bumps 4→5; the aggregate uses the default `Retain` policy so existing events replay (no `CompactedSnapshotClear`).
- The CLI builds a standalone `UsdcRebalance` store (mirrors `resume-usdc-transfer`) and rejects an unknown id or a non-`DepositFailed` aggregate before sending.
- `resume-usdc-transfer` now treats a `Reconciled` aggregate as a terminal no-op (`Ok(())`) instead of erroring.

## Testing

[...]

PR #808, opened 2026-06-11
feat: fail orphaned OffchainOrder on position release-hedge

JuaniRios:
## What

Implements [RAI-984](https://linear.app/makeitrain/issue/RAI-984) — Step 6 of the recovery-CLI epic [RAI-978](https://linear.app/makeitrain/issue/RAI-978). `repair fail-pending-offchain-order` now also fails the orphaned `OffchainOrder` aggregate.

## Why

The command cleared the `Position` pending pointer but never touched the `OffchainOrder` aggregate, so a stuck `Pending`/`Submitted` order lingered forever as a live-looking order in the view (`MarkFailed` existed but had no CLI). View and pointer could diverge.

## How

After `PositionCommand::FailOffChainOrder` clears the pointer, the command drives the `OffchainOrder` aggregate to `Failed` via `OffchainOrderCommand::MarkFailed`. Pointer-first and idempotent: an already-terminal (`Filled`/`Failed`) or absent order is reported and left as-is (the two aggregates aren't transactionally atomic). Uses a no-op `RepairOrderPlacer` since `MarkFailed` never places. All via the CQRS aggregate-command flow.

## Testing

Two integration tests: `fail_pending_also_fails_offchain_order_aggregate` (order ends `Failed`) and `fail_pending_leaves_already_terminal_offchain_order_untouched` (idempotent). 11 repair tests green; `clippy` + `fmt` green.

## Anything else

Part of epic [RAI-978](https://linear.app/makeitrain/issue/RAI-978).

### DIFF
--- a/src/cli/repair.rs
+++ b/src/cli/repair.rs
@@ -1,17 +1,44 @@
 //! Repair CLI commands for manually recovering stuck local CQRS state.
 
 use anyhow::{Context, bail};
+use async_trait::async_trait;
 use rain_math_float::Float;
 use sqlx::SqlitePool;
 use std::io::Write;
+use std::sync::Arc;
 
 use st0x_config::ExecutionThreshold;
-use st0x_event_sorcery::StoreBuilder;
-use st0x_execution::{FractionalShares, Symbol};
+use st0x_event_sorcery::{AggregateError, LifecycleError, StoreBuilder};
+use st0x_execution::{FractionalShares, MarketOrder, Symbol};
 
-use crate::offchain::order::OffchainOrderId;
+use crate::offchain::order::{
+    OffchainOrder, OffchainOrderCommand, OffchainOrderError, OffchainOrderId, OrderPlacementResult,
+    OrderPlacer, build_offchain_order_cqrs,
+};
 use crate::position::{Position, PositionCommand};
 
+/// An [`OrderPlacer`] for repair commands that must never place an order:
+/// `MarkFailed` is a pure terminal transition that never touches the placer.
+/// Returns an error on the unreachable placement path rather than panicking.
+struct RepairOrderPlacer;
+
+#[async_trait]
+impl OrderPlacer for RepairOrderPlacer {
+    async fn place_market_order(
+        &self,
+        _order: MarketOrder,
+    ) -> Result<OrderPlacementResult, Box<dyn std::error::Error + Send + Sync>> {
+        Err("repair must not place offchain orders; MarkFailed is terminal-only".into())
+    }
+}
+
+/// Fails a position's pending offchain order pointer and drives the orphaned
+/// `OffchainOrder` aggregate to `Failed`.
+///
+/// Operates directly on the database: the operator must ensure the bot is not
+/// concurrently driving the same order (per the recovery CLI execution-mode
+/// contract), since a fill landing between the state read and the commands
+/// here cannot be guarded against.
 pub(super) async fn fail_pending_offchain_order_command<W: Write>(
     stdout: &mut W,
     pool: &SqlitePool,
@@ -32,12 +59,68 @@ pub(super) async fn fail_pending_offchain_order_command<W: Write>(
         bail!("position {symbol} not found");
     };
 
+    let order = st0x_event_sorcery::load_entity::<OffchainOrder>(pool, &offchain_order_id)
+        .await
+        .context("failed to load offchain order aggregate")?;
+
+    if let Some(existing) = &order {
+        if existing.symbol() != symbol {
+            bail!(
+                "OffchainOrder {offchain_order_id} belongs to {}, not {symbol} -- refusing \
+                 to repair",
+                existing.symbol()
+            );
+        }
+        match existing {
+            OffchainOrder::PartiallyFilled { .. } => {
+                bail!(
+                    "OffchainOrder {offchain_order_id} is PartiallyFilled: shares already \
+                     executed offchain, and failing it would erase that hedge from the \
+                     position. Reconcile the partial fill first."
+                );
+            }
+            OffchainOrder::Filled { .. } => {
+                bail!(
+                    "OffchainOrder {offchain_order_id} is Filled: the hedge executed. This \
+                     command cannot repair a filled order -- reconcile the fill into the \
+                     position instead of failing it."
+                );
+            }
+            OffchainOrder::Pending { .. }
+            | OffchainOrder::Submitted { .. }
+            | OffchainOrder::Failed { .. } => {}
+        }
+    }
+
     match view.pending_offchain_order_id {
         Some(pending) if pending == offchain_order_id => {}
         Some(pending) => {
             bail!("position {symbol} pending offchain order is {pending}, not {offchain_order_id}");
         }
-        None => bail!("position {symbol} has no pending offchain order"),
+        // The pointer is already clear: either a partial prior run (pointer
+        // cleared, order still live) or a completed one. Skip the Position
+        // command and repair the orphaned order so re-runs are safe.
+        None => {
+            // Re-running a fully successful repair lands here. When the
+            // pointed-at order never had an aggregate (a pointer-only clear),
+            // there is nothing left to fix: the system is already consistent,
+            // so "nothing to repair" is the intended terminal outcome rather
+            // than a silent success.
+            if order.is_none() {
+                bail!(
+                    "position {symbol} has no pending offchain order and no OffchainOrder \
+                     aggregate {offchain_order_id} exists -- nothing to repair"
+                );
+            }
+
+            writeln!(
+                stdout,
+                "Position {symbol} pointer already clear; repairing OffchainOrder \
+                 {offchain_order_id}"
+            )?;
+            return fail_offchain_order_aggregate(stdout, pool, order, offchain_order_id, reason)
+                .await;
+        }
     }
 
     position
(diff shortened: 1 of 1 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/808?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#815** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/815?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#818](https://github.com/ST0x-Technology/st0x.liquidity/pull/818) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/818?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#935](https://github.com/ST0x-Technology/st0x.liquidity/pull/935) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/935?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#808** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/808?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/808?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=808).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor (approved):
Approved.

JuaniRios:
@coderabbitai full review

src/cli/repair.rs
```diff
@@ -56,9 +134,151 @@ pub(super) async fn fail_pending_offchain_order_command<W: Write>(
...
+    // projection updates immediately -- a stale 'Submitted' row in the view is
+    // the very symptom this command exists to repair.
+    let placer: Arc<dyn OrderPlacer> = Arc::new(RepairOrderPlacer);
+    let (store, _projection) = build_offchain_order_cqrs(pool, placer)
+        .await
+        .context("failed to build offchain order store")?;
+    let send_result = store
+        .send(
+            &offchain_order_id,
+            OffchainOrderCommand::MarkFailed { error: reason },
```
0xgleb, line 280: should fix: the pre-send re-load guards against `Filled`/`PartiallyFilled`, but `MarkFailed` is a legal transition from `PartiallyFilled` at the aggregate level, so a partial fill landing in the re-load to send gap makes this `send` succeed with `Ok` (not `AlreadyCompleted`), silently overwriting the order to `Failed` and erasing executed hedge shares while the position pointer is already cleared. The `AlreadyCompleted` net below only catches complete fills, so the partial-fill sliver stays unguarded. Consider re-checking state inside the same write path before append, or at minimum add a test that injects `UpdatePartialFill` between re-load and send and asserts the command errors.

src/cli/repair.rs
```diff
@@ -56,9 +134,151 @@ pub(super) async fn fail_pending_offchain_order_command<W: Write>(
         "Failed pending offchain order {offchain_order_id} for {symbol}"
     )?;
 
+    // Pointer-first: the position's pending pointer is cleared above. Now drive
+    // the OffchainOrder aggregate itself to its Failed terminal so it does not
+    // linger as a live-looking order in the view. The two aggregates are not
+    // transactionally atomic (separate CQRS boundaries); this second step is
+    // idempotent -- an already-terminal or absent order is left as-is.
+    fail_offchain_order_aggregate(stdout, pool, order, offchain_order_id, reason).await?;
```
0xgleb, line 142: minor: on the main path the success line `Failed pending offchain order {id} for {symbol}` is printed before this `fail_offchain_order_aggregate` call, so if this step returns `Err` (for example a concurrent fill detected on re-load) the command exits non-zero but the operator already saw a success message. During an incident that ordering is misleading and can mask that the pointer was cleared. Move the success `writeln!` to after this call returns `Ok`, or collapse both steps into a single summary line at the end.

src/cli/repair.rs
```diff
@@ -56,9 +134,151 @@ pub(super) async fn fail_pending_offchain_order_command<W: Write>(
...
+            )?;
+            Ok(())
+        }
+        // The bot can transition the order to a terminal state in the sliver
+        // between the re-load above and this command; a concurrent FAIL is
+        // equivalent to finding it terminal up front, but a concurrent FILL
+        // means the pointer was cleared for an order that actually executed --
+        // surface that as a hard error so the operator reconciles the
+        // position instead of trusting a clean exit.
+        Err(AggregateError::UserError(LifecycleError::Apply(
```
0xgleb, line 298: minor: both branches of this `AlreadyCompleted` arm are untested, because every existing concurrency test trips the pre-send re-load and returns before ever reaching `store.send()`, so the load-to-send window this arm exists for is never exercised. The `Filled`/`PartiallyFilled` escalation here is the last defence against returning `Ok` while a fill executed and the pointer was cleared, and an inverted condition or deleted branch would go unnoticed. Extract the result-handling into a testable helper that takes a pre-built `Result` plus a state loader so both branches can be unit-tested, or at minimum document that the arm has no deterministic coverage and why.

src/cli/repair.rs
```diff
@@ -32,20 +59,71 @@ pub(super) async fn fail_pending_offchain_order_command<W: Write>(
...
         Some(pending) if pending == offchain_order_id => {}
         Some(pending) => {
             bail!("position {symbol} pending offchain order is {pending}, not {offchain_order_id}");
         }
-        None => bail!("position {symbol} has no pending offchain order"),
+        // The pointer is already clear: either a partial prior run (pointer
+        // cleared, order still live) or a completed one. Skip the Position
+        // command and repair the orphaned order so re-runs are safe.
+        None => {
+            if order.is_none() {
```
0xgleb, line 109: minor: there is no test for re-running a successful repair where the pointer is already clear and no `OffchainOrder` aggregate was ever created, which hits this `nothing to repair` bail. The closest test uses a fresh `OffchainOrderId::new()` (a typo'd id), not a genuine re-run of the same id. Add a test that seeds a pending pointer, runs the repair so it clears the pointer with no aggregate, then re-runs with the same id and asserts the resulting behavior, and if `nothing to repair` is the intended re-run outcome, note that here so operators are not surprised by an error on a consistent system.

src/cli/repair.rs
```diff
@@ -56,9 +134,151 @@ pub(super) async fn fail_pending_offchain_order_command<W: Write>(
...
+        ))) => {
+            let terminal =
+                st0x_event_sorcery::load_entity::<OffchainOrder>(pool, &offchain_order_id)
+                    .await
+                    .context("failed to re-load offchain order after concurrent transition")?;
+            match terminal {
+                // Executed shares always escalate: PartiallyFilled cannot
+                // produce AlreadyCompleted today, but if it ever does, the
+                // same pointer-cleared-without-accounting hazard applies.
+                Some(Filled { .. } | PartiallyFilled { .. }) => {
```
0xgleb: minor: the executed-shares-escalate rule is now encoded twice, once in the pre-send re-load match above and again here in the `AlreadyCompleted` recovery match, with slightly different arm groupings. Since this load-bearing financial-safety invariant (executed shares must always escalate) lives in two places, a future change to the escalation rule has to be made in both or they silently diverge and reintroduce the silent-erasure bug. Extract the reloaded-state-to-outcome decision into one small private helper returning an enum like `Escalate`/`BenignTerminal`/`Proceed`, and have both call sites use it so the rule is single-sourced and exhaustively checked once.

src/cli/repair.rs
0xgleb, line 142: minor: there is no end-to-end test for the highest-risk main-path scenario where the pointer is cleared via `FailOffChainOrder`, then a concurrent `CompleteFill` lands before this `fail_offchain_order_aggregate` sends `MarkFailed`, leaving pointer clear plus filled order plus a command error. The inner helper has a direct test, but it bypasses the public command's pointer-clearing step and the initial state checks that route main vs orphan. Add a test driving the full public command: order non-filled at initial load so the upfront check passes, then filled concurrently between the two aggregate writes, asserting the executed-shares error surfaces.

src/cli/repair.rs
```diff
@@ -56,9 +134,151 @@ pub(super) async fn fail_pending_offchain_order_command<W: Write>(
...
+                    bail!(
+                        "OffchainOrder {offchain_order_id} acquired executed shares \
+                         concurrently: the position pointer was cleared without accounting \
+                         the fill -- reconcile the position manually"
+                    );
+                }
+                Some(Failed { .. } | Pending { .. } | Submitted { .. }) | None => {
```
0xgleb: minor: grouping `Pending` and `Submitted` with `Failed` here under a `reached a terminal state concurrently; left as-is` message is misleading, since those two are not terminal. As verified, `MarkFailed` only yields `AlreadyCompleted` from a `Filled` aggregate and `Filled` is terminal, so a post-reload showing `Pending`/`Submitted`/`None` is unreachable in practice rather than a real masked-corruption path, which keeps this from being a correctness bug. Still, splitting the arm so `Some(Failed { .. })` reports the clean concurrent-fail and the non-terminal variants `bail!` as an unexpected invariant violation would make the message accurate and guard against future regressions cheaply.

src/cli/repair.rs
```diff
@@ -208,6 +451,666 @@ mod tests {
...
+            None
+        );
+
+        // The OffchainOrder aggregate itself is now Failed -- no orphan.
+        let order = st0x_event_sorcery::load_entity::<OffchainOrder>(&pool, &order_id)
+            .await
+            .unwrap()
+            .unwrap();
+        assert!(
+            matches!(order, OffchainOrder::Failed { .. }),
```
0xgleb: nit: this happy-path test asserts the aggregate ends `Failed` but the `..` wildcard never checks the `error` field, so if `MarkFailed` accidentally stored an empty string or the wrong variable the audit reason would be blank and this test would still pass. Since the `reason` is the operator's record of why an order was manually killed, that record matters. Destructure the `Failed` variant and assert `error == "operator repair"`, the same way `fail_offchain_order_aggregate_tolerates_concurrent_fail` checks its reason with a let-else.

0xgleb (changes requested):

src/cli/repair.rs
JuaniRios: Documented this explicitly at the re-load site instead of adding an in-process guard: the partial-fill load->send sliver is unguarded by design and closed only by the no-concurrent-bot contract -- the aggregate must keep `MarkFailed` legal from `PartiallyFilled` for the bot's own rejection path, so there is no in-process guard to add without a dedicated operator-only command. Kept the pre-reload partial-fill test as the in-contract guarantee and spelled out the asymmetry (complete fill -> escalates, partial fill -> unguarded). Happy to add the operator-only command if you'd rather close it atomically. d66ecc2e

src/cli/repair.rs
JuaniRios: Moved the "Failed pending offchain order" line to after the aggregate step succeeds, so the operator never sees a success line ahead of a non-zero exit. d66ecc2e

src/cli/repair.rs
JuaniRios: Extracted the escalate/benign-terminal decision into `classify_reloaded_state` and added `classify_reloaded_state_routes_every_variant` covering all three outcomes -- including the `Filled`->Escalate and `Failed`->BenignTerminal branches the `AlreadyCompleted` arm relies on. d66ecc2e

src/cli/repair.rs
JuaniRios: Added `fail_pending_rerun_after_pointer_only_clear_reports_nothing_to_repair` (a genuine same-id re-run) plus a note at the bail that "nothing to repair" is the intended re-run outcome on an already-consistent system. d66ecc2e

src/cli/repair.rs
JuaniRios: Single-sourced via `classify_reloaded_state`, now used at both re-load sites, so the executed-shares-escalate rule lives in exactly one place. d66ecc2e

src/cli/repair.rs
JuaniRios: The deterministic "filled between the two writes" injection would need a production seam in the command, which I've kept out to stay light. Existing coverage exercises both halves: `fail_pending_command_errors_when_orphan_filled_after_pointer_clear` drives the public command's pointer-clear + routing to the executed-shares refusal, and `fail_offchain_order_aggregate_errors_on_concurrent_fill` covers the helper's pre-send re-load guard. Happy to add the seam + e2e test if you'd prefer it over the lighter footprint.

src/cli/repair.rs
JuaniRios: Split the `AlreadyCompleted` recovery arm: `Some(Failed)` reports the clean concurrent-fail, and the non-terminal/None states now bail as an invariant violation (unreachable, since `MarkFailed` only yields `AlreadyCompleted` from a terminal aggregate). d66ecc2e

src/cli/repair.rs
JuaniRios: Now destructures `Failed { error, .. }` and asserts `error == "operator repair"`, matching the reason check in `fail_offchain_order_aggregate_tolerates_concurrent_fail`. d66ecc2e

0xgleb (approved):

PR #813, opened 2026-06-11
feat: add equity transfer reconcile terminal and CLI

JuaniRios:
## What
Implements [RAI-985](https://linear.app/makeitrain/issue/RAI-985) — Step 7 of [RAI-978](https://linear.app/makeitrain/issue/RAI-978). Adds an operator "reconcile" terminal to the equity mint/redemption aggregates and `transfer reconcile --kind mint|redemption`.

## Why
A mint/redemption stuck in `Failed` whose residue was handled out-of-band (wrap-equity/vault-deposit runbook) had no way to be marked resolved — it lingered as `Failed` and could mislead operators/bookkeeping. Mirrors the USDC operator-reconcile (RAI-957), minus the guard/inventory machinery (equity transfers hold no in-progress guard, so this is a pure bookkeeping terminal).

## How
Both `TokenizedEquityMint` and `EquityRedemption` gain a `Reconciled` terminal state, an `OperatorReconciled { reason }` event, and a `Reconcile { reason }` command valid only from `Failed` (rejected elsewhere). `evolve`: `(OperatorReconciled, Failed) -> Reconciled`. Schema bumped 3→4 (additive — old events replay; bump clears stale snapshots). `Reconciled` maps to the `Completed` DTO status (matches the USDC reconcile mapping). CLI: `transfer reconcile` gains `--kind {usdc|mint|redemption}` (mirrors `transfer resume`/`fail`); usdc keeps its typed reason, equity passes a free-text reason; routed through the CQRS aggregate-command flow via a new `reconcile_equity_transfer_command` that preflight-verifies `Failed`.

## Testing
Aggregate tests (reconcile-from-Failed → Reconciled; rejection from non-Failed via `matches!`; DTO maps to Completed) on both aggregates; CLI parse/classify tests for all three kinds + usdc unknown-reason rejection. Full verify: 249 aggregate/cli tests pass, clippy clean (no allows), fmt clean.

## Anything else
Completes the implementable scope of epic [RAI-978](https://linear.app/makeitrain/issue/RAI-978). One new error variant `EquityRedemptionError::NotFailed` (the redemption error enum had no InvalidCommand-style catch-all).

## Summary by CodeRabbit

* **New Features**
  * Added `transfer reconcile` support for stuck equity mints and redemptions, alongside existing USDC reconciliation.
  * Reconciled equity transfers now move to a terminal completed state and appear as completed in status views.

* **Bug Fixes**
[...]

PR #812, opened 2026-06-11
refactor: remove legacy recovery command names

JuaniRios:
## What
[RAI-986](https://linear.app/makeitrain/issue/RAI-986) — repurposed. **Removes the legacy recovery command names entirely** (no back-compat), instead of deprecating them. After Step 1 ([#809](https://github.com/ST0x-Technology/st0x.liquidity/pull/809)) the old names were kept as hidden aliases/flat variants; this drops them so the CLI exposes only the grouped `transfer`/`position`/`view`/`cctp` surface.

## Why
This is an in-development operator tool that is rarely used, so carrying a dual command surface (and deprecation-warning machinery) isn't worth it. Clean single surface.

## How
Removes the 6 hidden flat `Commands` variants (`resume-usdc-transfer`, `reconcile-usdc-transfer`, `fail-transfer`, `recheck-transfer`, `rebuild-view`, `cctp-recover`), the 3 clap aliases (`repair`, `set-position`, `fail-pending-offchain-order`), their `classify_command` arms, and the (originally-planned) deprecation-notice code. The internal `SimpleCommand`/`ProviderCommand` dispatch and `run_*` handlers are untouched — the grouped commands still map to them. `classify_command` dropped under the line limit, so the `#[allow(clippy::too_many_lines)]` is removed too. Legacy-name tests deleted; clap-validation tests for the removed `repair set-position` rewritten under `position set` to preserve coverage.

## Testing
132 `cli::` tests pass; clippy clean (no allows); fmt clean. No legacy command names remain in `src/`.

## Anything else
Part of epic [RAI-978](https://linear.app/makeitrain/issue/RAI-978). (Branch name still says "deprecate" — the change is removal.)

PR #811, opened 2026-06-11
feat: require --reason on grouped destructive recovery verbs

JuaniRios:
## What
Implements [RAI-982](https://linear.app/makeitrain/issue/RAI-982) — Step 3 of [RAI-978](https://linear.app/makeitrain/issue/RAI-978). Makes `--reason` required on the new grouped destructive verbs.

## Why
A defaulted `--reason` ("Manually failed via CLI") is an audit-hostile record and violates the repo's no-silent-defaults rule. The reason is persisted as the audit trail for a manual intervention.

## How
Removes `default_value` from `--reason` on `transfer fail`, `transfer reconcile`, and `position release-hedge` (`position set` already required it). The hidden legacy flat commands (`fail-transfer`, `reconcile-usdc-transfer`) keep their defaults so existing runbooks/scripts still work. Note: `position release-hedge` and its alias `repair fail-pending-offchain-order` are the same clap variant, so the legacy alias now also requires `--reason` (it can't differ from the new name).

## Testing
Parse tests assert the new names error without `--reason` and the legacy flat name keeps its default. 148 `cli::` tests pass; clippy + fmt clean.

## Anything else
Part of epic [RAI-978](https://linear.app/makeitrain/issue/RAI-978).

## Summary by CodeRabbit

* **New Features**
  * Operator-intervention commands now require a non-empty `--reason`, preventing blank or missing reasons at the command line.

* **Bug Fixes**
  * Updated recovery and transfer workflows so grouped destructive actions no longer fall back to default reasons.
  * Preserved legacy default reasons for hidden compatibility commands, while keeping standalone commands working as expected.

* **Tests**
  * Added coverage for required reasons, blank-value rejection, and legacy command compatibility.

PR #810, opened 2026-06-11
feat: add transfer resume --kind equity REST wrapper

JuaniRios:
## What
Implements [RAI-981](https://linear.app/makeitrain/issue/RAI-981) — Step 2 of [RAI-978](https://linear.app/makeitrain/issue/RAI-978). Adds `transfer resume --kind equity` to bulk-resume interrupted equity transfers.

## Why
`POST /transfers/resume` (resumes all interrupted mints + redemptions) had no CLI wrapper — operators had to curl an unauthenticated endpoint by hand.

## How
`transfer resume` gains a required `--kind {usdc|equity}`. `usdc` keeps the per-id local resume (`--id`/`--direction`); `equity` calls the bot's `/transfers/resume` REST endpoint (all-or-nothing, requires the running bot), mirroring `recheck`. `--kind` is required (not defaulted) because clap's `required_if_eq` does not fire on a default value — without that, `transfer resume` with no id would slip past validation. `classify_command` carries an `#[allow(clippy::too_many_lines)]` (approved) as it's one flat match over all commands.

## Testing
classify/parse tests for usdc (provider) and equity (simple) routing, plus `--kind` required and `--kind usdc` requires `--id`. 148 `cli::` tests pass; clippy + fmt clean.

## Anything else
Part of epic [RAI-978](https://linear.app/makeitrain/issue/RAI-978).

PR #815, opened 2026-06-11
feat: surface per-object recovery commands in dashboard

JuaniRios:
## What
Implements [RAI-987](https://linear.app/makeitrain/issue/RAI-987) — Step 9 of [RAI-978](https://linear.app/makeitrain/issue/RAI-978). Turns the dashboard detail modals into a recovery command palette and adds a CLI recovery guide.

## Why
The dashboard only surfaced a recovery command in one narrow case (the "Stranded equity" hint, and only once a transfer reached a recognized failed/stranded terminal). A transfer stuck mid-flight but not yet failed showed no actionable command, so operators had to recall the exact `stox` invocation from memory or hunt through docs.

## How
- `dashboard/src/lib/transfer.ts`: replaces the single `recoveryCommand` with per-object generators (`transferRecoveryCommands`, `tradeRecoveryCommands`) returning every command valid for the object's `kind`+`status` — including the stuck-but-not-failed case — each as `{ command, label, description, mode }` with the **new** grouped command names (`transfer recheck|resume|fail|reconcile --kind ...`, `position release-hedge|set`, `view rebuild`, `process-tx`). Keeps the deployment prefix logic (`stox` vs sim `nix develop … cli --`).
- Transfer + trade detail modals render a "CLI commands" section: one copy-paste block per applicable command with its execution-mode badge (direct-DB vs requires-bot).
- New `cli-command-block.svelte` (reusable copy block) and `recovery-guide.svelte` (persistent button → dialog documenting every command grouped by object), wired into `header-bar.svelte`. The modals and guide read from the same `transfer.ts` source.

## Testing
17 new/updated assertions in `transfer.test.ts` (per-kind/per-status command sets, exact new-name strings, prod+sim prefix behavior, exec-mode flags). `bun run test:run` → 143 passed; eslint/prettier clean on all changed files.

## Anything else
The 4 `svelte-check` errors on this branch are pre-existing `alpacaUsdc` generated-DTO drift in `inventory-cash.*`/`connection.svelte.test.ts` (present on master, files untouched here; CI's DTO-regen step resolves them locally-stale types) — unrelated to this change. `position release-hedge` shows an `<order-id>` placeholder since the trade modal has no offchain order id. Completes epic [RAI-978](https://linear.app/makeitrain/issue/RAI-978).

### DIFF
--- a/crates/dto/src/transfer.rs
+++ b/crates/dto/src/transfer.rs
@@ -122,8 +122,22 @@ pub enum UsdcBridgeStatus {
     Withdrawing,
     Bridging,
     Depositing,
-    Completed { completed_at: DateTime<Utc> },
-    Failed { failed_at: DateTime<Utc> },
+    Completed {
+        completed_at: DateTime<Utc>,
+    },
+    Failed {
+        failed_at: DateTime<Utc>,
+        // True when the failure happened post-burn (`DepositFailed`, a post-burn
+        // `BridgingFailed`, or a `BaseToAlpaca ConversionFailed`) -- the only USDC
+        // failures `transfer reconcile --kind usdc` accepts. False for pre-burn
+        // failures, which strand nothing and the CLI rejects, so the dashboard can
+        // offer reconcile per-object only when this is true. NOTE: a plain `//`
+        // comment, not `///` -- ts-rs emits doc comments into the generated TS
+        // binding, and a field-level doc lands awkwardly inside the object type
+        // literal (`{ failedAt: string, /** ... */ postBurn: boolean }`), which
+        // trips the dashboard's type-aware eslint. Keep the rationale here in Rust.
+        post_burn: bool,
+    },
 }
 
 impl TransferOperation {
@@ -252,6 +266,43 @@ mod tests {
         assert_eq!(serialized["direction"], json!("alpaca_to_base"));
     }
 
+    #[test]
+    fn usdc_bridge_failed_serializes_post_burn_flag() {
+        let post_burn = TransferOperation::UsdcBridge(UsdcBridgeOperation {
+            id: Id::new("bridge-001"),
+            direction: UsdcBridgeDirection::BaseToAlpaca,
+            amount: Usdc::new(float!(1000)),
+            status: UsdcBridgeStatus::Failed {
+                failed_at: Utc::now(),
+                post_burn: true,
+            },
+            started_at: Utc::now(),
+            updated_at: Utc::now(),
+        });
+
+        let serialized = serde_json::to_value(&post_burn).expect("serialization should succeed");
+        assert_eq!(serialized["status"]["status"], json!("failed"));
+        assert_eq!(serialized["status"]["postBurn"], json!(true));
+        assert!(serialized["status"]["failedAt"].is_string());
+
+        let pre_burn = TransferOperation::UsdcBridge(UsdcBridgeOperation {
+            id: Id::new("bridge-002"),
+            direction: UsdcBridgeDirection::AlpacaToBase,
+            amount: Usdc::new(float!(1000)),
+            status: UsdcBridgeStatus::Failed {
+                failed_at: Utc::now(),
+                post_burn: false,
+            },
+            started_at: Utc::now(),
+            updated_at: Utc::now(),
+        });
+
+        let serialized = serde_json::to_value(&pre_burn).expect("serialization should succeed");
+        assert_eq!(serialized["status"]["status"], json!("failed"));
+        assert_eq!(serialized["status"]["postBurn"], json!(false));
+        assert!(serialized["status"]["failedAt"].is_string());
+    }
+
     fn mint_operation(status: EquityMintStatus, updated_at: DateTime<Utc>) -> TransferOperation {
         TransferOperation::EquityMint(EquityMintOperation {
             id: Id::new("mint-001"),
@@ -330,7 +381,29 @@ mod tests {
         assert!(
             bridge_operation(UsdcBridgeStatus::Completed { completed_at: now }, now).is_terminal()
         );
-        assert!(bridge_operation(UsdcBridgeStatus::Failed { failed_at: now }, now).is_terminal());
+        assert!(
+            bridge_operation(
+                UsdcBridgeStatus::Failed {
+                    failed_at: now,
+                    post_burn: true
+                },
+                now
+            )
+            .is_terminal()
+        );
+
+        // `post_burn` gates whether reconcile is offered, not terminality: a
+        // pre-burn failure is just as terminal as a post-burn one.
+        assert!(
+            bridge_operation(
+                UsdcBridgeStatus::Failed {
+                    failed_at: now,
+                    post_burn: false
+                },
+                now
+            )
+            .is_terminal()
+        );
 
         assert!(!bridge_operation(UsdcBridgeStatus::Converting, now).is_terminal());
         assert!(!bridge_operation(UsdcBridgeStatus::Withdrawing, now).is_terminal());
--- a/dashboard/src/lib/components/recovery-guide.svelte
+++ b/dashboard/src/lib/components/recovery-guide.svelte
@@ -1,0 +1,90 @@
+<script lang="ts">
+  import { getSimulateSourceId } from '$lib/env'
+  import { RECOVERY_GUIDE, recoveryModeColor, recoveryModeLabel } from '$lib/transfer'
+
+  let dialogEl: HTMLDialogElement | undefined = $state()
+
+  const open = () => dialogEl?.showModal()
+
+  // A simulation build runs the mock CLI against a /tmp harness config, so the
+  // guide's hardcoded `stox` commands are not runnable as written. Steer the
+  // operator to the per-object modal commands, which carry the right prefix.
+  const isSimulation = getSimulateSourceId() !== null
+</script>
+
+<button
+  class="rounded border bg-background px-2 py-1 text-xs hover:bg-accent"
+  title="CLI recovery guide"
+  onclick={open}
+>
+  CLI recovery guide
+</button>
+
+<dialog
+  bind:this={dialogEl}
+  class="w-full max-w-2xl rounded-lg border bg-card p-0 text-foreground shadow-lg backdrop:bg-black/50"
+  onclick={(event) => {
+    if (event.target === dialogEl) dialogEl.close()
+  }}
+>
+  <div class="flex items-center justify-between border-b px-5 py-3">
+    <div class="text-sm font-semibold">CLI recovery guide</div>
+    <button
+      class="text-lg leading-none text-muted-foreground hover:text-foreground"
+      onclick={() => dialogEl?.close()}
+    >
+      &times;
+    </button>
+  </div>
+
+  <div class="max-h-[70vh] overflow-y-auto px-5 py-4">
+    {#if isSimulation}
+      <div
+        class="mb-4 rounded-md border border-amber-500/40 bg-amber-500/10 px-3 py-2 text-xs text-amber-500"
+      >
+        <span class="font-semibold">Simulation build.</span> The
+        <span class="font-mono">stox</span> commands below are not runnable here -- this build drives
+        the mock CLI against a temporary harness config. Use the per-object recovery commands in each
+        transfer/trade modal instead; they carry the correct prefix.
+      </div>
+    {/if}
+
+    <p class="mb-4 text-xs text-muted-foreground">
+      Every recovery command, grouped by object. The <span class="font-mono">stox</span> prefix is the
+      production wrapper (run on the server via Tailscale ssh). Each command notes whether it mutates
+      the local CQRS state directly or dispatches through the running bot.
+    </p>
+
+    <div class="space-y-5">
+      {#each RECOVERY_GUIDE as group (group.object)}
+        <section>
+          <h3 class="mb-2 font-mono text-sm font-semibold capitalize text-foreground">
+            {group.object}
+          </h3>
+
+          <div class="space-y-3">
+            {#each group.commands as entry (entry.command)}
+              <div class="rounded-md border bg-muted/20 px-3 py-2">
+                <pre
+                  class="overflow-x-auto whitespace-pre-wrap break-all rounded bg-background/80 p-2 font-mono text-[11px] text-foreground">{entry.command}</pre>
+
+                <div class="mt-1.5 grid grid-cols-[auto_1fr] gap-x-3 gap-y-1 text-[11px]">
+                  <span class="text-muted-foreground">What</span>
+                  <span>{entry.description}</span>
+                  <span class="text-muted-foreground">When</span>
+                  <span>{entry.whenToUse}</span>
+                  <span class="text-muted-foreground">Applies to</span>
+                  <span>{entry.appliesTo}</span>
+                  <span class="text-muted-foreground">Mode</span>
+                  <span class={recoveryModeColor(entry.mode).text}>
+                    {recoveryModeLabel(entry.mode)}
+                  </span>
+                </div>
+              </div>
+            {/each}
+          </div>
+        </section>
+      {/each}
+    </div>
+  </div>
+</dialog>
--- a/dashboard/src/lib/components/trade-history-panel.svelte
+++ b/dashboard/src/lib/components/trade-history-panel.svelte
@@ -4,14 +4,20 @@
   import * as Card from '$lib/components/ui/card'
   import * as Table from '$lib/components/ui/table'
   import HoverTooltip from '$lib/components/hover-tooltip.svelte'
+  import CliCommandBlock from '$lib/components/cli-command-block.svelte'
   import type { Position } from '$lib/api/Position'
   import MultiSelect from '$lib/components/multi-select.svelte'
   import { reactive } from '$lib/frp.svelte'
-  import { getApiBaseUrl } from '$lib/env'
+  import {
+    getApiBaseUrl,
+    getExplorerTxUrl,
+    getSimulateBackendPort,
+    getSimulateSourceId
+  } from '$lib/env'
   import { formatUtc, toDatetimeLocal, TIME_PRESETS, FETCH_TIMEOUT_MS, toRfc3339 } from '$lib/time'
   import { formatDecimal } from '$lib/decimal'
-  import { getExplorerTxUrl } from '$lib/env'
   import { equityUsdTooltip } from '$lib/inventory-value'
+  import { tradeRecoveryCommands } from '$lib/transfer'
 
   type TradeEvent = {
     step: string
@@ -676,6 +682,24 @@
             </div>
           {/each}
         </div>
+
+        {@const recoveryCommands = tradeRecoveryCommands({
+          deployment: {
+            simulateSourceId: getSimulateSourceId(),
+            backendPort: getSimulateBackendPort()
+          },
+          symbol: trade.symbol
+        })}
+        {#if recoveryCommands.length > 0}
+          <div class="mt-5 border-t pt-4">
+            <div class="mb-2 text-xs font-semibold text-foreground">CLI commands</div>
+            <div class="space-y-2">
+              {#each recoveryCommands as entry (entry.command)}
+                <CliCommandBlock {entry} />
+              {/each}
+            </div>
+          </div>
+        {/if}
       {/if}
     </div>
   {/if}
--- a/dashboard/src/lib/components/transfer-panel.svelte
+++ b/dashboard/src/lib/components/transfer-panel.svelte
@@ -4,7 +4,9 @@
   import * as Card from '$lib/components/ui/card'
   import * as Table from '$lib/components/ui/table'
   import HoverTooltip from '$lib/components/hover-tooltip.svelte'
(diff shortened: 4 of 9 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/815?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#815** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/815?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#818](https://github.com/ST0x-Technology/st0x.liquidity/pull/818) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/818?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#935](https://github.com/ST0x-Technology/st0x.liquidity/pull/935) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/935?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/815?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#808** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/808?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=815).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

JuaniRios:
@coderabbitai full review

JuaniRios:
@coderabbitai full review

src/usdc_rebalance.rs
```diff
@@ -714,6 +714,28 @@ pub(crate) enum UsdcRebalance {
...
+            Self::DepositFailed { .. }
+            | Self::ConversionFailed {
+                direction: RebalanceDirection::BaseToAlpaca,
+                ..
+            } => true,
+            Self::BridgingFailed {
+                burn_tx_hash,
+                cctp_nonce,
+                ..
+            } => burn_tx_hash.is_some() || cctp_nonce.is_some(),
```
0xgleb, line 734: should fix: this `BridgingFailed` branch is the most regression-prone path in `is_post_burn_failure`, a two-field OR over `burn_tx_hash`/`cctp_nonce`, yet no `to_dto` test exercises it. The PR only covers `DepositFailed` (true) and `WithdrawalFailed` (false via the catch-all), so a future field rename or wrong field tested would silently flip `post_burn` and break the dashboard reconcile gating with no failing test. Add `to_dto` tests for `BridgingFailed { burn_tx_hash: Some, cctp_nonce: None }` (post_burn true) and both-None (post_burn false), plus `ConversionFailed { BaseToAlpaca }` (true) and `AlpacaToBase` (false).

src/usdc_rebalance.rs
```diff
@@ -714,6 +714,28 @@ pub(crate) enum UsdcRebalance {
...
+            _ => false,
```
0xgleb: should fix: this `_ => false` catch-all defeats the compiler's variant check on the owned `UsdcRebalance` enum, which is exactly what the doc comment's "MUST stay in sync with `transition_reconcile_stuck_rebalance`" relies on. A future post-burn failure variant would silently fall through to `false` and diverge from the reconcilable set with no compile error, and the repo guideline requires exhaustive matches in production so new variants are flagged. Replace the wildcard with explicit `false` arms for the remaining in-flight and non-reconcilable failure variants so adding a variant forces a conscious classification.

src/usdc_rebalance.rs
```diff
@@ -714,6 +714,28 @@ pub(crate) enum UsdcRebalance {
 }
 
 impl UsdcRebalance {
+    /// Whether a failed rebalance failed POST-burn -- the only USDC failures
+    /// `transfer reconcile --kind usdc` accepts (a pre-burn failure strands
+    /// nothing and the CLI rejects it). MUST stay in sync with the reconcilable
+    /// set in `transition_reconcile_stuck_rebalance`. A new (non-failure or
+    /// future) state defaults to `false`, i.e. not reconcilable -- the safe
+    /// default for surfacing the command on the dashboard.
+    fn is_post_burn_failure(&self) -> bool {
```
0xgleb, line 723: minor: `is_post_burn_failure` must stay aligned with both `transition_reconcile_stuck_rebalance` and the CLI reconcile preflight, but the only `to_dto` tests cover `DepositFailed` and `WithdrawalFailed`. The discriminating cases, pre-burn vs post-burn `BridgingFailed` and `ConversionFailed` by direction, go through `holds_rebalance_guard` tests on a separate match, so a refactor of one arm without the other would not be caught here. A small set of parametrized `to_dto` tests over each reconcilable/non-reconcilable failure variant would close the gap.

crates/dto/src/transfer.rs
```diff
@@ -330,7 +344,16 @@ mod tests {
         assert!(
             bridge_operation(UsdcBridgeStatus::Completed { completed_at: now }, now).is_terminal()
         );
-        assert!(bridge_operation(UsdcBridgeStatus::Failed { failed_at: now }, now).is_terminal());
+        assert!(
+            bridge_operation(
+                UsdcBridgeStatus::Failed {
+                    failed_at: now,
+                    post_burn: true
```
0xgleb, line 388: minor: this test only passes `post_burn: true` to the `Failed` variant, so it never confirms that a pre-burn failure (`post_burn: false`) is also terminal. `is_terminal()` should be true regardless of `post_burn` because that field gates a separate decision (reconcile availability), and if someone narrowed the pattern to `Failed { post_burn: true, .. }` this test would still pass. Add a matching assert with `post_burn: false` right after this one.

crates/dto/src/transfer.rs
```diff
@@ -122,8 +122,22 @@ pub enum UsdcBridgeStatus {
...
+        // True when the failure happened post-burn (`DepositFailed`, a post-burn
+        // `BridgingFailed`, or a `BaseToAlpaca ConversionFailed`) -- the only USDC
+        // failures `transfer reconcile --kind usdc` accepts. False for pre-burn
+        // failures, which strand nothing and the CLI rejects, so the dashboard can
+        // offer reconcile per-object only when this is true. NOTE: a plain `//`
+        // comment, not `///` -- ts-rs emits doc comments into the generated TS
+        // binding, and a field-level doc lands awkwardly inside the object type
+        // literal (`{ failedAt: string, /** ... */ postBurn: boolean }`), which
+        // trips the dashboard's type-aware eslint. Keep the rationale here in Rust.
+        post_burn: bool,
```
0xgleb, line 139: minor: adding `post_burn` to `Failed` introduces no serialization roundtrip test asserting the API shape (`{ status: "failed", failedAt, postBurn }`) that the dashboard reads via `transfer.status.postBurn`. The reconcile gating depends on `postBurn` surviving JSON plus ts-rs codegen with the camelCase rename, and a serde tagging mistake would silently show or hide reconcile for all failed bridges. Add a test asserting exact `serde_json::json!` literals for `post_burn: true` and `false`.

0xgleb, line 139: minor: `post_burn: bool` encodes a domain distinction (pre-burn vs post-burn failure, which gates whether `transfer reconcile --kind usdc` is offered) as a raw boolean, which the repo's "avoid boolean blindness" guideline discourages for domain states. A two-variant enum like `BurnPhase { PreBurn, PostBurn }` would make the meaning legible at every construction and match site and serialize via ts-rs as a string-literal union instead of a bare bool. This is a DTO-boundary judgement call since the bool maps cleanly to the existing `postBurn?: boolean` the dashboard consumes, so it is acceptable as-is, but worth considering.

dashboard/src/lib/components/trade-history-panel.svelte
```diff
@@ -676,6 +682,29 @@
             </div>
           {/each}
         </div>
+
+        {@const txHash =
+          trade.venue === 'raindex' && trade.id.includes(':')
+            ? trade.id.slice(0, trade.id.lastIndexOf(':'))
+            : null}
+        {@const recoveryCommands = tradeRecoveryCommands({
```
0xgleb, line 686: should fix: `tradeRecoveryCommands` is rendered for every trade detail modal with no gating on trade state, unlike `transferRecoveryCommands` which gates on status. For raindex fills this pre-fills `process-tx --tx-hash <hash>` (plus `position set`/`release-hedge`/`view rebuild`) on trades that were already successfully processed. Re-running `process-tx` on an older hash only gets rejected when the trade is still `last_acknowledged_trade_id`, so on a routine history view it is one click away from re-applying a fill and inflating net exposure. Gate these on actionable state (missed fill / stuck hedge), or at minimum drop per-trade `process-tx` unless the trade is known unaccounted.

dashboard/src/lib/transfer.ts
```diff
@@ -137,55 +147,446 @@ export const stuckReasonLabel = (reason: string): string =>
...
+  const prefix = commandPrefix(params.deployment)
+  if (prefix === null) return []
+
+  const { symbol, txHash } = params
+
+  const commands: RecoveryCommand[] = []
+
+  if (txHash !== null) {
+    commands.push({
+      command: `${prefix} process-tx --tx-hash ${txHash}`,
```
0xgleb: should fix: this surfaces `process-tx` as a prominent copy-paste button, but the underlying `process_found_trade` path goes straight to `AcknowledgeOnChainFill` without consulting the `OnChainTrade` aggregate or the full set of acknowledged trade ids. Duplicate detection is only `last_acknowledged_trade_id == trade_id`, a single-slot guard, so a retry after a later fill was acknowledged double-applies the older trade. Before emitting `AcknowledgeOnChainFill`, check the persisted `OnChainTrade` / fill witness for `(tx_hash, log_index)` and fail closed if already recorded, and align the UI copy with the actual idempotency guarantee.

dashboard/src/lib/transfer.ts
```diff
@@ -137,55 +147,446 @@ export const stuckReasonLabel = (reason: string): string =>
...
-      command: `nix develop --command cargo run --features mock --bin cli -- --config ${basePath}.config.toml --secrets ${basePath}.secrets.toml transfer recheck --kind ${transferType} --id ${params.id}`,
+  if (isTerminalStatus(status)) return []
+
+  // The CLI requires --direction and rejects a mismatch against the persisted
+  // value, so fill the bridge's known direction (translated into the CLI's flag
+  // vocabulary) when we have it rather than leaving a placeholder.
+  const directionArg = usdcDirectionToCliFlag(direction) ?? '<to-raindex|to-alpaca>'
+  return [
+    {
+      command: `${prefix} transfer resume --kind usdc --id ${id} --direction ${directionArg}`,
```
0xgleb, line 422: should fix: this resume command is offered for any non-terminal, non-failed status, which includes pre-burn stages `converting` and `withdrawing` where the bot is still actively driving the Alpaca withdrawal. The description says the bridge was "interrupted after the burn," so it is misleading for those statuses, and the `direct-db-live-rpc` mode means it hits live chain and can race the running bot. Narrow resume to post-burn statuses (`bridging`, `depositing`) and consider surfacing `fail-usdc-transfer` for pre-burn stuck bridges, which currently has no per-object affordance.

dashboard/src/lib/transfer.ts
```diff
@@ -137,55 +147,446 @@ export const stuckReasonLabel = (reason: string): string =>
...
+      },
+      {
+        command: 'stox transfer fail --kind <mint|redemption> --id <id> -r "<reason>"',
+        description: 'Force a stuck transfer into the terminal Failed state.',
+        whenToUse: 'A mint/redemption is permanently stuck and unrecoverable.',
+        appliesTo: 'Equity mint / redemption (non-terminal)',
+        mode: 'direct-db'
+      },
+      {
+        command: 'stox transfer reconcile --kind <usdc|mint|redemption> --id <id> -r "<reason>"',
```
0xgleb, line 551: minor: the guide documents equity `fail` and USDC `resume`/`reconcile` but omits `fail-usdc-transfer`, the documented operator path for clearing pre-burn USDC stuck states (`BridgingSubmitting`, `WithdrawalComplete`). Operators relying on this guide get no guidance for that known stuck state, and the per-object generator omits it too. Add a guide entry (and optionally per-object gating) for the pre-burn USDC fail command with the same execution-mode labeling.

crates/dto/src/transfer.rs
0xgleb, line 388: minor: the Rust `Failed` variant now carries `post_burn`, but the generated `dashboard/src/lib/api/UsdcBridgeStatus.ts` (not in this diff) still shows `{ status: "failed", failedAt }` without `postBurn`, so it will fail svelte-check until CI regenerates DTOs. Nothing currently imports the stale generated type (`transfer-panel.svelte` uses a local ad-hoc shape), so this is developer confusion rather than a runtime break, but the drift count grows. Run the DTO regeneration step locally and commit the regenerated `UsdcBridgeStatus.ts` so the type is self-documenting.

dashboard/src/lib/components/recovery-guide.svelte
```diff
@@ -0,0 +1,75 @@
...
+                <div class="mt-1.5 grid grid-cols-[auto_1fr] gap-x-3 gap-y-1 text-[11px]">
+                  <span class="text-muted-foreground">What</span>
+                  <span>{entry.description}</span>
+                  <span class="text-muted-foreground">When</span>
+                  <span>{entry.whenToUse}</span>
+                  <span class="text-muted-foreground">Applies to</span>
+                  <span>{entry.appliesTo}</span>
+                  <span class="text-muted-foreground">Mode</span>
+                  <span
+                    class={entry.mode === 'requires-bot' ? 'text-amber-500' : 'text-destructive'}
```
0xgleb: minor: this badge uses only two colors, amber for `requires-bot` and red for everything else, so `live-rpc-only`, `direct-db-live-rpc`, and `direct-db` all render identically red. `live-rpc-only` (`cctp complete-mint`) writes no DB state and is safe to run while the bot is up, yet it gets the same warning color as `view rebuild` which mandates stopping the bot. The text label via `recoveryModeLabel` does distinguish them, but consider a third color for `live-rpc-only` (and the same fix in `cli-command-block.svelte`) so the badge does not imply equal severity.

dashboard/src/lib/components/recovery-guide.svelte
```diff
@@ -0,0 +1,75 @@
...
+      onclick={() => dialogEl?.close()}
+    >
+      &times;
+    </button>
+  </div>
+
+  <div class="max-h-[70vh] overflow-y-auto px-5 py-4">
+    <p class="mb-4 text-xs text-muted-foreground">
+      Every recovery command, grouped by object. The <span class="font-mono">stox</span> prefix is the
+      production wrapper (run on the server via Tailscale ssh). Each command notes whether it mutates
```
0xgleb, line 54: minor: `RECOVERY_GUIDE` hardcodes the `stox` prefix and this component renders it with no deployment context, so in a simulation build (`PUBLIC_SIMULATE_SOURCE_ID` set) the guide shows `stox ...` commands that a sim operator cannot run; they need the mock-cli prefix the per-object modals already apply. This prose note partly mitigates it, but a sim operator may still copy a guide command and hit a CLI error. Either make the production-only caveat visually prominent when `isDashboardMockMode()` is true, or pass deployment context in and render via `commandPrefix`.

dashboard/src/lib/components/transfer-panel.svelte
```diff
@@ -48,13 +51,13 @@
...
-    kind: string
+    kind: TransferCategory
     id: string
     symbol?: string
     quantity?: string
     amount?: string
-    direction?: string
-    status: { status: string }
+    direction?: UsdcBridgeDirection
+    status: { status: string; postBurn?: boolean }
```
0xgleb, line 61: minor: widening `TransferEntry.status` to `{ status: string; postBurn?: boolean }` exposes `postBurn` on every transfer kind at the type level, even though it is only meaningful for `kind === 'usdc_bridge'`. Routing is correct at runtime because `transferRecoveryCommands` dispatches on `kind` before consulting `postBurn`, so this is a maintainability concern, not a live bug. Consider narrowing the status to a union that carries `postBurn` only for USDC bridges

0xgleb (changes requested):

src/usdc_rebalance.rs
JuaniRios: Added four `to_dto` tests covering both `BridgingFailed` OR-branches (burn-tx present / absent) and both `ConversionFailed` directions. 35f404ca

src/usdc_rebalance.rs
JuaniRios: Replaced the `_ => false` wildcard with exhaustive arms over all 17 variants, so a new variant forces a conscious post-burn classification instead of silently defaulting to false. 35f404ca

src/usdc_rebalance.rs
JuaniRios: Covered by the four new `to_dto_{bridging,conversion}_failed_*` tests, which exercise both OR-branches and both directions through `to_dto` — matching the file's individual-named-test convention. 35f404ca

crates/dto/src/transfer.rs
JuaniRios: Added the `Failed { post_burn: false }` terminality assertion (post_burn gates reconcile, not terminality). 35f404ca

crates/dto/src/transfer.rs
JuaniRios: Added `usdc_bridge_failed_serializes_post_burn_flag` asserting the `failed` / `postBurn` JSON shape with `json!()` literals (both true and false). 35f404ca

JuaniRios: Leaving as `bool`: it maps 1:1 to the dashboard's existing `postBurn?: boolean` consumer at a pure serialization boundary, and the field comment documents the pre/post-burn meaning. A `BurnPhase` enum would change the wire shape and the consumer for no behavioural benefit.

dashboard/src/lib/components/trade-history-panel.svelte
JuaniRios: Removed `process-tx` from the per-trade modal — a trade only reaches the history panel once recorded, so the modal could only ever re-apply an accounted fill, and the Position single-slot `last_acknowledged_trade_id` guard doesn't catch a re-run on an older trade. It stays in the static recovery guide as CLI reference. 35f404ca

dashboard/src/lib/transfer.ts
JuaniRios: Narrowed `resume` to post-burn (`bridging`/`depositing`); `converting`/`withdrawing` are pre-burn and now offer only `fail-usdc-transfer` (strands nothing, doesn't race the bot). 35f404ca

dashboard/src/lib/transfer.ts
JuaniRios: Added a `fail-usdc-transfer` entry to the recovery guide (pre-burn USDC stuck). 35f404ca

crates/dto/src/transfer.rs
JuaniRios: `src/lib/api/` is gitignored (generated ts-rs bindings, regenerated by the dashboard build/CI), so there's nothing to commit. Nothing imports `UsdcBridgeStatus` either — the panel uses a local shape — so there's no svelte-check break.

dashboard/src/lib/components/recovery-guide.svelte
JuaniRios: Added a shared `recoveryModeColor` helper with a third (sky) colour for `live-rpc-only`, wired into both the command block (badge) and the guide (text). 35f404ca

dashboard/src/lib/components/recovery-guide.svelte
JuaniRios: Added a prominent simulation-build banner (keyed on `getSimulateSourceId()`) telling the operator the `stox` commands aren't runnable there and to use the per-object modal commands. 35f404ca

dashboard/src/lib/components/transfer-panel.svelte
JuaniRios: `TransferEntry` already models all kind-specific data as flat optionals (`symbol`/`quantity`/`amount`/`direction`); `status.postBurn?` follows that pattern and `transferRecoveryCommands` dispatches on `kind` before reading it. Narrowing only `status` into a union would be inconsistent with the rest of the type for a non-bug.

dashboard/src/lib/transfer.ts
JuaniRios: Agreed -- `process_found_trade` only guards on `last_acknowledged_trade_id` (single slot) and can double-apply an older accounted fill. It's a pre-existing CLI-path weakness, not introduced here; T7 removes the one-click dashboard vector, and I've filed RAI-1157 to fail it closed on an already-recorded `(tx_hash, log_index)`.

0xgleb (approved):

PR #820, opened 2026-06-12
feat: hedge-latency read model (reactor-maintained tables)

JuaniRios:
## What

[RAI-990](https://linear.app/makeitrain/issue/RAI-990)

- Adds a "Performance Observability" section to SPEC.md defining the five hedge-pipeline metrics: detection, decision, submission, execution, and total exposure window.
- New `HedgeLatencyProjection` reactor (`src/performance.rs`) subscribing to `Position` and `OffchainOrder` event streams, maintaining three bespoke SQLite tables (`hedge_fill`, `hedge_cycle`, `hedge_attribution_state`).
- Migration `20260616160443_hedge_latency_read_model.sql` creates those tables with indexed lookups by symbol + timestamp.
- Report read path (`load_hedge_performance`) queries only the reactor tables — never folds the `events` table.

## Why

The dashboard's P&L tab covers the economic model; nothing measures the technological one (how fast fills are detected, how long unhedged delta lives). This is the base PR of the Performance-tab stack.

## How

- Reactor attribution rules: onchain fills accumulate in a per-symbol uncovered pool; `OffChainOrderPlaced` consumes the pool as the hedge's covered batch; a failed hedge returns its covered fills to the pool so the retry inherits attribution back to the original fills; a manual adjustment resets the pool and drops in-flight attribution.
- Attribution state is persisted to `hedge_attribution_state` so a restart resumes mid-pool without re-processing history.
- All five latency metrics clamp negatives to zero (block timestamps can sit slightly ahead of the bot's wall clock).
- Malformed read-model rows (unparseable timestamps, non-UUID order IDs) are skipped with a warning so one bad row cannot blank the report.
- Multi-table writes are transaction-wrapped per event for consistency.
- `OffchainOrderId` gains `Hash`, `PartialOrd`, and `Ord` derives to enable the persisted `BTreeMap` attribution state.
- Forward-dead `#[allow(dead_code)]` on the module until the next PR in the stack adds the consuming endpoint, per the stacked-PR policy in `docs/feature-flags.md`.

## Testing

- 29 unit/DB tests: covered-fill batching, failed-hedge attribution restore and retry inheritance, manual-adjustment resets, detection clamp, multi-symbol isolation with deterministic ordering, duplicate-event dedup, malformed read-model rows and out-of-order events for both aggregate types, and the serialize → DB → reactor round trip.

## Anything else

- Read model is forward-only (no startup backfill). A crash between event persistence and the reactor updating its tables can drop one event — accepted as best-effort, matching the `Broadcaster` reactor's guarantees.

## Summary by CodeRabbit

* **New Features**
  * Added performance observability dashboard with hedge operation latency metrics.
  * Introduced real-time tracking of exposure windows, hedge execution latency, and reliability metrics.
  * Performance monitoring now captures fill detection latency, decision/submission timing, and rebalance roundtrip metrics.

### DIFF
--- a/src/conductor.rs
+++ b/src/conductor.rs
@@ -61,6 +61,7 @@ use crate::onchain::approvals::{build_approval_targets, grant_startup_approvals}
 use crate::onchain::backfill::BackfillJobQueue;
 use crate::onchain::trade::{RaindexTradeEvent, extract_owned_vaults, extract_vaults_from_clear};
 use crate::onchain_trade::{OnChainTrade, OnChainTradeCommand, OnChainTradeError, OnChainTradeId};
+use crate::performance::HedgeLatencyProjection;
 use crate::position::{Position, PositionCommand, PositionError, TradeId};
 use crate::rebalancing::equity::{
     CrossVenueEquityTransfer, EquityTransferServices, ResumeTokenizationAggregate,
@@ -487,8 +488,17 @@ impl Conductor {
 
         let order_placer: Arc<dyn OrderPlacer> = Arc::new(ExecutorOrderPlacer(executor.clone()));
 
+        // The HedgeLatencyProjection subscribes to BOTH Position and
+        // OffchainOrder (see its deps!). This instance, registered on the
+        // OffchainOrder store, handles only OffchainOrderEvent::Submitted to
+        // stamp submitted_at. The cycle's authoritative filled_at/failed_at are
+        // stamped by the SAME projection registered on the Position store (the
+        // Position stream carries the broker's own fill timestamp). Do NOT
+        // consolidate the two registrations: routing OffchainOrder Filled/Failed
+        // here would drop the authoritative broker timestamp.
         let (offchain_order, offchain_order_projection) =
             StoreBuilder::<OffchainOrder>::new(pool.clone())
+                .with(Arc::new(HedgeLatencyProjection::new(pool.clone())))
                 .build(order_placer)
                 .await?;
 
@@ -1174,7 +1184,8 @@ fn spawn_rebalancing_infrastructure<Chain: Wallet + Clone>(
         ));
 
         let broadcaster = Arc::new(Broadcaster::new(deps.event_sender, deps.pool.clone()));
-        let manifest = QueryManifest::new(rebalancing_service.clone(), broadcaster);
+        let hedge_latency = Arc::new(HedgeLatencyProjection::new(deps.pool.clone()));
+        let manifest = QueryManifest::new(rebalancing_service.clone(), broadcaster, hedge_latency);
 
         let built = manifest
             .build(deps.pool.clone(), equity_transfer_services)
@@ -1638,6 +1649,7 @@ async fn build_position_cqrs(
 ) -> anyhow::Result<(Arc<Store<Position>>, Arc<Projection<Position>>)> {
     Ok(StoreBuilder::<Position>::new(pool.clone())
         .with(broadcaster)
+        .with(Arc::new(HedgeLatencyProjection::new(pool.clone())))
         .build(())
         .await?)
 }
--- a/src/performance.rs
+++ b/src/performance.rs
@@ -1,0 +1,2031 @@
+//! Hedge-latency read model maintained forward-only by a reactor.
+//!
+//! [`HedgeLatencyProjection`] subscribes to the `Position` and `OffchainOrder`
+//! event streams and maintains three append-only tables (`hedge_fill`,
+//! `hedge_cycle`, `hedge_attribution_reset`) from live events. The report read
+//! path ([`load_hedge_performance`]) queries ONLY those tables and never folds
+//! the `events` table.
+//!
+//! Metric definitions live in SPEC.md under "Performance Observability". The
+//! attribution rules: onchain fills accumulate in a per-symbol uncovered pool;
+//! an `OffChainOrderPlaced` consumes the pool as the placed hedge's covered
+//! batch; a failed hedge returns its covered fills to the uncovered pool so the
+//! retry inherits attribution back to the original fill; a manual adjustment
+//! resets the pool and drops in-flight attribution.
+//!
+//! Crucially there is NO persisted mutable attribution state -- the uncovered
+//! pool is a pure function of the durable, append-only tables, recomputed by
+//! [`uncovered_fills`] on demand (see its doc for the exact algorithm). Both the
+//! read path and the placement writer derive attribution by replaying those
+//! tables, so the read model survives a restart with no in-memory state.
+//!
+//! Forward-only: the reactor processes only events emitted after construction.
+//! There is NO startup backfill of pre-existing history. A crash between an
+//! event being persisted and this reactor's tables being updated can drop that
+//! single event from the read model -- accepted as best-effort, matching the
+//! `Broadcaster` reactor's guarantees.
+
+use std::collections::BTreeMap;
+
+use async_trait::async_trait;
+use chrono::{DateTime, Duration, Utc};
+use sqlx::SqlitePool;
+use thiserror::Error;
+use tracing::{debug, warn};
+
+use st0x_event_sorcery::{EntityList, Reactor, deps};
+use st0x_execution::Symbol;
+
+use crate::offchain::order::{OffchainOrder, OffchainOrderEvent, OffchainOrderId};
+use crate::position::{Position, PositionEvent, TradeId};
+
+/// Per-symbol hedge performance assembled from the read-model tables.
+#[derive(Debug, Clone)]
+pub(crate) struct SymbolPerformance {
+    pub(crate) symbol: Symbol,
+    pub(crate) fills: Vec<FillObservation>,
+    pub(crate) cycles: Vec<HedgeCycle>,
+    pub(crate) open_exposure: Option<OpenExposure>,
+}
+
+/// A single onchain fill as witnessed by the bot.
+#[derive(Debug, Clone)]
+pub(crate) struct FillObservation {
+    pub(crate) block_timestamp: DateTime<Utc>,
+    pub(crate) seen_at: DateTime<Utc>,
+}
+
+impl FillObservation {
+    /// Time from the fill's block to the bot observing it, clamped to zero:
+    /// a freshly mined block can carry a timestamp slightly ahead of the
+    /// bot's wall clock, and a negative latency would corrupt percentile
+    /// aggregations.
+    pub(crate) fn detection_latency(&self) -> Duration {
+        (self.seen_at - self.block_timestamp).max(Duration::zero())
+    }
+}
+
+/// One hedge order and the timestamps along its pipeline.
+///
+/// Because hedges trigger on execution thresholds, a single hedge order may
+/// cover several accumulated onchain fills ([`CoveredFills`]).
+#[derive(Debug, Clone)]
+pub(crate) struct HedgeCycle {
+    pub(crate) offchain_order_id: OffchainOrderId,
+    pub(crate) placed_at: DateTime<Utc>,
+    pub(crate) covered: Option<CoveredFills>,
+    pub(crate) submitted_at: Option<DateTime<Utc>>,
+    pub(crate) outcome: HedgeOutcome,
+}
+
+impl HedgeCycle {
+    /// Observation of the threshold-crossing fill to hedge placement, clamped
+    /// to zero: both timestamps are bot wall-clock values; NTP corrections can
+    /// produce placed_at < latest_seen_at, yielding a negative Duration that
+    /// would corrupt percentile aggregations.
+    pub(crate) fn decision_latency(&self) -> Option<Duration> {
+        let covered = self.covered.as_ref()?;
+        Some((self.placed_at - covered.latest_seen_at).max(Duration::zero()))
+    }
+
+    /// Hedge placement to broker acceptance, clamped to zero for the same
+    /// reason as decision_latency.
+    pub(crate) fn submission_latency(&self) -> Option<Duration> {
+        Some((self.submitted_at? - self.placed_at).max(Duration::zero()))
+    }
+
+    /// Broker acceptance to broker fill, clamped to zero for the same reason
+    /// as decision_latency.
+    pub(crate) fn execution_latency(&self) -> Option<Duration> {
+        let HedgeOutcome::Filled { filled_at } = self.outcome else {
+            return None;
+        };
+        Some((filled_at - self.submitted_at?).max(Duration::zero()))
+    }
+
+    /// Earliest covered fill's block to broker fill: the full period during
+    /// which the system carried unhedged delta for this batch, clamped to
+    /// zero: the broker clock and the block clock are independent time
+    /// sources, and cross-source jitter can produce a negative duration that
+    /// would corrupt percentile aggregations.
+    pub(crate) fn exposure_window(&self) -> Option<Duration> {
+        let HedgeOutcome::Filled { filled_at } = self.outcome else {
+            return None;
+        };
+        let covered = self.covered.as_ref()?;
+        Some((filled_at - covered.earliest_block_timestamp).max(Duration::zero()))
+    }
+}
+
+/// The batch of onchain fills a hedge order covers: every fill observed since
+/// the previous hedge placement (or attribution reset) in the Position stream.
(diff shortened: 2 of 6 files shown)


JuaniRios:
* **#820** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/820?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/820?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=820).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/performance.rs
```diff
@@ -0,0 +1,1714 @@
...
+    .bind(symbol)
+    .bind(json)
+    .execute(&mut **tx)
+    .await?;
+
+    Ok(())
+}
+
+/// Reactor maintaining the hedge-latency read model from live events.
+pub(crate) struct HedgeLatencyProjection {
```
0xgleb, line 466: critical: this isn't a true projection, it's a second source of truth. A projection has to be a pure derivation of the event log (drop the tables, replay events, get identical state back), but this persists mutable attribution state and is forward-only with no replay path, so per the module doc a crash can drop an event permanently. The hedge_attribution_state pool and the open_exposure figures derived from it then live nowhere but these tables and can't be rebuilt from events, which breaks the single-source-of-truth contract in docs/cqrs.md. A reactor-managed multi-aggregate projection is fine, but it needs to be rebuildable by replaying the Position and OffchainOrder streams

src/performance.rs
```diff
@@ -0,0 +1,1714 @@
...
+        offchain_order_id: OffchainOrderId,
+        placed_at: DateTime<Utc>,
+    ) -> Result<(), ProjectionError> {
+        // Load state before opening a transaction so the duplicate-redelivery
+        // guard can return early without a leaked open connection. The
+        // per-aggregate serial-delivery invariant ensures no concurrent writer
+        // races here.
+        let pre_state: Option<AttributionState> = {
+            let row: Option<(String,)> = sqlx::query_as(
+                "SELECT working_state FROM hedge_attribution_state WHERE symbol = ?",
```
0xgleb: should fix: this loads the attribution state outside the transaction (the tx opens further down), while on_chain_fill, offchain_order_filled, and offchain_order_failed all load inside the tx via load_state_tx. The read-modify-write is only safe because of the implicit per-aggregate serial-delivery invariant, so if that's ever relaxed this becomes a TOCTOU that silently clobbers concurrent uncovered-pool writes. The comment justifies it by avoiding a leaked connection, but sqlx auto-rolls-back a dropped transaction so that reasoning doesn't hold. Move the load inside the tx and do the duplicate guard there too.

src/performance.rs
```diff
@@ -0,0 +1,1714 @@
...
+        let symbol: Symbol = match symbol.parse() {
+            Ok(sym) => sym,
+            Err(error) => {
+                warn!(%error, raw_symbol = %symbol, "Attribution state row has invalid symbol, skipping");
+                continue;
+            }
+        };
+
+        let state: AttributionState =
+            match serde_json::from_str(&working_state).inspect_err(|error| {
```
0xgleb: should fix: when the attribution state JSON fails to parse you warn and continue, which skips setting open_exposure. If the symbol already entered by_symbol from the fills or cycles loop it stays in the report with open_exposure = None, silently reporting no open exposure for a symbol that may actually carry unhedged delta. That contradicts the PR's claim that one bad row cannot blank the report, on the most safety-critical metric. Either drop the whole symbol or surface the error instead of partially reporting it.

src/performance.rs
```diff
@@ -0,0 +1,1714 @@
...
+            .find(|perf| perf.symbol == Symbol::new("AAPL").unwrap());
+        if let Some(aapl_perf) = aapl {
+            assert!(aapl_perf.cycles.is_empty());
+        }
+    }
+
+    /// A `hedge_attribution_state` row with malformed JSON is skipped; valid
+    /// data for another symbol is still present.
+    #[tokio::test]
+    async fn malformed_attribution_state_json_in_read_model_is_skipped() {
```
0xgleb: should fix: this asserts aapl.is_none() with a comment saying a malformed state row means the symbol is absent entirely, but that's only true here because AAPL has no fills or cycles. With a hedge_fill row plus a malformed state row, AAPL would be present with open_exposure = None (the read-path bug above), so the comment is misleading. Add a case that injects a fill and a malformed state row and asserts the symbol is present with open_exposure none.

src/performance.rs
```diff
@@ -0,0 +1,1714 @@
...
+            row.map(|(json,)| serde_json::from_str(&json)).transpose()?
+        };
+
+        // Skip state mutation on redelivery: the hedge_cycle INSERT is
+        // idempotent (ON CONFLICT DO NOTHING), and re-draining an already-empty
+        // uncovered pool would clobber the parked batch with an empty Vec,
+        // destroying attribution for a later failure.
+        if pre_state
+            .as_ref()
+            .is_some_and(|state| state.in_flight.contains_key(&offchain_order_id))
```
0xgleb: should fix: this in_flight guard is the subtlest new idempotency path, it stops a redelivered OffChainOrderPlaced from re-draining the pool and clobbering a parked batch, but there's no test for it. Duplicate Submitted and duplicate Filled are tested, not duplicate Placed. Add one: fill then place, push another fill into the pool, redeliver the same Placed, and assert the second fill stays uncovered and the original batch is intact.

src/performance.rs
```diff
@@ -0,0 +1,1714 @@
...
+}
+
+deps!(HedgeLatencyProjection, [Position, OffchainOrder]);
+
+impl HedgeLatencyProjection {
+    pub(crate) fn new(pool: SqlitePool) -> Self {
+        Self { pool }
+    }
+
+    async fn save_state(
```
0xgleb: minor: this save_state carries the same INSERT ON CONFLICT DO UPDATE as save_state_tx, differing only in the execute target. A future column or serialization change touching one and not the other would silently diverge the ManualPositionAdjusted path. Extract one helper over a sqlx Executor, or open a one-statement tx and reuse save_state_tx.

src/performance.rs
```diff
@@ -0,0 +1,1714 @@
...
+    for (symbol, block_timestamp, seen_at) in fill_rows {
+        let symbol: Symbol = match symbol.parse() {
+            Ok(sym) => sym,
+            Err(error) => {
+                warn!(%error, raw_symbol = %symbol, "Fill row has invalid symbol, skipping");
+                continue;
+            }
+        };
+
+        let observation = (|| -> Result<FillObservation, PerformanceError> {
```
0xgleb: nit: the immediately-invoked closure here exists only to get a question-mark scope for the two timestamp parses. The cycle path already uses a named into_cycle constructor for the same shape, so a small parse_fill helper returning Result would read better and match it.

src/performance.rs
```diff
@@ -0,0 +1,1714 @@
...
+        event: OffchainOrderEvent,
+    ) -> Result<(), ProjectionError> {
+        match event {
+            OffchainOrderEvent::Submitted { submitted_at, .. } => {
+                // Keep the first submission timestamp if one was already
+                // recorded, matching the old fold's "keep the first" rule.
+                let id_str = id.to_string();
+                let updated = sqlx::query(
+                    "UPDATE hedge_cycle SET submitted_at = ? \
+                     WHERE offchain_order_id = ? AND submitted_at IS NULL",
```
0xgleb: minor: if a Submitted is delivered before the placement row exists (the two streams are dispatched by independent loops, so a replay can reorder them) this UPDATE matches no rows, you warn, and submitted_at is dropped with no later reconciliation, so submission_latency is missing forever for that cycle. Once the projection is rebuildable this becomes a transient gap, but worth driving submitted_at from a single ordered stream or backfilling it on placement.

src/conductor.rs
```diff
@@ -483,6 +484,7 @@ impl Conductor {
 
         let (offchain_order, offchain_order_projection) =
             StoreBuilder::<OffchainOrder>::new(pool.clone())
+                .with(Arc::new(HedgeLatencyProjection::new(pool.clone())))
```
0xgleb, line 501: minor: HedgeLatencyProjection is built as separate Arc instances for the Position store and the OffchainOrder store, rather than one Arc cloned into both like the Broadcaster pattern. Harmless today since the type only holds a pool, but if it ever gains in-memory state the two instances would diverge silently. Hoist one Arc in Conductor::run and clone it into both builders.

src/performance.rs
```diff
@@ -0,0 +1,1714 @@
...
+        assert_eq!(
+            retry.outcome,
+            HedgeOutcome::Filled {
+                filled_at: timestamp(12)
+            }
+        );
+    }
+
+    #[tokio::test]
+    async fn manual_adjustment_clears_in_flight_attribution() {
```
0xgleb, line 970: nit: this covers OffChainOrderFailed after a manual reset but not the fill path (place, ManualPositionAdjusted, then OffChainOrderFilled). The code handles it correctly today (the in_flight remove is a no-op and filled_at still stamps), but a future refactor gating filled_at on in_flight membership would silently drop the timestamp with no test catching it. Worth adding the complementary case.

src/performance.rs
```diff
@@ -0,0 +1,1714 @@
...
+                warn!(%error, %symbol, "Fill row has unparseable timestamp, skipping");
+            }
+        }
+    }
+
+    let cycle_rows: Vec<CycleRow> = sqlx::query_as(
+        "SELECT offchain_order_id, symbol, placed_at, covered_count, \
+         covered_earliest_block_timestamp, covered_latest_seen_at, \
+         submitted_at, filled_at, failed_at \
+         FROM hedge_cycle ORDER BY placed_at, offchain_order_id",
```
0xgleb: minor: this orders cycles by a TEXT placed_at, which only sorts correctly while every timestamp is the Z-suffixed UTC form. The writer is disciplined (to_rfc3339 on DateTime<Utc>), but there's no CHECK constraint, so an offset-qualified timestamp from any external writer would mis-order silently. Add a CHECK on the timestamp columns, or store unix-seconds integers.

src/performance.rs
```diff
@@ -0,0 +1,1714 @@
...
+        .await;
+
+        let cycle = &performance.cycles[0];
+        assert_eq!(cycle.submitted_at, None);
+        assert_eq!(cycle.execution_latency(), None);
+        assert_eq!(cycle.exposure_window(), Some(Duration::seconds(9)));
+    }
+
+    #[tokio::test]
+    async fn detection_latency_is_seen_minus_block_timestamp() {
```
0xgleb: nit: this asserts a one-line subtraction (seen minus block). The clamp branch is already covered by the clamps_to_zero test and the DB round-trip by every other stream test, so it doesn't add coverage. Either drop it or expand it to assert the full FillObservation round-trip.

src/performance.rs
```diff
@@ -0,0 +1,1714 @@
+//! Hedge-latency read model maintained forward-only by a reactor.
```
0xgleb, line 1: minor: this module is around 1700 lines spanning two separate concerns: the reactor write side (attribution state machine, per-event table writes) and the read side (load_hedge_performance, row structs, parsing, latency metrics). Per AGENTS.md, splitting into a directory once business boundaries emerge applies here. Consider performance/projection.rs for the reactor and performance/report.rs for the read path, with mod.rs holding the public types.

0xgleb (changes requested):

src/performance.rs
JuaniRios: You're right -- this was a second source of truth. Reworked to replay-on-read (`7ecb7315`): dropped `hedge_attribution_state` and the per-event read-modify-write; `open_exposure` is now recomputed on read from the durable, append-only tables `hedge_fill` + `hedge_cycle` + `hedge_attribution_reset` (the last stamps each `ManualPositionAdjusted` -- the one fact not otherwise reconstructable). The uncovered pool is now a pure function of the event-derived tables: drop them and replay and you get identical numbers. This dissolved the related read-modify-write findings (in-tx load, malformed-state blanking, duplicate `save_state`). Module split out in #891.

src/performance.rs
JuaniRios: Dissolved by the replay-on-read rework (`7ecb7315`): there's no persisted attribution state to load anymore, so the load-before-tx and the TOCTOU it risked are gone -- open exposure is recomputed from the durable tables on read.

src/performance.rs
JuaniRios: Dissolved by the rework (`7ecb7315`): `hedge_attribution_state` was removed, so there's no JSON blob to be malformed. Open exposure is recomputed from typed durable columns; a parse error there fails the load rather than silently blanking a symbol.

src/performance.rs
JuaniRios: Addressed (`7ecb7315`): the malformed-attribution-state test was removed with the table; the sibling malformed-fill-timestamp test now correctly asserts the symbol is absent (no state-row lookup keeping it around).

src/performance.rs
JuaniRios: Addressed (`7ecb7315`): with the blob gone, redelivery idempotency is now the cycle insert's `ON CONFLICT(offchain_order_id) DO NOTHING`, covered by the new `duplicate_placement_is_idempotent` test (redeliver a placement; cycle count, covered batch, and open exposure all unchanged).

src/performance.rs
JuaniRios: Dissolved by the rework (`7ecb7315`): both `save_state` and `save_state_tx` were removed -- there's no attribution blob to persist, so the duplicated INSERT is gone.

src/performance.rs
JuaniRios: Fixed (`7ecb7315`): extracted into a named `parse_fill_row` helper, matching `into_cycle`'s style.

src/performance.rs
JuaniRios: Fixed (`7ecb7315`): `submitted_at` now lives in a durable `hedge_submission` table written independently of the cycle row (`INSERT ... ON CONFLICT DO NOTHING`), so a `Submitted` that lands before its placement is preserved; the read path LEFT JOINs it back. Covered by `submitted_before_placed_preserves_submitted_at`.

src/conductor.rs
JuaniRios: Skipping this one. Sharing a single `HedgeLatencyProjection` across both stores requires threading it through `setup` -> `RebalancingDeps` -> `build_position_cqrs`, which is invasive and -- at the bottom of this stack -- re-conflicts the exact conductor wiring the upstack telemetry branches (998/1013) modify. Since the replay-on-read rework made the projection fully stateless, the two `Arc::new` instances are functionally identical and harmless, so the cross-cutting change isn't worth it. Happy to revisit if it ever gains state.

src/performance.rs
JuaniRios: Added (`7ecb7315`): `fill_after_manual_adjustment_records_outcome_and_clears_exposure` covers place -> ManualPositionAdjusted -> filled (the fill path, complementing the fail path).

src/performance.rs
JuaniRios: Fixed (`7ecb7315`): added `CHECK (... LIKE '%+00:00')` on the RFC3339 timestamp columns across the hedge tables, so the TEXT `ORDER BY` is guaranteed lexicographically correct rather than relying on the writer's convention.

src/performance.rs
JuaniRios: Removed (`7ecb7315`): dropped the subtraction-only test; the clamp and round-trip cases already cover the detection-latency logic.

src/performance.rs
JuaniRios: Done in #891 (stacked on top): split into `performance/projection.rs` (write/reactor side) + `performance/report.rs` (read/report side), with shared helpers in the parent. Kept it a separate PR so the ~2.5k-line move stays reviewable apart from the behavioral fixes; the 104 hedge-latency tests moved with their code, unchanged.

0xgleb (approved):

PR #821, opened 2026-06-12
feat: GET /performance/latencies endpoint and DTOs

JuaniRios:
## What

[RAI-991](https://linear.app/makeitrain/issue/RAI-991)

- `GET /performance/latencies`: per-hedge waterfall rows, nearest-rank p50/p90/p95/p99 summaries per pipeline stage, time-bucketed percentile series, and current open exposures.
- New `ReportRange` struct with SQL-pushed time-range filtering (indexed `seen_at`/`placed_at` columns) — no in-memory scan.
- New ts-rs DTOs in `crates/dto/src/performance.rs` (`HedgeLatencies`, `LatencyStats`, `HedgeCycleReport`, `StageLatencies`, …) exported to the dashboard bindings.
- Removes the forward-dead `#[allow(dead_code)]` introduced in the base PR.

## Why

Gives the dashboard a typed API surface over the `HedgeLatencyProjection` read model introduced in the parent PR.

## How

- Read-model queries are SQL-windowed by `seen_at`/`placed_at` using the existing per-column indexes.
- Integer nearest-rank percentiles (no float casts); bucket width mirrors the P&L tab's cadence (daily ≤ 31 days, weekly ≤ 183, else ~monthly).
- Waterfall rows capped at 100 newest-first; pre-cap count returned in `total_cycles`.
- Query params `from`/`to` default to the last 7 days; an inverted range returns 400.
- `offchain_order_id` rides as `Uuid` in the DTO.

## Testing

- Percentile/bucketing unit tests: nearest-rank exactness, single sample, empty, weekly/monthly cadences, truncation cap, negative-sample clamp.
- Endpoint tests: empty DB, seeded fill, inverted range → 400.
- DTO serialization tests asserting wire format against JSON literals, including null optional fields.

## Release Notes

* **New Features**
  * Added a new performance dashboard endpoint for viewing hedge latency metrics, pipeline stage latencies, cycle reports, and current open exposures with customizable time-range filtering (defaults to last 7 days).
  * Implemented latency percentile reporting (p50, p90, p95, p99) with automatic daily/weekly/monthly bucketing based on date range.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -16,12 +16,13 @@ use sqlx::SqlitePool;
 use tokio::sync::Mutex;
 use tracing::{error, info, warn};
 
-use st0x_dto::TradingVenue;
+use st0x_dto::{HedgeLatencies, TradingVenue};
 use st0x_execution::FractionalShares;
 
 use crate::AppState;
 use crate::dashboard::transfer_loader::{InvalidTransferKind, TransferKind};
 use crate::equity_redemption::{EquityRedemptionEvent, RedemptionAggregateId};
+use crate::performance::{ReportRange, hedge_latency_report, load_hedge_performance};
 use crate::rebalancing::RebalancingService;
 use crate::rebalancing::equity::{CrossVenueEquityTransfer, RecheckError, RecheckOutcome};
 use crate::tokenized_equity_mint::{IssuerRequestId, TokenizedEquityMintEvent};
@@ -1487,9 +1488,37 @@ fn recheck_error_response(error: &RecheckError) -> (StatusCode, String) {
     }
 }
 
+/// Query window for `GET /performance/latencies`. Defaults to the last
+/// 7 days ending now.
+#[derive(Debug, Deserialize)]
+struct PerformanceLatenciesQuery {
+    from: Option<DateTime<Utc>>,
+    to: Option<DateTime<Utc>>,
+}
+
+async fn performance_latencies(
+    State(state): State<AppState>,
+    Query(query): Query<PerformanceLatenciesQuery>,
+) -> Result<Json<HedgeLatencies>, StatusCode> {
+    let to = query.to.unwrap_or_else(Utc::now);
+    let from = query.from.unwrap_or(to - chrono::Duration::days(7));
+    if from >= to {
+        return Err(StatusCode::BAD_REQUEST);
+    }
+
+    let report_range = ReportRange { from, to };
+    let performances = load_hedge_performance(&state.pool, &report_range)
+        .await
+        .inspect_err(|error| error!(%error, "Failed to load hedge performance"))
+        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+
+    Ok(Json(hedge_latency_report(&performances, &report_range)))
+}
+
 pub(crate) fn routes() -> Router<AppState> {
     Router::new()
         .route("/health", get(health))
+        .route("/performance/latencies", get(performance_latencies))
         .route("/logs", get(logs))
         .route("/orders/pending", get(pending_orders))
         .route("/trades", get(trades))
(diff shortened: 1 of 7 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/821?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#821** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/821?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/821?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=821).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

JuaniRios:
@coderabbit review

src/performance.rs
```diff
@@ -161,15 +169,28 @@ pub(crate) enum PerformanceError {
...
     pool: &SqlitePool,
+    range: &ReportRange,
 ) -> Result<Vec<SymbolPerformance>, PerformanceError> {
     let mut by_symbol: BTreeMap<Symbol, SymbolAccumulator> = BTreeMap::new();
 
-    let fill_rows: Vec<(String, String, String)> =
-        sqlx::query_as("SELECT symbol, block_timestamp, seen_at FROM hedge_fill ORDER BY id")
-            .fetch_all(pool)
-            .await?;
+    let from = range.from.to_rfc3339();
```
0xgleb, line 186: should fix: this filters by binding to_rfc3339() strings and comparing the stored TEXT with >= / <=, which only works while both sides share the exact same format. Stored seen_at/placed_at come from Utc::now() and carry subseconds plus a +00:00 offset, while a dashboard date picker sends a whole-second Z bound, and lexicographically a stored ...00:00:00.5...+00:00 sorts before the bound ...00:00:00Z (because . is less than Z), so in-range rows get silently dropped. Every test uses zero-fraction timestamps so CI can't see this. Compare numerically instead (strftime('%s', seen_at) against a unix-epoch bound), or load a superset and filter on parsed DateTime like the report layer already does.

src/performance.rs
```diff
@@ -795,6 +819,218 @@ fn covered_fills(uncovered: &[UncoveredFill]) -> Option<CoveredFills> {
...
+
+    fn contains(&self, timestamp: DateTime<Utc>) -> bool {
+        self.from <= timestamp && timestamp <= self.to
+    }
+
+    /// Bucket width mirroring the P&L tab's cadence: daily up to a month,
+    /// weekly up to half a year, otherwise ~monthly.
+    fn bucket_width(&self) -> Duration {
+        let span = self.to - self.from;
+        if span <= Duration::days(31) {
```
0xgleb: should fix: this uses span <= 31 days / <= 183 days on the exclusive to-from duration, but the P&L tab computes an inclusive day count (floor(diff)+1) and thresholds > 31 / > 183. At a span of exactly 31 days this stays daily while P&L switches to weekly, and at exactly 183 days this stays weekly while P&L goes monthly, so the same range shows different granularity on the two tabs even though the docstring says it mirrors P&L. Align on the inclusive day count and add tests at exactly 31 and 183 days.

src/api.rs
```diff
@@ -1894,6 +1923,196 @@ mod tests {
...
+        let ctx = create_test_ctx_with_order_owner(Address::ZERO);
+        let state = empty_app_state(ctx).await;
+        let now = chrono::Utc::now();
+        // The hedge-latency read model now reads the reactor-maintained tables,
+        // not the events table: seed a fill (one detection sample, zero latency)
+        // plus its uncovered attribution state (open exposure), exactly as the
+        // HedgeLatencyProjection reactor would have written them.
+        let timestamp = now.to_rfc3339();
+        sqlx::query(
+            "INSERT INTO hedge_fill (symbol, block_timestamp, seen_at) \
```
0xgleb: should fix: this seeds hedge_fill and hedge_attribution_state with raw SQL and a hand-rolled working_state JSON instead of driving the reactor, while the sibling performance_latencies_reports_nonzero_detection_latency uses ReactorHarness::receive with a real OnChainOrderFilled event. The working_state shape is an implementation detail of the projection, so if its serde layout changes this test keeps passing against stale fixtures while the real write to read path is broken, and it only asserts fillCount == 1 so the attribution path is never exercised. Drive it through ReactorHarness like the adjacent test does.

0xgleb (changes requested):

src/performance.rs
JuaniRios: Addressed via the schema CHECK constraints from the rai-990 rework (`7ecb7315`): every RFC3339 timestamp column now has `CHECK (... LIKE '%+00:00')`, so the TEXT `>=`/`<=` range comparison is guaranteed lexicographically correct rather than relying on convention. The sole writer already used `to_rfc3339()` on `DateTime<Utc>`; the constraint makes a non-UTC value impossible.

src/api.rs
JuaniRios: Updated (`69aec165`): the test no longer seeds `hedge_attribution_state` (removed in the rework -- open exposure is recomputed from `hedge_fill`). It seeds one `hedge_fill` row, which is an uncovered fill => one detection sample (zero latency) + one open-exposure entry, exercising the endpoint end-to-end. I kept the raw-SQL seed of the projection-owned table to match the other read-model tests; happy to switch to a reactor-driven seed if you'd prefer.

0xgleb (approved):

PR #822, opened 2026-06-12
feat: rebalance stage-timing read model and GET /performance/rebalances

JuaniRios:
## What

[RAI-992](https://linear.app/makeitrain/issue/RAI-992)

- `GET /performance/rebalances`: per-operation stage breakdowns of USDC rebalances (conversion, withdrawal, CCTP burn → attestation → mint, deposit), per-stage percentile summaries, and a CCTP attestation-time trend.
- New `RebalanceTimingProjection` reactor (`src/performance/rebalance.rs`) subscribing to the `UsdcRebalance` event stream, maintaining the `rebalance_stage_timing` table (one row per operation keyed by `UsdcRebalanceId`).
- Migration `20260616221000_rebalance_timing_read_model.sql`; range filtering on `started_at` pushed into SQL (indexed).
- New rebalance DTOs in `crates/dto/src/performance.rs` (`RebalanceTimings`, `RebalanceOperationTiming`, `RebalanceStageTiming`, …); `PerformanceRangeQuery` replaces the previous per-endpoint duplicate.

## Why

Full round-trip liquidity time (trade → hedge → inventory restored) was invisible; CCTP attestation is the dominant and most variable stage and had no trend data.

## How

- `fold_operation` runs a small stage tracker over each `UsdcRebalance` stream: direction-specific terminal states close the operation (`DepositConfirmed` for AlpacaToBase, `ConversionConfirmed` for BaseToAlpaca); `AttestationTimedOut` keeps the stage open (recoverable stall); `BridgingFailed` closes open stages with `StageOutcome::Failed`; `BridgingCompletionRecovered` un-fails the operation and records the mint as an unmeasured stage (duration `None`) since its true duration is unknowable.
- `OperatorReconciled` completes stranded operations that would otherwise remain open indefinitely after a manual recovery.
- Operations without a `BridgingSubmitting` intent record the burn as unmeasured rather than a synthetic 0 ms sample that would drag percentiles down.
- Stage durations clamp negatives to zero; malformed rows skipped with a warning.
- Adds `impl From<&RebalanceDirection> for UsdcBridgeDirection` to replace a duplicated inline mapping.

## Testing

- 15+ read-model and report tests: both directions' happy paths with exact stage durations, withdrawal/deposit/conversion failures, timeout-then-attested, failure-then-recovery with `OperatorReconciled`, row cap, range exclusion, negative clamp, DB round trip with stage assertions.
- Endpoint tests: empty DB, seeded, inverted range → 400.
- DTO serialization tests asserting wire format against JSON literals.

## Anything else

- Read model is forward-only (no startup backfill); the endpoint reads the reactor-maintained `rebalance_stage_timing` table — never the `events` table directly. A crash between event persistence and the reactor can drop one event — accepted as best-effort.
- `UsdcBridgeDirection` gains `Copy`/`PartialEq`/`Eq` derives.
- Like the latencies endpoint, this relies on the reactor for all history. Covered by the same follow-up [RAI-1005](https://linear.app/makeitrain/issue/RAI-1005/performance-endpoints-re-fold-full-event-history-per-request).

### DIFF
--- a/crates/dto/src/performance.rs
+++ b/crates/dto/src/performance.rs
@@ -1,11 +1,13 @@
-//! Performance tab DTOs: hedge latency pipeline metrics.
+//! Performance tab DTOs: hedge latency pipeline and rebalance timing metrics.
 
 use chrono::{DateTime, Utc};
 use serde::Serialize;
 use ts_rs::TS;
 use uuid::Uuid;
 
-use st0x_finance::Symbol;
+use st0x_finance::{Symbol, Usdc};
+
+use crate::UsdcBridgeDirection;
 
 /// Response of `GET /performance/latencies`.
 #[derive(Debug, Clone, Serialize, TS)]
@@ -117,6 +119,123 @@ pub struct OpenExposureReport {
     pub oldest_fill_block_timestamp: DateTime<Utc>,
 }
 
+/// Response of `GET /performance/rebalances`.
+#[derive(Debug, Clone, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct RebalanceTimings {
+    /// Most recent operations (stage breakdown rows), newest first.
+    pub operations: Vec<RebalanceOperationTiming>,
+    /// Total operations in range before the `operations` cap was applied.
+    #[ts(type = "number")]
+    pub total_operations: usize,
+    /// Operations dropped because their stored timing JSON failed to
+    /// deserialize. They are excluded from every field above; this count
+    /// surfaces the gap so a malformed read-model row is not silently invisible.
+    #[ts(type = "number")]
+    pub skipped_operations: u32,
+    /// Percentiles per stage across all operations in range.
+    ///
+    /// SPARSE: a stage with no completed (`Succeeded`) samples in range is
+    /// omitted entirely rather than emitted with an empty/zero entry. Consumers
+    /// must look stages up by name and tolerate absence, not index positionally.
+    pub stage_summary: Vec<RebalanceStageStats>,
+    /// CCTP attestation duration over time, oldest first.
+    pub attestation_trend: Vec<AttestationSample>,
+}
+
+/// One USDC rebalance operation's per-stage timing breakdown.
+#[derive(Debug, Clone, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct RebalanceOperationTiming {
+    #[ts(type = "string")]
+    pub operation_id: Uuid,
+    pub direction: Option<UsdcBridgeDirection>,
+    #[ts(type = "string | null")]
+    pub amount: Option<Usdc>,
+    /// Genuine operation start (the first-phase conversion or withdrawal event).
+    /// `null` when the read model first observed the operation mid-stream
+    /// (e.g. after a deploy), in which case its start time is unknown and
+    /// `total_ms` is unmeasured.
+    pub started_at: Option<DateTime<Utc>>,
+    pub completed_at: Option<DateTime<Utc>>,
+    pub status: RebalanceTimingStatus,
+    pub stages: Vec<RebalanceStageTiming>,
+    /// Genuine start to terminal success, when both endpoints are known.
+    /// `null` when the operation is unfinished, or when `started_at` is unknown
+    /// (mid-stream first observation), or when completion was an out-of-band
+    /// `OperatorReconciled` (whose manual-response window must not pollute
+    /// round-trip latency metrics).
+    #[ts(type = "number | null")]
+    pub total_ms: Option<i64>,
+}
+
+/// Where a rebalance operation currently stands.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, TS)]
+#[serde(rename_all = "snake_case")]
+pub enum RebalanceTimingStatus {
+    InProgress,
+    Completed,
+    Failed,
+}
+
+/// Timing of one stage within a rebalance operation.
+#[derive(Debug, Clone, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct RebalanceStageTiming {
+    pub stage: RebalanceStageName,
+    pub started_at: DateTime<Utc>,
+    pub ended_at: Option<DateTime<Utc>>,
+    #[ts(type = "number | null")]
+    pub duration_ms: Option<i64>,
+    pub outcome: StageOutcome,
+}
+
+/// How a rebalance stage run ended.
+///
+/// Only `Succeeded` runs carry a meaningful `duration_ms` for percentile
+/// aggregation. `Failed` runs are timed but excluded from latency stats so a
+/// failure does not contaminate the metric. `Unmeasured` runs demonstrably
+/// happened but have no measurable duration (e.g. a burn with no submitting
+/// intent, or a mint discovered after the fact by recovery); they carry a
+/// `null` `duration_ms` and never enter percentiles. An in-progress stage that
+/// has not yet ended also reports `Unmeasured` until it closes.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, TS)]
+#[serde(rename_all = "snake_case")]
+pub enum StageOutcome {
+    Succeeded,
+    Failed,
+    Unmeasured,
+}
+
+/// Stages of the USDC rebalance pipeline, in flow order.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, TS)]
+#[serde(rename_all = "snake_case")]
+pub enum RebalanceStageName {
+    Conversion,
+    Withdrawal,
+    Burn,
+    Attestation,
+    Mint,
+    Deposit,
+}
+
+/// Percentiles for one rebalance stage.
+#[derive(Debug, Clone, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct RebalanceStageStats {
+    pub stage: RebalanceStageName,
+    pub stats: LatencyStats,
+}
+
+/// One completed CCTP attestation: burn time and how long Circle took.
+#[derive(Debug, Clone, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct AttestationSample {
+    pub burned_at: DateTime<Utc>,
+    #[ts(type = "number")]
+    pub duration_ms: i64,
+}
+
 #[cfg(test)]
 mod tests {
     use serde_json::json;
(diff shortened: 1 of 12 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/822?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#822** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/822?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/822?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=822).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

JuaniRios:
@coderabbit review

JuaniRios:
@coderabbit full review

src/performance/rebalance.rs
```diff
@@ -0,0 +1,1209 @@
...
+        operation_id: UsdcRebalanceId,
+        event: UsdcRebalanceEvent,
+    ) -> Result<(), ProjectionError> {
+        // The first event seen for an operation seeds its started_at.
+        let mut operation = self
+            .load_operation(&operation_id)
+            .await?
+            .unwrap_or_else(|| StoredOperation::new(operation_id.clone(), event_timestamp(&event)));
+        operation.apply(&operation_id, &event);
+        self.save_operation(&operation).await
```
0xgleb: should fix: on_event does load_operation then save_operation as separate statements with no enclosing transaction, and the event-sorcery ReactorBridge logs-and-continues on a reactor Err rather than propagating it. So a transient save failure (db contention, disk full) silently drops the event from rebalance_stage_timing and the read model diverges from the event log permanently, with no retry. Wrapping the load-apply-save in one transaction makes the write atomic, and a startup catch-up would let a dropped event recover.

src/performance/rebalance.rs
```diff
@@ -0,0 +1,1209 @@
...
+//! withdrawal, CCTP burn -> attestation -> mint, deposit). The report read
+//! path ([`load_rebalance_timings`]) queries ONLY that table and never folds
+//! the `events` table.
+//!
+//! Each operation's accumulated state is stored as a JSON [`StoredOperation`]
+//! so the reactor can apply one event at a time (load -> apply -> upsert),
+//! reproducing the same stage breakdown the old fold computed over a full
+//! stream. The read path converts each stored operation into the dashboard DTO.
+//!
+//! Forward-only: the reactor processes only events emitted after construction.
```
0xgleb, line 15: should fix: this is the same forward-only, no-replay concern raised on the hedge-latency read-model PR. The read path queries only this table and never folds events, and there is no backfill or rebuild, so dropping the table or any drift is unrecoverable, which breaks the projection-replay contract in docs/cqrs.md. It is milder than the sibling because each row is a pure single-aggregate fold of one UsdcRebalance stream, so a rebuild_all that re-folds every stream from sequence 0 is cheap to add here. SPEC.md sanctions forward-only for phase-1 and RAI-1005 tracks it, so the call is whether to add the rebuild now or accept the spec posture, but it is worth making that an explicit decision rather than implicit.

src/performance/rebalance.rs
```diff
@@ -0,0 +1,1209 @@
...
+    )
+    .bind(&from)
+    .bind(&to)
+    .fetch_all(pool)
+    .await?;
+
+    let operations = rows
+        .into_iter()
+        .filter_map(
+            |(operation_id, timing)| match serde_json::from_str(&timing) {
```
0xgleb: should fix: a row whose timing JSON fails to deserialize is silently skipped with just a warning, so it vanishes from total_operations, stage_summary and attestation_trend with nothing surfaced to the caller. Partial corruption then produces a quietly wrong dashboard rather than an error. Either fail the request, or return a skipped_operations count so the undercount is visible.

src/performance/rebalance.rs
```diff
@@ -0,0 +1,1209 @@
...
+    async fn on_event(
+        &self,
+        operation_id: UsdcRebalanceId,
+        event: UsdcRebalanceEvent,
+    ) -> Result<(), ProjectionError> {
+        // The first event seen for an operation seeds its started_at.
+        let mut operation = self
+            .load_operation(&operation_id)
+            .await?
+            .unwrap_or_else(|| StoredOperation::new(operation_id.clone(), event_timestamp(&event)));
```
0xgleb: should fix: when the reactor first sees an operation it seeds started_at from that event's timestamp, not the operation's true first event. On a deploy while a transfer is mid-flight, the first event seen might be BridgeAttestationReceived, so started_at becomes attested_at and total_ms = completed_at minus attested_at, understating the headline round-trip latency and mis-bucketing the op in the started_at range filter. The OperatorReconciled-first case is handled, but DepositFailed, BridgingInitiated and others are not. A startup backfill from sequence 1 is the real fix.

src/performance/rebalance.rs
```diff
@@ -0,0 +1,1209 @@
...
+                // Funds were settled out-of-band: the operation is resolved
+                // at reconciliation time, but no pipeline stage ran -- the
+                // failed Deposit stage stays failed so percentiles ignore it.
+                self.status = StoredStatus::Completed;
+                self.completed_at = Some(*reconciled_at);
+            }
+        }
+    }
+
+    fn open(&mut self, stage: StoredStageName, at: DateTime<Utc>) {
```
0xgleb: should fix: open and record_unmeasured push a stage unconditionally, so apply is not idempotent. It is safe today only because cqrs-es delivery is at-most-once, but that couples to the forward-only point above: the natural fix (a catch_up or replay) would re-apply already-reflected events onto the accumulated row and double-count stages in the percentiles and attestation trend. Either make apply idempotent per stage and marker, or have replay rebuild each row from scratch. There is also no test delivering the same event twice to guard this.

src/performance/rebalance.rs
```diff
@@ -0,0 +1,1209 @@
...
+                amount,
+                reconciled_at,
+                ..
+            } => {
+                self.direction.get_or_insert_with(|| direction.clone());
+                self.amount.get_or_insert(*amount);
+                // Funds were settled out-of-band: the operation is resolved
+                // at reconciliation time, but no pipeline stage ran -- the
+                // failed Deposit stage stays failed so percentiles ignore it.
+                self.status = StoredStatus::Completed;
```
0xgleb, line 575: should fix: OperatorReconciled sets completed_at to reconciled_at, which is a manual operator action that can land hours or days after the operation started, so total_ms folds the human response window into the round-trip liquidity metric and the op renders as Completed with a red failed Deposit stage. BridgingCompletionRecovered takes the opposite approach and leaves total_ms None. Worth deciding explicitly: either exclude reconciled ops from total_ms, or give them a distinct status the dashboard can separate.

src/performance/rebalance.rs
```diff
@@ -0,0 +1,1209 @@
...
+        operation
+            .stages
+            .iter()
+            .find(|stage| stage.stage == name)
+            .unwrap()
+    }
+
+    /// Fold an event stream by applying each event to a fresh stored operation,
+    /// exactly as the reactor would, then convert to the dashboard DTO.
+    fn fold_operation(
```
0xgleb: minor: most fold tests use this in-memory helper, so only two tests exercise the real reactor load to apply to save to reload cycle and both run only the happy path. Bugs in the JSON round-trip or the upsert (for example started_at being rewritten on every conflict) would not be caught. Promoting the recovery, timeout and reconcile cases to run_through_reactor would cover the path that actually runs in production.

src/performance/rebalance.rs
```diff
@@ -0,0 +1,1209 @@
...
+            ..
+        } => *deposit_initiated_at,
+        UsdcRebalanceEvent::DepositConfirmed {
+            deposit_confirmed_at,
+            ..
+        } => *deposit_confirmed_at,
+    }
+}
+
+#[cfg(test)]
```
0xgleb, line 782: should fix: the riskiest paths are untested. There is no coverage for a redelivered event, for the reactor first observing an operation mid-stream (the started_at seeding above), or for OperatorReconciled arriving with stages still open (the arm does not close them). AGENTS.md requires coverage for new logic, and these are exactly the forward-only and at-least-once edges the design makes hardest to get right.

crates/dto/src/performance.rs
```diff
@@ -117,6 +119,88 @@ pub struct OpenExposureReport {
...
+    pub stage_summary: Vec<RebalanceStageStats>,
+    /// CCTP attestation duration over time, oldest first.
+    pub attestation_trend: Vec<AttestationSample>,
+}
+
+/// One USDC rebalance operation's per-stage timing breakdown.
+#[derive(Debug, Clone, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct RebalanceOperationTiming {
+    pub operation_id: String,
```
0xgleb: should fix: operation_id is a String here while the analogous HedgeCycleReport::offchain_order_id in this same file is a Uuid, and uuid is already a dependency. A raw String gives no construction-time guarantee the value is a valid UUID and breaks the file's own convention. into_dto can hand the inner Uuid from UsdcRebalanceId straight through instead of calling to_string.

crates/dto/src/performance.rs
```diff
@@ -117,6 +119,88 @@ pub struct OpenExposureReport {
...
+/// Timing of one stage within a rebalance operation.
+#[derive(Debug, Clone, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct RebalanceStageTiming {
+    pub stage: RebalanceStageName,
+    pub started_at: DateTime<Utc>,
+    pub ended_at: Option<DateTime<Utc>>,
+    #[ts(type = "number | null")]
+    pub duration_ms: Option<i64>,
+    pub failed: bool,
```
0xgleb: minor: failed: bool erases the StageOutcome enum that already exists internally, and it cannot express the third state (record_unmeasured sets failed false to mean neither succeeded nor failed). The not-failed filters work today only because unmeasured stages also carry a None duration. Promoting StageOutcome (Succeeded, Failed, Unmeasured) into the DTO would make the read sites and the unmeasured case explicit.

src/performance/rebalance.rs
```diff
@@ -0,0 +1,1209 @@
...
+    RebalanceTimings {
+        operations,
+        total_operations,
+        stage_summary,
+        attestation_trend,
+    }
+}
+
+/// Percentiles per stage over completed, non-failed stage runs.
+fn stage_summary(operations: &[RebalanceOperationTiming]) -> Vec<RebalanceStageStats> {
```
0xgleb: minor: stage_summary drops a stage entirely when it has no completed non-failed samples, so the returned array is variable length zero to six rather than a fixed grid. There is no consumer yet, but the dashboard will need to handle missing stages. Worth either documenting the sparse contract on the DTO or always emitting six rows with sample_count zero.

src/performance/rebalance.rs
```diff
@@ -0,0 +1,1209 @@
...
+        use StoredStageName::*;
+
+        match event {
+            UsdcRebalanceEvent::ConversionInitiated {
+                direction,
+                amount,
+                initiated_at,
+                ..
+            } => {
+                self.direction.get_or_insert_with(|| direction.clone());
```
0xgleb: nit: RebalanceDirection is a fieldless two-variant enum but is not Copy, so the fold clones it at four get_or_insert_with sites while the adjacent amount uses get_or_insert(*amount) because Usdc is Copy. Adding Copy to the derive and switching to get_or_insert(*direction) removes the clones and reads uniformly.

src/performance/rebalance.rs
```diff
@@ -0,0 +1,1209 @@
...
+    amount: Option<Usdc>,
+    started_at: DateTime<Utc>,
+    completed_at: Option<DateTime<Utc>>,
+    status: StoredStatus,
+    stages: Vec<StoredStage>,
+}
+
+/// Operation status, mirroring [`RebalanceTimingStatus`] with serde support.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+enum StoredStatus {
```
0xgleb, line 321: nit: StoredStatus and StoredStageName are exact one-to-one mirrors of the DTO enums with hand-written From impls, existing only because the DTO enums derive Serialize but not Deserialize. Keeping StoredOperation decoupled is fine, but these two leaf enums buy nothing from it and a new variant has to be added in three places or the conversion silently drifts. Adding Deserialize to the DTO enums would delete both mirrors and their From impls.

0xgleb (changes requested):

src/performance/rebalance.rs
JuaniRios: Fixed: `on_event` now wraps load -> apply -> save in a single transaction (`load_operation_tx`/`save_operation_tx`), so a transient save failure can't drop the event after a partial apply.

src/performance/rebalance.rs
JuaniRios: Valid -- it's the same forward-only/no-replay class as the hedge-latency model (which was reworked to replay-on-read). I've fixed the concrete COUPLED issues here -- the read-modify-write is now in one transaction, `apply` is idempotent under redelivery, and malformed rows are surfaced via a `skipped_operations` count instead of silently dropped -- and filed the full replay-on-read rebuild as a tracked follow-up: [RAI-1119](https://linear.app/makeitrain/issue/RAI-1119). Resolving against that issue.

src/performance/rebalance.rs
JuaniRios: Fixed: the read path now counts undeserializable rows and surfaces them as `skipped_operations` on the report DTO, so partial corruption is visible rather than silently missing from the totals.

src/performance/rebalance.rs
JuaniRios: Fixed: `started_at` is now `Option`, seeded ONLY by genuine start events (ConversionInitiated / WithdrawalSubmitting / Initiated -- all carry the operation's true start). When the reactor first observes an operation mid-stream (e.g. `BridgeAttestationReceived` after a deploy), `started_at` stays `None` and `total_ms` is unmeasured rather than fabricated from the mid-stream timestamp. A separate always-present `first_seen_at` backs the SQL sort/window column.

src/performance/rebalance.rs
JuaniRios: Fixed: `apply` is now idempotent under redelivery -- `open_once` won't double-open a stage and the close / record_unmeasured arms no-op on an already-terminal run. Added redelivery tests (in-memory apply-twice + full reactor->DB->reload).

src/performance/rebalance.rs
JuaniRios: Fixed: operator-reconciled operations are flagged and EXCLUDED from the round-trip `total_ms` latency metric (they stay Completed in the report but out of the percentiles), so the manual-reconcile response window no longer folds into the liquidity metric.

src/performance/rebalance.rs
JuaniRios: Addressed: added reactor-cycle (load->apply->save->reload) coverage for the riskier paths -- redelivery and mid-stream first-observe -- not just the in-memory fold helper.

src/performance/rebalance.rs
JuaniRios: Addressed: added tests for a redelivered event, the reactor first observing an operation mid-stream (started_at seeding -> None), and OperatorReconciled handling (started_at/total_ms unmeasured).

crates/dto/src/performance.rs
JuaniRios: Fixed: `RebalanceOperationTiming.operation_id` is now `Uuid` (matching `HedgeCycleReport::offchain_order_id`); `into_dto` passes the inner `Uuid` from `UsdcRebalanceId` directly.

crates/dto/src/performance.rs
JuaniRios: Fixed: promoted a `StageOutcome { Succeeded, Failed, Unmeasured }` enum into the DTO, replacing `failed: bool`. The not-failed filters now match `Succeeded`; open/in-progress stages report `Unmeasured` until they close (carrying a null duration that never enters percentiles).

src/performance/rebalance.rs
JuaniRios: Addressed: documented the SPARSE-array contract on the report DTO -- `stage_summary` omits stages with no completed (`Succeeded`) samples, so consumers must look stages up by name and tolerate absence rather than index positionally.

src/performance/rebalance.rs
JuaniRios: Fixed: `RebalanceDirection` now derives `Copy`; the `get_or_insert_with(|| direction.clone())` sites use `get_or_insert(*direction)`, dropping the clones.

src/performance/rebalance.rs
JuaniRios: Addressed with a note: kept `StoredStatus`/`StoredStageName` as the persisted leaf enums (they exist because the DTO enums derive Serialize+TS but not Deserialize) and added a comment explaining why. The `From` impls are exhaustive matches, so a new variant is a compile error rather than a silent gap.

0xgleb (approved):

PR #823, opened 2026-06-12
feat: reliability metrics endpoint

JuaniRios:
## What

[RAI-993](https://linear.app/makeitrain/issue/RAI-993/reliability-metrics-endpoint-errors-lifecycle-failures-job-queue)

- `GET /performance/reliability`: error/warning log volume (time-bucketed + per-target counts), money-at-risk lifecycle failure events, and apalis job-queue health per job type.

## Why

- Errors were scattered across log files, terminal `Failed` events, and the jobs table with no aggregated view separating money-at-risk failures from cosmetic noise.

## How

- 13 failure event types are counted: hedge failures, all USDC-rebalance stage failures/timeouts, and equity mint/redemption failures and rejections. A `LifecycleFailureProjection` reactor maintains a `lifecycle_failure_event` read model (forward-only — historical events before wiring are not backfilled). A drift test pins representative strings from each event family against `DomainEvent::event_type()` so renames can't silently drop coverage. The query is capped at the 10k newest rows.
- Log aggregation reuses the `/logs` file-reading machinery, runs in `spawn_blocking` (sync file I/O off the async workers), and caps at 50k entries with a truncation warning so an error storm cannot exhaust the dashboard process.
- `JobQueueHealth` distinguishes apalis semantics: retry-eligible `Failed` jobs (`attempts < max_attempts`) surface as `awaiting_retry` live backlog and feed `oldest_pending_run_at`; exhausted `Failed` jobs (`attempts >= max_attempts`) count as `failed`; status `Killed` counts separately.

## Testing

- Unit tests for log aggregation (multi-day buckets, level discrimination, malformed/out-of-range entries, phantom-bucket prevention), failure-event counting with range filtering, the event-type drift test, and job-queue aggregation incl. Queued status and retry/exhausted split; endpoint tests for empty DB, seeded data, and inverted range.

## Anything else

- Stringly-typed `level`/`event_type` DTO fields are tracked as [RAI-1006](https://linear.app/makeitrain/issue/RAI-1006/type-reliability-dto-level-and-failure-event-type-fields-as-enums).

## Summary by CodeRabbit

* **New Features**
  * Added a new `/performance/reliability` endpoint providing performance and reliability metrics, including time-series log volume data, per-target log counts, lifecycle failure event tracking, and job queue health snapshots.

### DIFF
--- a/crates/dto/src/performance.rs
+++ b/crates/dto/src/performance.rs
@@ -236,6 +236,92 @@ pub struct AttestationSample {
     pub duration_ms: i64,
 }
 
+/// Response of `GET /performance/reliability`.
+#[derive(Debug, Clone, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct ReliabilityReport {
+    /// Error/warning log volume over time, oldest bucket first.
+    pub log_buckets: Vec<LogVolumeBucket>,
+    /// Error/warning counts per log target, highest count first.
+    pub log_targets: Vec<LogTargetCount>,
+    /// Money-at-risk lifecycle failure events in range, by event type.
+    pub failure_events: Vec<FailureEventCount>,
+    /// Current job queue health per job type. This is an instantaneous
+    /// snapshot of the queue, NOT windowed by the report range: counts span
+    /// all rows currently in the `Jobs` table regardless of `from`/`to`, so a
+    /// short range shows all-time `done`/`failed` next to windowed failures
+    /// and logs. Queue types with no rows are absent rather than zero.
+    pub job_queues: Vec<JobQueueHealth>,
+    /// `true` when the error/warning log scan hit its entry cap, so
+    /// `log_buckets` and `log_targets` undercount the noisiest windows. Lets
+    /// the dashboard render an honest partial-data state. Failure events are
+    /// aggregated in SQL and are never capped.
+    pub log_entries_truncated: bool,
+}
+
+/// Error and warning counts within one time bucket.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct LogVolumeBucket {
+    pub start: DateTime<Utc>,
+    #[ts(type = "number")]
+    pub errors: usize,
+    #[ts(type = "number")]
+    pub warnings: usize,
+}
+
+/// How often one log target emitted at one level.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct LogTargetCount {
+    pub target: String,
+    pub level: String,
+    #[ts(type = "number")]
+    pub count: usize,
+}
+
+/// Occurrences of one lifecycle failure event type.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct FailureEventCount {
+    pub event_type: String,
+    #[ts(type = "number")]
+    pub count: usize,
+    pub last_at: DateTime<Utc>,
+}
+
+/// Snapshot of one apalis job queue's health.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct JobQueueHealth {
+    pub job_type: String,
+    #[ts(type = "number")]
+    pub pending: usize,
+    #[ts(type = "number")]
+    pub running: usize,
+    #[ts(type = "number")]
+    pub done: usize,
+    /// Terminal failures from retry exhaustion. apalis writes status `Killed`
+    /// (not `Failed`) once a job exhausts `max_attempts`, carrying
+    /// `attempts >= max_attempts`, so this counts `Killed AND attempts >=
+    /// max_attempts`. Disjoint from `awaiting_retry` and `killed`.
+    #[ts(type = "number")]
+    pub failed: usize,
+    /// Failed jobs apalis will retry (status `Failed`, attempts <
+    /// max_attempts): live backlog, also included in `oldest_pending_run_at`.
+    #[ts(type = "number")]
+    pub awaiting_retry: usize,
+    /// Jobs killed before exhausting retries (status `Killed`, attempts <
+    /// max_attempts), e.g. an explicit non-retryable abort. Disjoint from the
+    /// exhaustion-driven `failed` count.
+    #[ts(type = "number")]
+    pub killed: usize,
+    /// Jobs that needed more than one attempt.
+    #[ts(type = "number")]
+    pub retried: usize,
+    pub oldest_pending_run_at: Option<DateTime<Utc>>,
+}
+
 #[cfg(test)]
 mod tests {
     use serde_json::json;
--- a/src/api.rs
+++ b/src/api.rs
@@ -16,13 +16,16 @@ use sqlx::SqlitePool;
 use tokio::sync::Mutex;
 use tracing::{error, info, warn};
 
-use st0x_dto::{HedgeLatencies, RebalanceTimings, TradingVenue};
+use st0x_dto::{HedgeLatencies, RebalanceTimings, ReliabilityReport, TradingVenue};
 use st0x_execution::FractionalShares;
 
 use crate::AppState;
 use crate::dashboard::transfer_loader::{InvalidTransferKind, TransferKind};
 use crate::equity_redemption::{EquityRedemptionEvent, RedemptionAggregateId};
 use crate::performance::rebalance::load_rebalance_timings;
+use crate::performance::reliability::{
+    aggregate_log_entries, load_failure_events, load_job_queue_health,
+};
 use crate::performance::{ReportRange, hedge_latency_report, load_hedge_performance};
 use crate::rebalancing::RebalancingService;
 use crate::rebalancing::equity::{CrossVenueEquityTransfer, RecheckError, RecheckOutcome};
@@ -43,6 +46,9 @@ fn parse_transfer_kind_filter(value: &str) -> Result<Vec<TransferKind>, InvalidT
 static STARTED_AT: LazyLock<DateTime<Utc>> = LazyLock::new(Utc::now);
 const DEFAULT_RAINDEX_ORDERS_PAGE_SIZE: u32 = 50;
 const MAX_RAINDEX_ORDERS_PAGE_SIZE: u32 = 100;
+
+/// Upper bound on ERROR/WARN log entries aggregated per reliability report.
+const MAX_RELIABILITY_LOG_ENTRIES: usize = 50_000;
 
 const GIT_COMMIT: &str = match option_env!("ST0X_GIT_COMMIT") {
     Some(val) => val,
@@ -1534,11 +1540,74 @@ async fn performance_rebalances(
     Ok(Json(timings))
 }
 
+async fn performance_reliability(
+    State(state): State<AppState>,
+    Query(query): Query<PerformanceRangeQuery>,
+) -> Result<Json<ReliabilityReport>, StatusCode> {
+    let to = query.to.unwrap_or_else(Utc::now);
+    let from = query.from.unwrap_or(to - chrono::Duration::days(7));
+    if from >= to {
+        return Err(StatusCode::BAD_REQUEST);
+    }
+    let range = ReportRange { from, to };
+
+    let (entries, log_entries_truncated) = if let Some(log_dir) = state.ctx.log_dir.as_deref() {
+        let log_dir = log_dir.to_owned();
+        let filter = LogFilter {
+            search_lower: None,
+            levels: Some(vec!["ERROR".to_string(), "WARN".to_string()]),
+            targets: None,
+            since: Some(from),
+            until: Some(to),
+        };
+        // Log scanning is synchronous file I/O; keep it off the async
+        // workers. The cap bounds memory during error storms, slightly
+        // undercounting the noisiest windows rather than exhausting the
+        // dashboard process.
+        let (entries, total, _) = tokio::task::spawn_blocking(move || {
+            read_matching_entries(&log_dir, &filter, 0, MAX_RELIABILITY_LOG_ENTRIES)
+        })
+        .await
+        .inspect_err(|error| error!(%error, "Log aggregation task failed"))
+        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+        let truncated = total > MAX_RELIABILITY_LOG_ENTRIES;
+        if truncated {
+            warn!(
+                total,
+                cap = MAX_RELIABILITY_LOG_ENTRIES,
+                "Reliability log aggregation truncated by entry cap"
+            );
+        }
+        (entries, truncated)
+    } else {
+        (Vec::new(), false)
+    };
+    let (log_buckets, log_targets) = aggregate_log_entries(&entries, &range);
+
+    let failure_events = load_failure_events(&state.pool, &range)
+        .await
+        .inspect_err(|error| error!(%error, "Failed to load failure events"))
+        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+    let job_queues = load_job_queue_health(&state.pool)
+        .await
+        .inspect_err(|error| error!(%error, "Failed to load job queue health"))
+        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+
+    Ok(Json(ReliabilityReport {
+        log_buckets,
+        log_targets,
+        failure_events,
+        job_queues,
+        log_entries_truncated,
+    }))
+}
+
 pub(crate) fn routes() -> Router<AppState> {
     Router::new()
         .route("/health", get(health))
         .route("/performance/latencies", get(performance_latencies))
         .route("/performance/rebalances", get(performance_rebalances))
+        .route("/performance/reliability", get(performance_reliability))
         .route("/logs", get(logs))
         .route("/orders/pending", get(pending_orders))
         .route("/trades", get(trades))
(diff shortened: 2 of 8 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/823?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#823** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/823?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/823?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=823).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

JuaniRios:
@coderabbit review

src/performance/reliability.rs
```diff
@@ -0,0 +1,808 @@
...
+) -> Result<Vec<JobQueueHealth>, PerformanceError> {
+    // apalis leaves retry-eligible jobs in status 'Failed' with
+    // attempts < max_attempts and re-fetches them; they are live backlog,
+    // not terminal failures (those become 'Killed').
+    let rows: Vec<JobQueueRow> = sqlx::query_as(
+        "SELECT job_type, \
+            SUM(CASE WHEN status IN ('Pending', 'Queued') THEN 1 ELSE 0 END) AS pending, \
+            SUM(CASE WHEN status = 'Running' THEN 1 ELSE 0 END) AS running, \
+            SUM(CASE WHEN status = 'Done' THEN 1 ELSE 0 END) AS done, \
+            SUM(CASE WHEN status = 'Failed' AND attempts >= max_attempts \
```
0xgleb: critical: this `failed` bucket is structurally always zero in production. apalis-sqlite's calculate_status writes status 'Killed' (not 'Failed') when a job exhausts max_attempts, so `status = 'Failed' AND attempts >= max_attempts` is a state apalis never persists. The most operationally important counter on this panel reads 0 while the real terminal failures hide under `killed`, and the comment just above this SQL already says terminal failures become 'Killed', which contradicts this line. The unit test seeds a hand-built ('Failed', 25, 25) row apalis could never write, so it passes while masking this. Define `failed` as `status = 'Killed'`, and split killed into aborted vs exhausted if you need that distinction.

src/performance/reliability.rs
```diff
@@ -0,0 +1,808 @@
...
+/// capped at [`MAX_RELIABILITY_FAILURE_EVENTS`] newest rows.
+pub(crate) async fn load_failure_events(
+    pool: &SqlitePool,
+    range: &ReportRange,
+) -> Result<Vec<FailureEventCount>, PerformanceError> {
+    let rows: Vec<(String, String)> = sqlx::query_as(
+        "SELECT event_type, occurred_at \
+         FROM lifecycle_failure_event \
+         WHERE occurred_at >= ? AND occurred_at <= ? \
+         ORDER BY occurred_at DESC \
```
0xgleb: nit: load_failure_events takes the newest 10k rows then folds per-type counts in Rust, so during a failure storm with more than 10k rows in range the per-type counts silently undercount and whole older categories can vanish, with no truncated flag in the DTO and no warn (the 50k log path in this same PR does warn, so the omission is inconsistent). This is the money-at-risk view and the cap bites exactly during the incident it exists for. Aggregate in SQL instead: SELECT event_type, COUNT(*), MAX(occurred_at) ... GROUP BY event_type removes the cap entirely (one row per type, index-backed). There is also no test seeding more than 10k rows to catch this.

src/performance/reliability.rs
```diff
@@ -0,0 +1,808 @@
...
+    /// Record a money-at-risk failure event. Non-failure events map to `None`
+    /// and are ignored.
+    async fn record(
+        &self,
+        failure: Option<(&'static str, DateTime<Utc>)>,
+    ) -> Result<(), FailureProjectionError> {
+        let Some((event_type, occurred_at)) = failure else {
+            return Ok(());
+        };
+        sqlx::query("INSERT INTO lifecycle_failure_event (event_type, occurred_at) VALUES (?, ?)")
```
0xgleb, line 114: should fix: record returns FailureProjectionError from this INSERT, but the event-sorcery ReactorBridge logs the reactor Err and continues, so a transient DB failure (contention, disk full) silently drops a money-at-risk failure from the read model while the event stays in the log. The docstring's best-effort note covers the crash window, not in-process write failures during normal operation. Short term, retry the INSERT with loud metrics; medium term a truncate-and-replay catch-up at startup would let a dropped failure recover. The swallow path is also untested.

src/performance/reliability.rs
```diff
@@ -0,0 +1,808 @@
...
+
+        // (id, job_type, status, attempts, max_attempts, run_at)
+        let jobs = [
+            ("a-1", "queue::A", "Pending", 0, 25, 1_750_000_100_i64),
+            ("a-2", "queue::A", "Queued", 0, 25, 1_750_000_080),
+            ("a-3", "queue::A", "Running", 1, 25, 1_750_000_050),
+            // Retry-eligible failure: live backlog, not terminal.
+            ("a-4", "queue::A", "Failed", 3, 25, 1_750_000_000),
+            // Exhausted failure: terminal, excluded from backlog metrics.
+            ("a-5", "queue::A", "Failed", 25, 25, 1_749_999_000),
```
0xgleb: should fix: this fixture seeds ('Failed', 25, 25) to exercise the `failed` bucket, but apalis never writes a 'Failed' row with attempts >= max_attempts (it writes 'Killed'), so the test asserts a state the real state machine cannot produce, and that is why the failed-bucket bug above ships green. Drive a real job to exhaustion (a worker on a task that errors until max_attempts) and assert which status it actually lands in. That pins the true apalis ack outcome instead of a hand-built row.

src/performance/reliability.rs
```diff
@@ -0,0 +1,808 @@
...
+    usize::try_from(value).unwrap_or_else(|_| {
+        warn!(
+            value,
+            "Negative count from job queue aggregation; reporting zero"
+        );
+        0
+    })
+}
+
+#[cfg(test)]
```
0xgleb, line 398: should fix: the riskiest incident-time paths are untested. There is no test for the 10k failure-cap undercount (seed more than 10k rows), the reactor INSERT-failure swallow, the 50k log truncation warn branch, or boundary timestamps at exactly range.from and range.to. Also performance_reliability_rejects_inverted_range covers from > to but not from == to, while the latencies endpoint tests the equal case. AGENTS.md requires coverage for new logic, and these are exactly the behaviors this feature targets.

src/performance/reliability.rs
```diff
@@ -0,0 +1,808 @@
...
+                .oldest_pending_run_at
+                .and_then(|seconds| DateTime::from_timestamp(seconds, 0)),
+        })
+        .collect())
+}
+
+/// SQLite SUM/COUNT results are non-negative; a negative value would mean a
+/// corrupted query, so it is logged loudly before falling back to zero.
+fn count(value: i64) -> usize {
+    usize::try_from(value).unwrap_or_else(|_| {
```
0xgleb: minor: count() warns and returns 0 on a negative i64::try_from. The SUM(CASE ...) shape only yields 0 or 1 per row so it cannot actually go negative, but if it ever did, masking a corrupted aggregate to a healthy-looking zero is the opposite of the repo's fail-fast posture on operational metrics. Returning a PerformanceError (the endpoint already surfaces it as a 500) would fail loudly instead.

src/performance/reliability.rs
```diff
@@ -0,0 +1,808 @@
...
+            SUM(CASE WHEN status = 'Failed' AND attempts >= max_attempts \
+                THEN 1 ELSE 0 END) AS failed, \
+            SUM(CASE WHEN status = 'Failed' AND attempts < max_attempts \
+                THEN 1 ELSE 0 END) AS awaiting_retry, \
+            SUM(CASE WHEN status = 'Killed' THEN 1 ELSE 0 END) AS killed, \
+            SUM(CASE WHEN attempts > 1 THEN 1 ELSE 0 END) AS retried, \
+            MIN(CASE WHEN status IN ('Pending', 'Queued') \
+                OR (status = 'Failed' AND attempts < max_attempts) \
+                THEN run_at ELSE NULL END) AS oldest_pending_run_at \
+         FROM Jobs \
```
0xgleb, line 364: minor: GROUP BY job_type returns only job types with at least one row in Jobs, so a configured worker queue that has never enqueued (or was cleaned up) is absent rather than showing zeros. An operator comparing against a mental checklist of queues cannot tell an empty queue from a missing one. Either document the snapshot semantics or union against the known job_type registry from the conductor wiring.

src/performance/reliability.rs
```diff
@@ -0,0 +1,808 @@
...
+    )
+    .bind(range.from.to_rfc3339())
+    .bind(range.to.to_rfc3339())
+    .bind(MAX_RELIABILITY_FAILURE_EVENTS)
+    .fetch_all(pool)
+    .await?;
+
+    let mut counts: BTreeMap<String, FailureEventCount> = BTreeMap::new();
+    for (event_type, occurred_at) in rows {
+        let Ok(occurred_at) = DateTime::parse_from_rfc3339(&occurred_at) else {
```
0xgleb: minor: this parse-and-skip warn for a malformed occurred_at is effectively dead code. The SQL filter occurred_at >= ? AND occurred_at <= ? compares TEXT lexicographically, and a value like 'not-a-valid-rfc3339' sorts after any real timestamp so it is excluded before reaching Rust. The test named after this path actually exercises the SQL filter, not this branch. Either drop the fallback (SQL filtering suffices) or test it with a value that sorts in range but is invalid RFC3339, like '2026-06-16T99:99:99Z'.

crates/dto/src/performance.rs
```diff
@@ -201,6 +201,78 @@ pub struct AttestationSample {
     pub duration_ms: i64,
 }
 
+/// Response of `GET /performance/reliability`.
+#[derive(Debug, Clone, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct ReliabilityReport {
```
0xgleb, line 242: should fix: ReliabilityReport exposes the four result vectors with no field signalling partial data, so both the 10k failure cap and the 50k log cap are invisible to the dashboard. The sibling HedgeLatencies DTO already carries total_cycles to expose its cap, so the pattern exists in this crate. Add failureEventsTruncated / logEntriesTruncated (or in-range totals) and wire them from load_failure_events and the total > MAX_RELIABILITY_LOG_ENTRIES branch so the UI can render an honest partial-data state.

crates/dto/src/performance.rs
```diff
@@ -201,6 +201,78 @@ pub struct AttestationSample {
...
+    pub level: String,
+    #[ts(type = "number")]
+    pub count: usize,
+}
+
+/// Occurrences of one lifecycle failure event type.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct FailureEventCount {
+    pub event_type: String,
```
0xgleb, line 287: should fix: event_type is a String but the value set is closed: exactly 13 variants emitted as &'static str literals by the four mapper functions. The sibling rebalance DTOs in this same file model their closed sets as enums (RebalanceStageName), and the author's own drift test exists precisely because these hardcoded strings can desync. A 13-variant FailureEventType enum with serde rename would pin the wire contract and let the compiler flag an untracked failure variant. Deferring to RAI-1006 is defensible, but it is the less idiomatic of two patterns already in this crate.

crates/dto/src/performance.rs
```diff
@@ -201,6 +201,78 @@ pub struct AttestationSample {
...
+    #[ts(type = "number")]
+    pub warnings: usize,
+}
+
+/// How often one log target emitted at one level.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct LogTargetCount {
+    pub target: String,
+    pub level: String,
```
0xgleb, line 278: should fix: level is an unconstrained String, but aggregate_log_entries only ever emits 'ERROR' or 'WARN' (every other level hits the _ => continue arm). A two-variant enum here, serialized as ERROR/WARN, would make the closed set visible to the type system and the TypeScript consumer without needing the config crate's LogLevel.

src/api.rs
```diff
@@ -1534,11 +1540,72 @@ async fn performance_rebalances(
...
+    } else {
+        Vec::new()
+    };
+    let (log_buckets, log_targets) = aggregate_log_entries(&entries, &range);
+
+    let failure_events = load_failure_events(&state.pool, &range)
+        .await
+        .inspect_err(|error| error!(%error, "Failed to load failure events"))
+        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+    let job_queues = load_job_queue_health(&state.pool)
```
0xgleb, line 1591: minor: load_job_queue_health takes no range and queries the whole Jobs table, while failure events and logs above are filtered by from/to. The assembled report mixes a windowed failure/log view with an all-time job-queue snapshot, and the DTO does not say so, so a user on last-24-hours sees done counts spanning all time next to windowed failures. Either filter job-queue by run_at/done_at, or document (and ideally surface) that job_queues ignores the range.

src/api.rs
```diff
@@ -2035,6 +2102,166 @@ mod tests {
...
+    async fn performance_reliability_reports_seeded_failures_and_jobs() {
+        let ctx = create_test_ctx_with_order_owner(Address::ZERO);
+        let state = empty_app_state(ctx).await;
+        let now = chrono::Utc::now();
+
+        // The failure read model now reads the reactor-maintained table, not
+        // the events table: seed one failure row exactly as the
+        // LifecycleFailureProjection reactor would have written it.
+        sqlx::query(
+            "INSERT INTO lifecycle_failure_event (event_type, occurred_at) \
```
0xgleb: minor: this seeds lifecycle_failure_event with a raw INSERT, but that row is reachable through the reactor, so per the no-direct-inserts convention it should be driven by ReactorHarness::receive::<OffchainOrder>(..., OffchainOrderEvent::Failed { ... }) like the unit tests do. A direct insert bypasses record's code path, so a schema change there leaves this end-to-end test green while the real reactor breaks. The Jobs insert in the same test is fine since there is no domain API for it.

src/performance/reliability.rs
```diff
@@ -0,0 +1,660 @@
...
+///
+/// Reads ONLY the reactor-maintained `lifecycle_failure_event` table.
+pub(crate) async fn load_failure_events(
+    pool: &SqlitePool,
+    range: &ReportRange,
+) -> Result<Vec<FailureEventCount>, PerformanceError> {
+    let rows: Vec<(String, String)> =
+        sqlx::query_as("SELECT event_type, occurred_at FROM lifecycle_failure_event")
+            .fetch_all(pool)
+            .await?;
```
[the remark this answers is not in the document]
JuaniRios: Already addressed in the current revision: `load_failure_events` filters by `range` in SQL (`WHERE occurred_at >= ? AND occurred_at <= ?`, index-backed on `occurred_at`) and caps the result with `ORDER BY occurred_at DESC LIMIT ?` bound to `MAX_RELIABILITY_FAILURE_EVENTS` (10_000). It no longer loads all rows into memory or filters in Rust after the fetch -- this review was on an earlier commit.

findolor (approved):

0xgleb (changes requested):

src/performance/reliability.rs
JuaniRios: Fixed in 30b908e3. Confirmed against apalis-sqlite `calculate_status` (1.0.0-rc.8): retry exhaustion writes `Killed` carrying `attempts >= max_attempts`, so `failed` now counts `Killed AND attempts >= max_attempts` (real terminal failures) and `killed` is split to aborts (`Killed AND attempts < max_attempts`). DTO docs updated to match.

src/performance/reliability.rs
JuaniRios: Fixed in 30b908e3. `load_failure_events` now aggregates in SQL (`COUNT(*)`, `MAX(occurred_at)`, `GROUP BY event_type`), so the 10k cap is gone and per-type counts stay exact during a storm.

src/performance/reliability.rs
JuaniRios: Fixed in 30b908e3. The drop is now logged at error level with the event's type and timestamp, and `record_propagates_insert_failure_instead_of_swallowing` covers the error path. I deliberately did **not** add INSERT retry: the table has no idempotency key, so retrying after a partially-applied write would double-count a money-at-risk failure. The docstring states this. A truncate-and-replay catch-up contradicts the documented forward-only design — happy to file a follow-up if you want it.

src/performance/reliability.rs
JuaniRios: Fixed in 30b908e3. The fixture seeds `('Killed', 25, 25)` (a real exhaustion row) plus a `('Killed', 1, 25)` abort row, asserting `failed == 1` and `killed == 1`.

src/performance/reliability.rs
JuaniRios: Fixed in 30b908e3. Added `performance_reliability_rejects_equal_range` (from == to), inclusive-boundary tests for the failure path and `aggregate_log_entries`, and the INSERT-swallow test. The 10k-cap undercount test is obsolete now that #2 removed the cap; the 50k-log true-branch needs 50k seeded entries (impractical as a unit test), so I assert `logEntriesTruncated == false` on the populated path instead.

src/performance/reliability.rs
JuaniRios: Fixed in 30b908e3. `count` returns `Result<usize, PerformanceError>` and surfaces a negative aggregate as `AggregateCount { value }` (a 500) instead of masking to zero. `count_rejects_negative_aggregate` covers it.

src/performance/reliability.rs
JuaniRios: Fixed in 30b908e3 (documented). The `job_queues` DTO doc now states it's an instantaneous snapshot and that queue types with no rows are absent rather than zero. Unioning against the conductor's registry is a larger change I left out.

src/performance/reliability.rs
JuaniRios: Fixed in 30b908e3. Rather than keep the dead skip branch, the read path fails fast on an unparseable `occurred_at` (consistent with the repo's fail-fast posture). `record` only ever writes valid RFC 3339, so this triggers only on corruption. `load_failure_events_fails_fast_on_in_range_invalid_timestamp` seeds an in-lexical-range invalid value (`...T99:99:99Z`) — the case the old test missed.

crates/dto/src/performance.rs
JuaniRios: Fixed in 30b908e3. Added `logEntriesTruncated: bool` to `ReliabilityReport`, wired from the log cap. The failure cap is gone (#2), so only the log path needs a partial-data signal.

crates/dto/src/performance.rs
JuaniRios: Implemented in the stacked follow-up RAI-1006 (reliability-dto-enums): a 13-variant `FailureEventType` enum with `#[serde(rename)]` matching the discriminators, plus a drift test. Resolving here as addressed by that branch.

crates/dto/src/performance.rs
JuaniRios: Also in RAI-1006: a `CountedLogLevel` enum (ERROR/WARN) replaces `level: String`. Resolving here as addressed by that branch.

src/api.rs
JuaniRios: Fixed in 30b908e3 (documented). The `job_queues` DTO doc now states queue health is an instantaneous, range-independent snapshot, so a short range no longer silently implies windowed `done` counts.

src/api.rs
JuaniRios: Fixed in 30b908e3. The e2e seed drives `ReactorHarness...receive::<OffchainOrder>(.., OffchainOrderEvent::Failed { .. })` instead of a raw INSERT, exercising `record()`'s real path. The `Jobs` insert stays raw (no domain API), as you noted.

0xgleb (approved):

PR #884, opened 2026-06-17
fix: break COIN equity rebalance guard/recovery deadlock

JuaniRios:
## What

- Breaks the COIN equity rebalance deadlock where a mint that succeeded but failed to wrap/deposit left the `equity_in_progress` guard set in a way that blocked its own recovery job, freezing rebalancing indefinitely.
- Replaces the boolean `equity_in_progress` set with a `GuardState` enum (`ActiveTransfer { generation }` | `HeldForRecovery`) so "held by a live transfer" and "held for recovery hand-off" are distinct, representable states.
- Makes the mint-succeeded-but-wrap-failed state self-heal: the stranded tokens get wrapped + deposited by the recovery job and the guard is released, with no manual intervention.

Closes [RAI-1070](https://linear.app/makeitrain/issue/RAI-1070/coin-equity-rebalance-deadlock-equity-in-progress-guard-kept-for).

## Why

- On prod (2026-06-16) a 15.52 COIN mint succeeded, the follow-up ERC-4626 wrap reverted (stale-RPC class), and the flow kept the guard set "for recovery" — but `UnwrappedEquityRecovery`'s precondition was that the guard is *not* held, so it rescheduled forever. The guard and its own recovery were mutually exclusive.
- This deadlock is a distinct latent bug reachable from any wrap failure after a successful mint, independent of the stale-RPC trigger fixed in the parent PR.

## How

- **Hand-off on failure (`rebalancing/equity/job.rs`):** on a transfer-job `PostReceipt` error, if the mint aggregate is in a pre-wrap state (`TokensReceived`/`WrapSubmitted`) and recovery is enabled for the symbol, transition the guard `ActiveTransfer -> HeldForRecovery` and return `Ok(())` so the periodic `UnwrappedEquityRecovery` job can take over. `mark_held_for_recovery` returns a `MarkHeldResult` enum and is generation-checked so it never clobbers a newer transfer's claim.
- **Deposit stage unchanged:** `TokensWrapped`/`VaultDepositSubmitted` are excluded from the hand-off and keep using apalis retry of the idempotent `resume_mint` — avoiding double-submitted deposits and startup double-drive.
- **Config-aware:** when `wrapped_equity_recovery` is disabled (the default in example/e2e configs) there is no recovery job to hand off to, so `PostReceipt` propagates `Err` and apalis retries as before — no silently-stranded tokens.
- **Guard lifecycle (`rebalancing/trigger/equity.rs`):** `RecoveryGuard` is an RAII claim with origin tracking. `UnwrappedEquityRecovery` claims a `HeldForRecovery` or absent (orphan) slot; `WrappedEquityRecovery` is orphan-only. The guard is `release()`d only on successful terminal recovery; on any failure/skip its `Drop` restores the prior state (`HeldForRecovery` stays blocking new transfers, preventing double-mint). A `generation` counter on `ActiveTransfer` prevents an ABA where a stale guard drop clobbers a fresh transfer's claim.
- **Timeout sweeper (`rebalancing/trigger/mod.rs`):** `expire_stuck_mints` skips `HeldForRecovery` mints (recovery owns them) under a single atomic write-lock, so a concurrent hand-off can't be raced into a spurious `FailWrapping`. `ActiveTransfer` mints still time out as before.
- **Startup (`conductor.rs`):** interrupted pre-wrap mints whose guard is `HeldForRecovery` are excluded from direct resume (recovery owns them); deposit-stage mints are resumed directly.
- The new `generation` field on the `TransferEquityToMarketMaking` job payload uses `#[serde(default)]` so already-queued jobs deserialize unchanged.

## Testing

- Unit tests: guard-state transitions, generation-mismatch no-op, `mark_held_for_recovery` outcomes, `RecoveryGuard` release-vs-restore on success/failure/orphan, config-disabled `Err` propagation, and the `expire_stuck_mints` `HeldForRecovery` skip.
- Integration tests: the production deadlock-break path (wrap-failed -> `HeldForRecovery` -> `UnwrappedEquityRecovery` re-wraps, deposits, and releases the guard rather than rescheduling forever), `WrappedEquityRecovery` rescheduling when the slot is held for recovery, and startup reconstruction across all four post-receipt states.
- Verified locally: full `st0x-hedge` test suite green and `cargo clippy -p st0x-hedge --all-targets` clean.

## Anything else

- Redemption-side timeout handling of `HeldForRecovery` is out of scope (this PR covers the equity mint/rebalance guard); flagged for a potential follow-up.
- Stacked on the RAI-1069 stale-RPC fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed race conditions between concurrent transfer and wrapped/unwrapped recovery for the same equity symbol.
  * Prevented transfers from starting when a symbol is reserved for recovery, avoiding double-mint conflicts.
  * Improved startup recovery by selectively resuming only safe interrupted mints and skipping pre-wrap states held for recovery.
  * Enhanced wrap-failed and no-balance recovery handling, including correct rescheduling behavior until balances appear and safe guard release on completion.

### DIFF
--- a/src/conductor.rs
+++ b/src/conductor.rs
@@ -15,7 +15,7 @@ use chrono::{DateTime, Utc};
 use sqlx::SqlitePool;
 use sqlx_apalis::sqlite::{SqliteAutoVacuum, SqliteConnectOptions, SqliteJournalMode};
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
 use std::time::Duration;
 use task_supervisor::SupervisorHandle;
 use tokio::sync::{Mutex, broadcast};
@@ -67,6 +67,7 @@ use crate::rebalancing::equity::{
     CrossVenueEquityTransfer, EquityTransferServices, TransferEquityToHedgingCtx,
     TransferEquityToMarketMakingCtx,
 };
+use crate::rebalancing::trigger::GuardState;
 use crate::rebalancing::usdc::{
     TransferUsdcToHedgingCtx, TransferUsdcToMarketMakingCtx, UsdcSettlementParams,
 };
@@ -1282,6 +1283,9 @@ fn spawn_rebalancing_infrastructure<Chain: Wallet + Clone>(
 
         let transfer_equity_to_market_making_ctx = Arc::new(TransferEquityToMarketMakingCtx {
             transfer: recovery_transfer.clone(),
+            equity_in_progress: rebalancing_service.equity_in_progress.clone(),
+            mint_store: mint_store.clone(),
+            equities_config: deps.ctx.assets.equities.clone(),
         });
 
         let transfer_equity_to_hedging_ctx = Arc::new(TransferEquityToHedgingCtx {
@@ -1391,6 +1395,8 @@ async fn recover_interrupted_tokenization_aggregates(
     let interrupted_mints = interrupted_mint_ids(pool).await?;
     let interrupted_redemptions = interrupted_redemption_ids(pool).await?;
 
+    let mut resume_mints = Vec::new();
+
     for mint_id in &interrupted_mints {
         let Some(mint) = mint_store.load(mint_id).await? else {
             return Err(anyhow::anyhow!(
@@ -1401,6 +1407,17 @@ async fn recover_interrupted_tokenization_aggregates(
         rebalancing_service
             .recover_mint_state(mint_id, &mint)
             .await?;
+
+        // Exclude pre-wrap mints (TokensReceived, WrapSubmitted) that are
+        // HeldForRecovery from direct resume. UnwrappedEquityRecovery owns
+        // those and will re-wrap + deposit the tokens. Calling resume_mint
+        // on the same aggregate would race with the recovery job (both would
+        // try to submit the wrap tx). TokensWrapped and VaultDepositSubmitted
+        // mints are always safe to resume (vault deposit is idempotent and
+        // the aggregate/inventory guard both converge to Done).
+        if !is_pre_wrap_held_for_recovery(&mint, &rebalancing_service.equity_in_progress) {
+            resume_mints.push(mint_id.clone());
+        }
     }
 
     for redemption_id in &interrupted_redemptions {
@@ -1416,9 +1433,34 @@ async fn recover_interrupted_tokenization_aggregates(
     }
 
     recover_stuck_redemptions(pool, inventory).await?;
-    resume_interrupted_transfers(transfer, &interrupted_mints, &interrupted_redemptions).await;
+    resume_interrupted_transfers(transfer, &resume_mints, &interrupted_redemptions).await;
 
     Ok(())
+}
+
+/// Returns `true` when `mint` is a pre-wrap post-receipt state
+/// (`TokensReceived` or `WrapSubmitted`) AND its symbol's guard is
+/// `HeldForRecovery`. These mints are excluded from direct `resume_mints`
+/// because `UnwrappedEquityRecovery` owns the slot and will re-wrap and
+/// deposit the tokens; calling `resume_mint` concurrently would race.
+fn is_pre_wrap_held_for_recovery(
+    mint: &TokenizedEquityMint,
+    equity_in_progress: &RwLock<HashMap<Symbol, GuardState>>,
+) -> bool {
+    if let TokenizedEquityMint::TokensReceived { symbol, .. }
+    | TokenizedEquityMint::WrapSubmitted { symbol, .. } = mint
+    {
+        let guard = match equity_in_progress.read() {
+            Ok(guard) => guard,
+            Err(poison) => poison.into_inner(),
+        };
+        match guard.get(symbol) {
+            Some(GuardState::HeldForRecovery) => true,
+            Some(GuardState::ActiveTransfer { .. }) | None => false,
+        }
+    } else {
+        false
+    }
 }
 
 async fn resume_interrupted_transfers(
@@ -2532,8 +2574,8 @@ mod tests {
     use rain_math_float::Float;
     use sqlx::{ConnectOptions, SqlitePool};
     use std::future::pending;
-    use std::sync::Arc;
     use std::sync::atomic::{AtomicUsize, Ordering};
+    use std::sync::{Arc, RwLock};
     use task_supervisor::SupervisorBuilder;
     use tokio::sync::broadcast;
 
@@ -2566,6 +2608,7 @@ mod tests {
         OnchainTradeBuilder, get_test_log, get_test_order, rebalancing_enabled_equities,
         setup_test_db, setup_test_pools,
     };
+    use crate::tokenized_equity_mint::{TokenizationRequestId, issuer_request_id};
     use crate::trading::onchain::inclusion::EmittedOnChain;
     use crate::trading::onchain::trade_accountant::AccountForDexTrade;
     use crate::unwrapped_equity_recovery::UnwrappedEquityRecoveryJob;
@@ -6797,5 +6840,93 @@ mod tests {
         let join_error = handle.await.unwrap_err();
 
         check_monitor_drain_result(Err(join_error)).unwrap_err();
+    }
+
+    /// `is_pre_wrap_held_for_recovery` excludes `TokensReceived` and
+    /// `WrapSubmitted` mints whose guard is `HeldForRecovery` from
+    /// `resume_mints`. All other states are included regardless.
+    #[test]
+    fn resume_exclusion_predicate_excludes_only_pre_wrap_held_for_recovery() {
+        let symbol = Symbol::new("AAPL").unwrap();
+        let now = Utc::now();
+
+        let equity_in_progress: Arc<RwLock<HashMap<Symbol, GuardState>>> =
+            Arc::new(RwLock::new(HashMap::new()));
+
+        let tokens_received = TokenizedEquityMint::TokensReceived {
+            symbol: symbol.clone(),
+            quantity: float!(5),
+            wallet: Address::ZERO,
+            issuer_request_id: issuer_request_id("exc-tokens-received"),
+            tokenization_request_id: TokenizationRequestId("TOK-TR".to_string()),
+            tx_hash: TxHash::ZERO,
+            shares_minted: U256::from(5u64),
+            fees: None,
+            requested_at: now,
+            accepted_at: now,
+            received_at: now,
+        };
+        let wrap_submitted = TokenizedEquityMint::WrapSubmitted {
+            symbol: symbol.clone(),
+            quantity: float!(5),
+            wallet: Address::ZERO,
+            issuer_request_id: issuer_request_id("exc-wrap-submitted"),
+            tokenization_request_id: TokenizationRequestId("TOK-WS".to_string()),
+            tx_hash: TxHash::ZERO,
+            shares_minted: U256::from(5u64),
+            fees: None,
+            requested_at: now,
+            accepted_at: now,
+            received_at: now,
+            wrap_tx_hash: TxHash::ZERO,
+        };
+        let tokens_wrapped = TokenizedEquityMint::TokensWrapped {
+            symbol: symbol.clone(),
+            quantity: float!(5),
+            wallet: Address::ZERO,
+            issuer_request_id: issuer_request_id("exc-tokens-wrapped"),
+            tokenization_request_id: TokenizationRequestId("TOK-TW".to_string()),
+            tx_hash: TxHash::ZERO,
+            shares_minted: U256::from(5u64),
+            requested_at: now,
+            accepted_at: now,
+            received_at: now,
+            wrap_tx_hash: TxHash::ZERO,
+            wrapped_shares: U256::from(5u64),
+            wrap_block: None,
+            wrapped_at: now,
+        };
+
+        // Without HeldForRecovery -- no mint is excluded.
+        equity_in_progress
+            .write()
+            .unwrap()
+            .insert(symbol.clone(), GuardState::ActiveTransfer { generation: 0 });
+        assert!(
+            !is_pre_wrap_held_for_recovery(&tokens_received, &equity_in_progress),
+            "TokensReceived + ActiveTransfer must NOT be excluded"
+        );
+        assert!(
+            !is_pre_wrap_held_for_recovery(&wrap_submitted, &equity_in_progress),
+            "WrapSubmitted + ActiveTransfer must NOT be excluded"
+        );
+
+        // With HeldForRecovery -- only pre-wrap states are excluded.
+        equity_in_progress
+            .write()
+            .unwrap()
+            .insert(symbol, GuardState::HeldForRecovery);
+        assert!(
+            is_pre_wrap_held_for_recovery(&tokens_received, &equity_in_progress),
+            "TokensReceived + HeldForRecovery must be excluded from resume_mints"
+        );
+        assert!(
+            is_pre_wrap_held_for_recovery(&wrap_submitted, &equity_in_progress),
+            "WrapSubmitted + HeldForRecovery must be excluded from resume_mints"
+        );
+        assert!(
+            !is_pre_wrap_held_for_recovery(&tokens_wrapped, &equity_in_progress),
+            "TokensWrapped must NEVER be excluded (deposit idempotent)"
+        );
     }
 }
--- a/src/integration_tests/rebalancing.rs
+++ b/src/integration_tests/rebalancing.rs
@@ -14,10 +14,11 @@ use httpmock::Mock;
 use httpmock::prelude::*;
 use serde_json::json;
 use sqlx::SqlitePool;
-use std::collections::HashMap;
-use std::sync::Arc;
+use std::collections::{BTreeMap, HashMap};
+use std::sync::{Arc, RwLock};
 use std::time::Duration;
 use tokio::sync::broadcast;
+use uuid::Uuid;
 
 use rain_math_float::Float;
 use st0x_config::{
@@ -32,7 +33,7 @@ use st0x_execution::{
 use st0x_finance::{Usd, Usdc};
 use st0x_float_macro::float;
 use st0x_raindex::{Raindex, RaindexVaultId};
-use st0x_wrapper::MockWrapper;
+use st0x_wrapper::{MockWrapper, Wrapper};
 
 use super::{ExpectedEvent, assert_events, fetch_events};
 use crate::bindings::{IERC20, TestERC20};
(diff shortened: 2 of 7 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/884?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#886** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/886?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#885** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/885?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#884** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/884?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/884?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#877** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/877?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=884).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor (approved):

JuaniRios:
@coderabbit review

src/rebalancing/equity/job.rs
```diff
@@ -103,14 +144,210 @@ impl Job<TransferEquityToMarketMakingCtx> for TransferEquityToMarketMaking {
...
     ) -> Result<Self::Output, Self::Error> {
-        ctx.transfer
+        let result = ctx
+            .transfer
             .resume_equity_to_market_making(&self.issuer_request_id, &self.symbol, self.quantity)
-            .await?;
+            .await;
 
-        Ok(())
+        match result {
```
0xgleb: nit: consider reducing nesting using let-else, e.g. something like

```rust
let Err(err) = result else {
    return Ok(());
};
```

src/rebalancing/trigger/mod.rs
```diff
@@ -705,6 +709,44 @@ impl RebalancingService {
...
+                );
+                continue;
+            };
+
+            // Atomically check whether recovery owns this slot. We hold the WRITE
+            // lock so `mark_held_for_recovery` (which also needs the write lock)
+            // cannot slip a symbol into `HeldForRecovery` between the check and
+            // the `cleanup_timed_out_mint` call. The guard drops at the end of
+            // this block; async cleanup runs after.
+            let held_for_recovery = match self.equity_in_progress.write() {
```
0xgleb: critical: the write lock here only guards the `held_for_recovery` read. It's a temporary in the match scrutinee, so it drops at the `;`, well before `cleanup_timed_out_mint().await` and the unconditional `clear_equity_in_progress()` below. A concurrent `mark_held_for_recovery` on the transfer worker can flip `ActiveTransfer` to `HeldForRecovery` in that gap, then `clear_equity_in_progress` removes the guard and `FailWrapping` is sent while the tokens are still unwrapped in the wallet. That defeats the guard and re-opens the window for a fresh transfer to start on top of the stranded tokens, which is the double-mint this guard exists to prevent. The comment claiming the lock spans the cleanup is the opposite of what the code does. Either hold the lock across the decision and the clear, or re-check the guard under the same lock inside `cleanup_timed_out_mint` and short-circuit on `HeldForRecovery`.

src/rebalancing/equity/job.rs
```diff
@@ -103,14 +144,210 @@ impl Job<TransferEquityToMarketMakingCtx> for TransferEquityToMarketMaking {
...
+                            "PostReceipt handler: failed to load mint aggregate state; \
+                             treating as non-recoverable and propagating error for apalis retry"
+                        );
+                    });
+
+                let is_post_receipt_recoverable = matches!(
+                    state,
+                    Ok(Some(
+                        TokenizedEquityMint::TokensReceived { .. }
+                            | TokenizedEquityMint::WrapSubmitted { .. } // TokensWrapped and VaultDepositSubmitted intentionally excluded:
```
0xgleb, line 192: should fix: routing `WrapSubmitted` into the HeldForRecovery handoff assumes the tokens are unwrapped. That holds when the wrap reverted, but `WrapSubmitted` means the wrap tx was already submitted, so it can also fail at PostReceipt because the tx actually landed and only the confirmation read failed (a stale-RPC outcome). In that case the tokens are wrapped, `UnwrappedEquityRecovery` reads only `BaseWalletUnwrapped` so it `NoBalanceSkip`s and the guard drop restores `HeldForRecovery`, while `WrappedEquityRecovery` uses `claim_guard_for_orphan` which refuses any present guard state including `HeldForRecovery`. The slot then stays `HeldForRecovery` with no job able to drain it, blocking all future transfers for the symbol. Worth confirming whether that landed-but-unconfirmed path is actually reachable for `WrapSubmitted`, and if so, letting the wrapped path claim a `HeldForRecovery` slot when inventory shows a wrapped balance, or having the unwrapped path hand the slot back when it sees one.

src/unwrapped_equity_recovery/job.rs
```diff
@@ -219,8 +236,29 @@ impl Job<UnwrappedEquityRecoveryCtx> for UnwrappedEquityRecoveryJob {
...
+        // - Err(NoBalanceSkip): log and map to Ok(()); guard drops normally
+        //   (HeldForRecovery origin restores to HeldForRecovery; orphan removes).
+        //   Apalis does NOT retry; next inventory poll re-triggers recovery.
+        // - Err(other): propagate; guard drops, restoring prior state.
+        match result {
+            Ok(()) => {
+                guard.release();
+                Ok(())
+            }
+            Err(UnwrappedEquityRecoveryJobError::NoBalanceSkip { symbol: ref sym }) => {
```
0xgleb, line 252: should fix: when this is reached from a `HeldForRecovery`-origin claim, the guard drop restores `HeldForRecovery` but nothing re-enqueues recovery, because the reactor's `BaseWalletUnwrappedEquity` arm only dispatches this job for symbols with a positive balance (it `continue`s on zero). If the tokens have genuinely left the wallet (recovered out of band) so the balance is and stays zero, the slot is wedged in `HeldForRecovery` and both `try_claim_for_transfer` and `claim_guard_for_orphan` refuse it, blocking the symbol with no self-heal. The comment a few lines up saying the next inventory poll re-triggers recovery doesn't hold for that zero-balance case. Either release the guard when there's genuinely no balance, or re-enqueue with a delay so it keeps polling.

src/rebalancing/equity/job.rs
```diff
@@ -103,14 +144,210 @@ impl Job<TransferEquityToMarketMakingCtx> for TransferEquityToMarketMaking {
...
+                                symbol = %self.symbol,
+                                issuer_request_id = %self.issuer_request_id,
+                                "PostReceipt handler: guard entry absent (cleared by timeout \
+                                 sweeper?); propagating Err so apalis retries"
+                            );
+                            Err(TransferEquityToMarketMakingJobError::Transfer(
+                                MintTransferError::PostReceipt(mint_error),
+                            ))
+                        }
+                        MarkHeldResult::GenerationMismatch => {
```
0xgleb, line 238: should fix: this arm has no test driving it through `perform`. It's new code that returns `Ok(())` so apalis won't retry and intentionally skips the recovery handoff, yet every other `MarkHeldResult` arm has a `perform` test and the only generation-aware test covers `RecoveryGuard::drop`, not this path. Add a `perform` test that seeds the guard as `ActiveTransfer { generation: 0 }`, builds the job with `generation: 0`, swaps the guard to `ActiveTransfer { generation: 1 }`, seeds the aggregate in `TokensReceived` with recovery enabled, and asserts `Ok(())` plus that the `generation: 1` slot is left untouched.

src/integration_tests/rebalancing.rs
```diff
@@ -2552,3 +2616,247 @@ async fn transfer_failed_cancels_redemption_inflight() {
...
+        },
+    ));
+
+    // Inventory reports UNWRAPPED balance: wrap failed, tokens are raw tSTOCK
+    // in the base wallet.
+    let shares = FractionalShares::new(float!(5));
+    let mut balances = BTreeMap::new();
+    balances.insert(symbol.clone(), shares);
+    let now = Utc::now();
+    let view = InventoryView::default().set_inflight_equity_at_location(
```
0xgleb, line 2943: should fix: this test sets only `BaseWalletUnwrapped` and never calls `set_active_mint`, so `decide_dispatch` returns `Orphan` and the test drives the standalone orphan wrap+deposit chain. In prod a wrap-failed mint is `TokensReceived`/`WrapSubmitted`, and `update_active_mint` populates `active_mint` on every non-terminal mint event, so `decide_dispatch` returns `ActiveMint` and recovery runs through `DispatchToMint` -> `resume_mint`, re-driving the mint from its persisted pre-wrap state. That ActiveMint path is the one prod takes for this scenario, and it's untested here even though the test is labelled the primary prod scenario. Add `set_active_mint(symbol, mint_id)` and seed the aggregate so the test reaches `DispatchedToMint`, and keep the orphan assertion as a separate test.

src/rebalancing/trigger/mod.rs
```diff
@@ -16920,4 +17465,81 @@ mod tests {
             "replayed Initiated must not double-count inflight"
         );
     }
+
+    /// `expire_stuck_mints` must skip mints whose symbol's in-progress guard is
+    /// `HeldForRecovery`. The recovery job owns those mints and is responsible for
+    /// driving them to a terminal state; timing them out would clear the guard and
+    /// allow a double-mint while the original tokens are still in the wallet.
+    #[tokio::test]
+    async fn expire_stuck_mints_skips_held_for_recovery_mints() {
```
0xgleb, line 17513: should fix: this only covers the steady-state case where the guard is already `HeldForRecovery` before the sweep. It doesn't cover the race where the slot is `ActiveTransfer` at the check and a concurrent `mark_held_for_recovery` flips it before `clear_equity_in_progress` runs. A test that interleaves those two operations would fail against the current dropped-lock code and pass once the lock fix lands, so it's worth adding alongside the fix.

src/rebalancing/equity/job.rs
```diff
@@ -250,43 +572,586 @@ mod tests {
...
+        // until the aggregate reaches a terminal state.
         assert!(matches!(
             error,
             TransferEquityToMarketMakingJobError::Transfer(MintTransferError::PreReceipt(_))
         ));
     }
 
+    /// PostReceipt with the aggregate ABSENT from the store (never persisted,
+    /// or already cleaned up) must propagate as Err so apalis retries.
+    /// Note: `TokensReceived`, `WrapSubmitted`, `TokensWrapped`, and
```
0xgleb: nit: this note says `TokensReceived`, `WrapSubmitted`, `TokensWrapped`, and `VaultDepositSubmitted` all transition the guard to `HeldForRecovery`, but only the first two do. `TokensWrapped` and `VaultDepositSubmitted` propagate `Err` for apalis retry because the deposit is idempotent, which the rest of the file documents correctly. Worth fixing so the next reader doesn't get the wrong model of which states hand off.

src/conductor.rs
```diff
@@ -1416,11 +1432,38 @@ async fn recover_interrupted_tokenization_aggregates(
...
 
     Ok(())
 }
 
+/// Returns `true` when `mint` is a pre-wrap post-receipt state
+/// (`TokensReceived` or `WrapSubmitted`) AND its symbol's guard is
+/// `HeldForRecovery`. These mints are excluded from direct `resume_mints`
+/// because `UnwrappedEquityRecovery` owns the slot and will re-wrap and
+/// deposit the tokens; calling `resume_mint` concurrently would race.
+fn is_pre_wrap_held_for_recovery(
```
0xgleb, line 1446: nit: this spells out `std::sync::RwLock`, `std::collections::HashMap`, and `crate::rebalancing::trigger::GuardState` inline, but `HashMap` and `Symbol` are already imported at the top of the file and used unqualified elsewhere. Pulling `RwLock` and `GuardState` into the module imports and using the short forms in the signature and match arms would match the surrounding style.

0xgleb (changes requested):

src/conductor.rs
JuaniRios: Done in 75d60d6d -- pulled `RwLock` into the `std::sync` import and `GuardState` into the module imports; the signature and match arms now use the short forms.

src/rebalancing/equity/job.rs
JuaniRios: Fixed in 75d60d6d -- corrected the doc to list only `TokensReceived` and `WrapSubmitted` as the states that hand off to `HeldForRecovery`; `TokensWrapped`/`VaultDepositSubmitted` propagate Err for apalis retry (idempotent deposit).

src/rebalancing/trigger/mod.rs
JuaniRios: Added in 75d60d6d -- `clear_equity_in_progress_unless_held_for_recovery_preserves_recovery_ownership` exercises the atomic check-and-clear directly: `HeldForRecovery` is left intact (returns false -> caller skips the failure event), `ActiveTransfer`/absent are cleared. The fix collapses the check and clear into one lock acquisition, so there's no longer a deterministic gap to interleave; this asserts the invariant the race hinged on, alongside the existing steady-state `expire_stuck_mints_skips_held_for_recovery_mints`.

src/rebalancing/equity/job.rs
JuaniRios: Good catch -- it's reachable. Fixed in 75d60d6d by letting `WrappedEquityRecovery` claim a `HeldForRecovery` slot (switched it to `claim_guard_for_recovery_or_orphan`; the atomic claim makes a concurrent `UnwrappedEquityRecovery` claim safe -- one wins, the other reschedules). When the wrap landed (wrapped balance present), the reactor dispatches `WrappedEquityRecovery`, which now drains the slot instead of rescheduling forever. Removed the now-dead `claim_guard_for_orphan`. New integration test `recovery_job_breaks_deadlock_when_wrap_landed_wrapped_equity_recovery` covers it.

src/rebalancing/trigger/mod.rs
JuaniRios: Fixed in 75d60d6d. The check and the clear now happen under a single write-lock acquisition (`clear_equity_in_progress_unless_held_for_recovery`): if the slot is `HeldForRecovery` it's left untouched and the whole mint is skipped (no cleanup, no `FailWrapping`); otherwise the guard is removed and the failure proceeds. The misleading "lock spans the cleanup" comment is gone. A concurrent `mark_held_for_recovery` flip can no longer slip between a separate check and clear.

src/rebalancing/equity/job.rs
JuaniRios: Done in 75d60d6d -- flattened to a let-else guard (`let Err(transfer_error) = result else { return Ok(()); };`), then match on the error. Kept the remaining arm exhaustive (`other @ MintTransferError::PreReceipt(_)`) so the compiler still flags new variants.

src/unwrapped_equity_recovery/job.rs
JuaniRios: Fixed in 75d60d6d. On a `NoBalanceSkip` from a `HeldForRecovery`-origin claim, the job now re-enqueues itself with a delay (`push_with_delay`) to keep polling -- the reactor only re-dispatches on a positive balance, so dropping would wedge the slot. Orphan-origin skips still just drop (the reactor re-triggers on the next positive-balance poll). Applied to both recovery jobs, since the wrapped path can now claim `HeldForRecovery` too. Covered by `perform_no_balance_skip_restores_held_for_recovery_and_reschedules`.

src/rebalancing/equity/job.rs
JuaniRios: Added in 75d60d6d -- `perform_post_receipt_generation_mismatch_leaves_new_claim_untouched`: seeds `ActiveTransfer { generation: 0 }`, builds the job at gen 0, swaps the slot to `ActiveTransfer { generation: 1 }`, seeds the aggregate in `TokensReceived`, and asserts `Ok(())` plus that the gen-1 slot is left untouched.

src/integration_tests/rebalancing.rs
JuaniRios: Added in 75d60d6d. New `recovery_job_breaks_deadlock_when_wrap_failed_dispatches_active_mint` calls `set_active_mint` + seeds the mint in `TokensReceived`, so `decide_dispatch` returns `ActiveMint` and recovery runs through `DispatchToMint -> resume_mint`, reaching `DispatchedToMint`. Kept the orphan assertion as the separate `..._unwrapped_equity_recovery` test and relabelled its doc as the no-active-mint branch.

src/rebalancing/trigger/equity.rs
```diff
@@ -35,44 +35,88 @@ pub(crate) enum EquityTriggerError {
...
+    /// `WrappedEquityRecovery`.
+    HeldForRecovery,
+}
+
+/// Monotonic counter for `GuardState::ActiveTransfer` generation tokens.
+static GUARD_GENERATION: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
+
+pub(super) fn next_generation() -> u64 {
+    GUARD_GENERATION.fetch_add(1, std::sync::atomic::Ordering::Relaxed)
+}
```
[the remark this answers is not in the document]
JuaniRios: Fixed in b4b0e1cc -- `GUARD_GENERATION` now starts at 1 so `next_generation()` is always >= 1, reserving 0 for `#[serde(default)]` legacy payloads. A stale gen-0 payload can no longer coincide with a real `ActiveTransfer` slot (always >= 1), so its `mark_held_for_recovery`/`Drop` generation check always mismatches (a safe no-op) instead of clobbering the wrong active transfer. Added an invariant test (`next_generation_never_returns_legacy_zero_sentinel`).

One caveat worth flagging: this closes the legacy-default collision, but `GUARD_GENERATION` is a process-local static that resets to 1 on every restart, so generation values still recur across restarts -- they're only unique within a single process run. The in-memory `equity_in_progress` map also resets on restart, so within a run the ABA guard holds; a true cross-restart guarantee would need a persisted/monotonic source. That's beyond this PR's scope, just noting it.

0xgleb (approved):

PR #886, opened 2026-06-17
fix: run interrupted-tokenization resume off the startup path

JuaniRios:
## What

- Move interrupted-tokenization resume (mint/redemption detection + completion polling) off the blocking startup path so a slow/down issuer can no longer wedge the bot. Implements [RAI-1081](https://linear.app/makeitrain/issue/RAI-1081/run-interrupted-tokenization-resume-off-the-startup-path-so-a-slowdown).
- `recover_interrupted_tokenization_aggregates` now enqueues a `ResumeTokenizationAggregate` job per interrupted aggregate instead of awaiting the issuer inline; the order-fill and inventory monitors start unconditionally.

## Why

- Startup awaited `recover_interrupted_tokenization_aggregates(...).await?` in `spawn_rebalancing_infrastructure` **before** spawning the monitors. For a redemption left in `Sent`, that polled the issuer for up to the full `PollingConfig` timeout (~30 min).
- While that poll ran, the bot stopped hedging and stopped detecting on-chain fills, yet systemd still reported `active (running)`. On timeout with a non-terminal redemption, startup errored and the process exited — systemd restarted it and the 30-minute block repeated: a silent freeze/crash-loop driven purely by issuer latency.
- Found diagnosing a prod freeze (2026-06-17): a QQQM redemption stuck in `Sent` blocked startup because the issuer had registered nothing that day.

## How

- New `ResumeTokenizationAggregate` apalis job (`src/rebalancing/equity/resume_job.rs`): enum target `Mint(IssuerRequestId)` / `Redemption(RedemptionAggregateId)`, ctx holds `Arc<CrossVenueEquityTransfer>`. `perform()` dispatches to `resume_mint`/`resume_redemption`; transient errors propagate for apalis retry; already-terminal aggregates return `Ok(())`.
- `recover_interrupted_tokenization_aggregates` (`src/conductor.rs`) now `requeue_orphaned()` then `cancel_all_pending()` then pushes one fresh job per interrupted aggregate — replacing the deleted blocking `resume_interrupted_transfers`. This ordering prevents duplicate Pending jobs across crash-restarts (a former-Running orphan is reset then cancelled before fresh pushes). The fast in-memory state recovery (`recover_mint_state`, `recover_redemption_state`, `recover_stuck_redemptions`) stays inline. The pre-wrap `HeldForRecovery` exclusion is preserved.
- The resume worker is registered as a **best-effort** worker, not fail-stop: new `build_best_effort_worker!` macro + `on_terminal_failure_log_only` handler (`src/conductor/job.rs`), wired with a permissive circuit-breaker config (`failure_threshold = u32::MAX`, 10s recovery) rather than the conductor-wide fail-stop. Exhausted retries on one aggregate log at `error!` and leave the conductor running — they neither crash the process nor latch the worker idle, so other interrupted aggregates still resume. `build_supervised_worker!` and `build_best_effort_worker!` now share a private `build_worker_inner!` macro.
- `apalis_pool` is threaded into `RebalancingDeps`; the resume queue is created in `spawn_rebalancing_infrastructure` and the worker registered in `builder::spawn`.
- The `requeue_orphaned` step here is intentionally soft-fail (logs and continues) — unlike sibling startup queues that abort — because gating startup on the resume queue is exactly what this PR removes; resume is idempotent so a surviving orphan is at worst a tolerated duplicate.

## Testing

- `recover_interrupted_tokenization_aggregates` tests (in-memory SQLite, aggregates seeded via CQRS commands + `MockTokenizer`, no raw `events` inserts): `issuer_down_does_not_block_tokenization_resume` (panic-on-issuer mock — completes <1s, jobs enqueued, zero issuer calls), `..._does_not_duplicate_jobs_on_restart`, `..._does_not_duplicate_running_orphan` (crash-restart with a Running orphan), `..._excludes_held_for_recovery_pre_wrap_mint`.
- `resume_job.rs` unit tests: dispatch per target variant, terminal/failed/completed aggregates return `Ok` with zero issuer I/O, missing aggregates propagate errors, payload JSON roundtrip.
- `best_effort_circuit_does_not_latch_on_single_terminal_failure` (`src/conductor/job.rs`): runs a job through `build_best_effort_worker!` with the production circuit config; one terminal failure does not stop the worker — a second job still succeeds.
- Full local CI green: `cargo check`/`clippy --all-targets --all-features` (0 warnings), `cargo fmt --check`, nixfmt, pre-commit hooks.

## Anything else

- No SPEC.md change — the startup-sequencing invariant was already stated; this is a code correction to match it. `docs/conductor.md` startup section updated.
- Self-reviewed via a multi-model panel over four passes; the headline fix (best-effort worker must not crash *or* latch on a persistently-failing aggregate) was surfaced and fixed during review.

## Summary by CodeRabbit

* **New Features**
  * Enhanced conductor startup with improved resumption of interrupted tokenization workflows for mint and redemption operations
  * Optimized recovery handling for orphaned transaction states during startup

* **Tests**
  * Extended test coverage for tokenization recovery scenarios, including failure injection and state consistency validation

### DIFF
--- a/src/conductor.rs
+++ b/src/conductor.rs
@@ -6,6 +6,7 @@ mod exit;
 pub(crate) mod job;
 mod manifest;
 pub(crate) mod monitor;
+mod trading_queues;
 
 use alloy::primitives::Address;
 use alloy::providers::{Provider, ProviderBuilder};
@@ -48,9 +49,8 @@ use crate::inventory::{
     BroadcastingInventory, Inventory, InventoryProjection, InventorySnapshot, Venue,
 };
 use crate::offchain::order::{
-    ExecutorOrderPlacer, HandleOrderRejectionJobQueue, OffchainOrder, OffchainOrderCommand,
-    OffchainOrderId, OrderPlacer, PollOrderStatus, PollOrderStatusJobQueue,
-    ReconcileOrderFillJobQueue, recover_submitted_offchain_orders,
+    ExecutorOrderPlacer, OffchainOrder, OffchainOrderCommand, OffchainOrderId, OrderPlacer,
+    PollOrderStatus, PollOrderStatusJobQueue,
 };
 use crate::onchain::OnchainTrade;
 use crate::onchain::USDC_BASE;
@@ -62,10 +62,10 @@ use crate::onchain::backfill::BackfillJobQueue;
 use crate::onchain::trade::{RaindexTradeEvent, extract_owned_vaults, extract_vaults_from_clear};
 use crate::onchain_trade::{OnChainTrade, OnChainTradeCommand, OnChainTradeError, OnChainTradeId};
 use crate::position::{Position, PositionCommand, PositionError, TradeId};
-use crate::position_check::{CheckPositionsJobQueue, bootstrap_check_positions};
 use crate::rebalancing::equity::{
-    CrossVenueEquityTransfer, EquityTransferServices, TransferEquityToHedgingCtx,
-    TransferEquityToMarketMakingCtx,
+    CrossVenueEquityTransfer, EquityTransferServices, ResumeTokenizationAggregate,
+    ResumeTokenizationCtx, ResumeTokenizationJobQueue, ResumeTokenizationTarget,
+    TransferEquityToHedgingCtx, TransferEquityToMarketMakingCtx,
 };
 use crate::rebalancing::trigger::GuardState;
 use crate::rebalancing::usdc::{
@@ -97,6 +97,7 @@ use crate::wrapped_equity_recovery::{
 
 pub(crate) use builder::CqrsFrameworks;
 use manifest::QueryManifest;
+use trading_queues::{EquityRecoveryInputs, TradingJobQueues, setup_trading_job_queues};
 
 /// CQRS/event-store and apalis worker pools over the same SQLite database.
 #[derive(Clone)]
@@ -448,10 +449,14 @@ impl Conductor {
             transfer_usdc_to_market_making_ctx,
             transfer_equity_to_market_making_ctx,
             transfer_equity_to_hedging_ctx,
+            resume_tokenization_queue,
         } = PositionAndRebalancing::setup(
             rebalancing,
             &ctx,
-            &pool,
+            &DatabasePools {
+                cqrs: pool.clone(),
+                apalis: apalis_pool.clone(),
+            },
             inventory.clone(),
             event_sender,
             vault_registry.clone(),
@@ -500,64 +505,32 @@ impl Conductor {
             snapshot,
         };
 
-        let hedge_queue = crate::conductor::job::JobQueue::new(&apalis_pool);
-        let mut poll_status_queue = PollOrderStatusJobQueue::new(&apalis_pool);
-        let reconcile_queue = ReconcileOrderFillJobQueue::new(&apalis_pool);
-        let rejection_queue = HandleOrderRejectionJobQueue::new(&apalis_pool);
-
-        // A previous process may have died mid-job anywhere on the
-        // fill-to-hedge path, leaving apalis `Running` rows no live worker
-        // owns. Reset them before the monitor spawns, exactly like the
-        // transfer/backfill/recovery queues above. Trade accounting is the
-        // one that cannot be recovered any other way: once the backfill
-        // checkpoint advances, its job row is the only record of the fill.
-        requeue_trading_orphans(&job_queue, "trade accounting").await?;
-        requeue_trading_orphans(&hedge_queue, "hedge placement").await?;
-        requeue_trading_orphans(&poll_status_queue, "order status polling").await?;
-        requeue_trading_orphans(&reconcile_queue, "order fill reconciliation").await?;
-        requeue_trading_orphans(&rejection_queue, "order rejection handling").await?;
-
-        let wrapped_equity_recovery_queue = WrappedEquityRecoveryJobQueue::new(&apalis_pool);
-        let unwrapped_equity_recovery_queue = UnwrappedEquityRecoveryJobQueue::new(&apalis_pool);
-
-        let wrapped_equity_recovery_ctx = build_wrapped_equity_recovery_ctx(
-            wrapped_equity_recovery_store,
-            rebalancing_service.clone(),
-            mint_store.clone(),
-            redemption_store.clone(),
-            inventory.clone(),
-            wrapped_equity_recovery_queue.clone(),
-            Duration::from_secs(ctx.inventory_poll_interval),
-        );
-
-        let unwrapped_equity_recovery_ctx = build_unwrapped_equity_recovery_ctx(
-            unwrapped_equity_recovery_store,
-            rebalancing_service.clone(),
-            mint_store,
-            redemption_store,
-            inventory.clone(),
-            unwrapped_equity_recovery_queue.clone(),
-            Duration::from_secs(ctx.inventory_poll_interval),
-        );
-
-        requeue_equity_recovery_orphans(
-            wrapped_equity_recovery_ctx.as_ref(),
-            unwrapped_equity_recovery_ctx.as_ref(),
-            &wrapped_equity_recovery_queue,
-            &unwrapped_equity_recovery_queue,
-        )
-        .await?;
-
-        let check_positions_queue = CheckPositionsJobQueue::new(&apalis_pool);
-
-        recover_submitted_offchain_orders(
+        let TradingJobQueues {
+            hedge_queue,
+            poll_status_queue,
+            reconcile_queue,
+            rejection_queue,
+            wrapped_equity_recovery_queue,
+            unwrapped_equity_recovery_queue,
+            wrapped_equity_recovery_ctx,
+            unwrapped_equity_recovery_ctx,
+            check_positions_queue,
+        } = setup_trading_job_queues(
+            &apalis_pool,
+            &job_queue,
+            EquityRecoveryInputs {
+                wrapped_store: wrapped_equity_recovery_store,
+                unwrapped_store: unwrapped_equity_recovery_store,
+                rebalancing_service: rebalancing_service.clone(),
+                mint_store: mint_store.clone(),
+                redemption_store: redemption_store.clone(),
+                inventory: inventory.clone(),
+                inventory_poll_interval: Duration::from_secs(ctx.inventory_poll_interval),
+            },
             &frameworks.offchain_order_projection,
-            &mut poll_status_queue,
             executor.to_supported_executor(),
         )
         .await?;
-
-        bootstrap_check_positions(&apalis_pool, &check_positions_queue).await?;
 
         let job_cleanup = spawn_finished_job_cleanup(
             pool.clone(),
@@ -583,6 +556,25 @@ impl Conductor {
         // Clone before the builder consumes it; the recovery handle needs the
         // rebalancing service to rebuild tracking during recheck recovery.
         let recovery_rebalancing_service = rebalancing_service.clone();
+
+        // Build ResumeTokenizationCtx only when BOTH the queue and the
+        // recovery_transfer are present (rebalancing enabled). zip() makes the
+        // dual-Some requirement explicit: if either is None the ctx is None,
+        // and the compiler flags any mismatch if one arm is accidentally set
+        // without the other.
+        let resume_tokenization_ctx = resume_tokenization_queue
+            .as_ref()
+            .zip(recovery_transfer.as_ref())
+            .map(|(_, transfer)| {
+                Arc::new(ResumeTokenizationCtx {
+                    transfer: transfer.clone(),
+                })
+            });
+
+        // Provide a fallback empty queue when rebalancing is disabled, so the
+        // builder always receives a concrete queue (it is never consumed).
+        let resume_tokenization_queue = resume_tokenization_queue
+            .unwrap_or_else(|| ResumeTokenizationJobQueue::new(&apalis_pool));
 
         let mut conductor = builder::spawn()
             .context(conductor_ctx)
@@ -610,6 +602,8 @@ impl Conductor {
             .maybe_rebalancing_service(rebalancing_service)
             .seed_vault_registry_queue(seed_vault_registry_queue)
             .seed_vault_registry_ctx(seed_vault_registry_ctx)
+            .resume_tokenization_queue(resume_tokenization_queue)
+            .maybe_resume_tokenization_ctx(resume_tokenization_ctx)
             .job_cleanup(job_cleanup)
             .call();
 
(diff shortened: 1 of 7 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/886?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#886** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/886?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/886?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#885** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/885?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#884** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/884?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#877** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/877?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=886).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor (approved):

src/conductor.rs
```diff
@@ -1003,19 +1001,143 @@ struct PositionAndRebalancing {
...
+    unwrapped_equity_recovery_ctx: Option<Arc<UnwrappedEquityRecoveryCtx>>,
+    check_positions_queue: CheckPositionsJobQueue,
+}
+
+/// Creates all trading-side and equity-recovery job queues, resets any orphaned
+/// `Running` rows left by a previous crash, recovers in-flight submitted orders,
+/// and bootstraps the position-check queue. Called once during conductor startup,
+/// before the apalis monitor spawns, so every `Running` row is orphaned by
+/// definition.
+async fn setup_trading_job_queues(
```
0xgleb: nit: conductor module is already huge and this helper could be a good candidate to go into a submodule

src/conductor/builder.rs
```diff
@@ -526,6 +540,13 @@ where
         let fail_stop_for_transfer_usdc_to_market_making = fail_stop.clone();
         let fail_stop_for_transfer_equity_to_market_making = fail_stop.clone();
         let fail_stop_for_transfer_equity_to_hedging = fail_stop.clone();
+        // Best-effort workers get a permissive circuit breaker: a single
+        // terminal failure must not latch the worker idle. High threshold
+        // ensures no realistic failure count ever opens the circuit; short
+        // recovery timeout means even a transient open self-heals in seconds.
+        let best_effort_circuit = CircuitBreakerConfig::default()
+            .with_failure_threshold(u32::MAX)
```
0xgleb, line 555: should fix: the comment says this high threshold means the circuit never opens, but apalis-core 1.0.0-rc.9 ignores `failure_threshold` on the live worker path. `Service::call` hardcodes `if failure_count >= 5` (worker/ext/circuit_breaker/service.rs:251); the config-aware `record_failure()` is never called from that path. So after 5 interrupted aggregates exhaust retries the circuit opens and throttles resume, exactly when the issuer is degraded. It self-heals after the 10s recovery_timeout so the blast radius is bounded, but the rationale is false and the single-failure test passes under both thresholds so it pins nothing. Correct the comment/doc and add a >=5-failure test, or drop the circuit layer for best-effort workers.

src/rebalancing/equity/resume_job.rs
```diff
@@ -0,0 +1,437 @@
...
+            ResumeTokenizationTarget::Mint(issuer_request_id) => Label::new(format!(
+                "ResumeTokenizationAggregate:mint:{issuer_request_id}"
+            )),
+            ResumeTokenizationTarget::Redemption(aggregate_id) => Label::new(format!(
+                "ResumeTokenizationAggregate:redemption:{aggregate_id}"
+            )),
+        }
+    }
+
+    async fn perform(&self, ctx: &ResumeTokenizationCtx) -> Result<Self::Output, Self::Error> {
```
0xgleb, line 81: should fix: none of the perform() tests drive a non-terminal (interrupted) aggregate, only terminal and missing ones. If this body were stubbed to `Ok(())` every test would still pass, since the terminal tests assert call_count is unchanged and the missing-aggregate tests assert error propagation. Add a test that seeds e.g. a MintAccepted aggregate, calls perform(), and asserts the tokenizer was called and the aggregate advanced in the store (mirror for a redemption).

src/conductor.rs
```diff
@@ -1387,15 +1515,40 @@ async fn recover_interrupted_tokenization_aggregates(
...
+    if let Err(error) = resume_queue.requeue_orphaned().await {
+        warn!(
+            target: "tokenization",
+            %error,
+            "Failed to reset orphaned resume-tokenization rows; a Running orphan \
+             from a crashed process may run alongside the new Pending job for \
+             the same aggregate. Tolerated because resume is idempotent."
+        );
+    }
+    resume_queue.cancel_all_pending().await;
```
0xgleb, line 1434: minor: worth documenting one more window here. If the process dies after cancel_all_pending() but before the first push(), the interrupted aggregates have no Pending or Running row, so they are not resumed until the next restart re-derives them from the events table. It self-heals on reboot, but the existing comment only covers the duplicate-tolerated case, so a note would save a future reader the trace.

src/conductor/builder.rs
```diff
@@ -991,3 +1021,35 @@ fn register_transfer_equity_to_hedging_worker(
...
+
+/// Conditionally registers the `ResumeTokenizationAggregate` worker. The ctx is
+/// `None` when rebalancing is disabled; in that case the queue exists but no
+/// worker consumes it. When enabled, runs interrupted mint/redemption aggregates
+/// off the startup path so a slow or down issuer cannot block monitoring.
+fn register_resume_tokenization_worker(
+    monitor: Monitor,
+    resume_ctx: Option<Arc<ResumeTokenizationCtx>>,
+    resume_queue: ResumeTokenizationJobQueue,
+    circuit: CircuitBreakerConfig,
```
0xgleb: minor: this takes the same raw CircuitBreakerConfig as every fail-stop register_*_worker, so nothing stops a future caller passing the fail_stop config here (which would latch the worker idle on the first failure, the exact freeze this PR fixes) or vice versa. The guarantee currently lives in a doc comment. A `BestEffortCircuit` / `FailStopCircuit` newtype pair would turn a cross-assignment into a compile error.

src/conductor.rs
```diff
@@ -2687,6 +2825,567 @@ mod tests {
...
+        assert_eq!(
+            tokenizer.call_count(),
+            calls_before,
+            "recover_interrupted_tokenization_aggregates must not call the issuer"
+        );
+
+        // Queue must have pending jobs for both interrupted aggregates.
+        let pending_jobs = pending_job_count::<ResumeTokenizationAggregate>(&apalis_pool).await;
+        assert_eq!(
+            pending_jobs, 2,
```
0xgleb: minor: this asserts the pending job count is 2 but never checks the payloads, so it does not prove one targets the mint and the other the redemption. The mint_id/redemption_id are discarded a couple lines down. Deserialize the queued payloads and assert both ResumeTokenizationTarget::Mint(mint_id) and ::Redemption(redemption_id) are present.

src/conductor.rs
```diff
@@ -1387,15 +1515,40 @@ async fn recover_interrupted_tokenization_aggregates(
...
+    // call at startup. Propagating would re-introduce the startup-gating that
+    // moving resume off the startup path removes: the whole point of the async
+    // resume worker is that a slow or failing issuer cannot block monitoring
+    // from starting. When
+    // requeue_orphaned fails, a Running orphan is NOT promoted to Pending and
+    // therefore survives the subsequent cancel_all_pending. The fresh push below
+    // then creates a duplicate Pending row for the same aggregate, but
+    // resume_mint/resume_redemption are both idempotent, so the duplicate is
+    // tolerated and the aggregate converges to its terminal state.
+    if let Err(error) = resume_queue.requeue_orphaned().await {
```
0xgleb, line 1424: minor: when this soft-fail path is taken, a Running orphan survives cancel_all_pending and the fresh push creates a Running+Pending pair, not the plain duplicate Pending the comment implies. It is tolerated only because resume_mint/resume_redemption are idempotent, so it is worth flagging idempotency as a load-bearing invariant here: any future non-idempotent side-effect in resume would break this design.

src/rebalancing/equity/resume_job.rs
```diff
@@ -0,0 +1,437 @@
...
+
+/// Dependencies the job needs.
+pub(crate) struct ResumeTokenizationCtx {
+    pub(crate) transfer: Arc<CrossVenueEquityTransfer>,
+}
+
+/// Errors emitted by [`ResumeTokenizationAggregate::perform`].
+#[derive(Debug, Error)]
+pub(crate) enum ResumeTokenizationJobError {
+    #[error("mint resume failed: {0}")]
```
0xgleb: nit: the sibling TransferEquityToMarketMakingJobError uses `#[error(transparent)]` over its #[from] variant, but this uses `{0}` + #[from], which double-renders the inner error (inline via {0} and again via the #[source] chain a chain-aware logger walks). Match the sibling with `#[error(transparent)]`.

src/conductor.rs
```diff
@@ -1387,15 +1515,40 @@ async fn recover_interrupted_tokenization_aggregates(
...
+        warn!(
+            target: "tokenization",
+            %error,
+            "Failed to reset orphaned resume-tokenization rows; a Running orphan \
```
0xgleb, line 1428: nit: this says the orphan 'may run alongside the new Pending job', but the deterministic worker name means a restarted process keeps the orphan's heartbeat fresh, so apalis never re-enqueues it. It stays stuck, not running, and only the fresh Pending row executes. The idempotency rationale is right; just the concurrent-run framing is off.

src/conductor.rs
```diff
@@ -1003,19 +1001,143 @@ struct PositionAndRebalancing {
...
+/// Bundles the optional rebalancing-side inputs needed to build the equity
+/// recovery contexts, keeping the function argument count within the lint limit.
+struct EquityRecoveryInputs {
+    wrapped_store: Option<Arc<Store<WrappedEquityRecovery>>>,
+    unwrapped_store: Option<Arc<Store<UnwrappedEquityRecovery>>>,
+    rebalancing_service: Option<Arc<RebalancingService>>,
+    mint_store: Option<Arc<Store<TokenizedEquityMint>>>,
+    redemption_store: Option<Arc<Store<EquityRedemption>>>,
+    inventory: Arc<BroadcastingInventory>,
+    inventory_poll_interval: u64,
```
0xgleb: nit: this is raw u64 seconds, consumed only via Duration::from_secs at two call sites. Taking a Duration here would encode the unit and drop the conversions. Internal DTO so low stakes, just an ambiguity a Duration removes.

0xgleb (changes requested):

src/conductor/builder.rs
JuaniRios: Corrected the comment in 6956d75c -- verified against apalis-core 1.0.0-rc.9: `CircuitBreakerService::call` hardcodes `failure_count >= 5` (service.rs:251) and never consults the configured `failure_threshold`, so `u32::MAX` does NOT keep the circuit closed. The comment now describes the real behavior (opens after 5 failed calls; the 10s `recovery_timeout`, which IS honored, bounds the throttle).

On the >=5 test: writing it surfaced a sharper issue than throttling. With retries 1+2+4=7s < the 10s `recovery_timeout`, any job that runs while the circuit is open exhausts all 3 retries against `CircuitOpen` and goes terminal WITHOUT executing. So once 5 aggregates have failed, a 6th resume job landing in the open window is silently dropped (not resumed until the next restart) -- exactly when the issuer is degraded. That makes a "self-heals, later job completes" assertion impossible to write cleanly (the success is deterministically dropped; a 5-fail + 1-success test just times out), so I left the corrected comment and skipped the contradictory test. This seems to strengthen your "drop the circuit layer for best-effort workers" alternative -- happy to do that (it needs a no-circuit worker variant) or file a follow-up if you'd prefer.

src/rebalancing/equity/resume_job.rs
JuaniRios: Added in 6956d75c -- `perform_mint_target_resumes_interrupted_aggregate` seeds a non-terminal `MintAccepted` aggregate, calls `perform`, and asserts the tokenizer was called (Poll) and the aggregate advanced to `DepositedIntoRaindex`. Mirrored with `perform_redemption_target_resumes_interrupted_aggregate` (seeds `SendPending`, asserts the tokenizer was called and it reached `Completed`). A stubbed `Ok(())` body now fails both.

src/conductor.rs
JuaniRios: Done in 6956d75c -- `EquityRecoveryInputs.inventory_poll_interval` is now a `Duration`, converted once at construction; the two `Duration::from_secs` call-site conversions are gone.

src/conductor.rs
JuaniRios: Done in 6956d75c -- fixed the framing: the deterministic `WORKER_NAME` keeps the orphan's heartbeat fresh so apalis never re-runs it; it stays a stuck row while only the fresh Pending job executes (idempotency rationale kept).

src/rebalancing/equity/resume_job.rs
JuaniRios: Done in 6956d75c -- switched both variants to `#[error(transparent)]`, matching the sibling and avoiding the double-render.

src/conductor.rs
JuaniRios: Done in 6956d75c -- corrected to "Running+Pending pair" (not a plain duplicate Pending), and flagged idempotency of `resume_mint`/`resume_redemption` as a load-bearing invariant (any future non-idempotent side-effect would break this).

src/conductor.rs
JuaniRios: Done in 6956d75c -- moved `setup_trading_job_queues` (and its `EquityRecoveryInputs`/`TradingJobQueues` I/O structs) into a new `src/conductor/trading_queues.rs` submodule. The shared `requeue_*` helpers stay in `conductor` and the submodule reaches them via `super::` (child modules can see ancestor privates), so only the moved items needed `pub(super)`.

src/conductor.rs
JuaniRios: Done in 6956d75c -- added a note that a crash between `cancel_all_pending()` and the first `push()` leaves the interrupted aggregates with no Pending/Running row, so they aren't resumed until the next restart re-derives them from the events table (self-heals on reboot).

src/conductor/builder.rs
JuaniRios: Done in 6956d75c -- added `FailStopCircuit`/`BestEffortCircuit` newtypes wrapping `CircuitBreakerConfig`; the `register_*` functions now take the typed wrapper (destructured in the param pattern), so cross-assigning the two policies is a compile error.

src/conductor.rs
JuaniRios: Done in 6956d75c -- the test now deserializes the queued `job` payloads and asserts both `ResumeTokenizationTarget::Mint(mint_id)` and `::Redemption(redemption_id)` are present, not just `count == 2`.

0xgleb (approved):

PR #926, opened 2026-06-25
Add PnL dashboard v1

data4rtist:
## What

Adds a PnL dashboard v1 backed by a standalone backend `/pnl` report endpoint.

This PR includes:

- backend `/pnl` endpoint for computed PnL reports
- Alpaca account activity ingestion for fees, credits, margin interest, dividends, and distributions
- tracked TokenizedEquityMint and UsdcRebalance cost inclusion
- market-session filtering for the dashboard report
- aggregate gross/net PnL, cost/revenue coverage, per-symbol breakdowns, and expandable docs/debug notes
- a full-width per-asset PnL view with traded volume and clickable trade drilldown
- focused Rust and frontend coverage for the PnL report surface

## Why

PnL calculation previously lived in frontend-adjacent TypeScript and did not include tracked costs or revenue. The dashboard needs a backend-computed report that is closer to accounting-grade with the data currently persisted, while clearly surfacing the remaining persistence limits.

## How

The backend endpoint builds the report from:

- SQLite `events` Position fill events
- SQLite `position_view` for reconciliation and symbol universe
- SQLite TokenizedEquityMint and UsdcRebalance cost events
- Alpaca Broker API account activities for fees, credits, margin interest, dividends, and distributions

The PnL calculation replays fills by execution timestamp through one per-symbol FIFO inventory ledger. On-chain and off-chain fills both mutate the same signed inventory book. Opposite-side fills close oldest open lots and realize PnL. Off-chain overshoots or off-chain-origin inventory are carried forward in the same ledger so later fills can close them.

Bucket attribution is an analytics layer on top of the ledger:

- prompt off-chain closes are attributed to counter-trade PnL
- opposite on-chain closes are attributed to on-chain netting PnL
- delayed or offchain-origin closes are attributed to directional realized PnL

Costs and revenues are included bottom-up from explicit sources. Account-level costs or revenues are included at aggregate level and are not artificially allocated to individual assets when the source does not provide a symbol.

The frontend consumes the backend report endpoint, paginates complete PnL entries for the selected report, and renders the report, filters, charts, costs/revenues, per-asset PnL, trade drilldown, docs, and debug notes.

## Testing

Tested locally on the pushed rebased PR head `8bb91c1c`.

Commands run:

```text
nix develop .#ci-backend -c cargo nextest run --lib -p st0x-hedge dashboard::pnl
nix develop .#ci-dashboard -c bun run test:run src/lib/pnl/api.test.ts src/lib/decimal.test.ts
nix develop .#ci-dashboard -c bunx eslint src/lib/components/pnl-panel.svelte src/lib/pnl/api.ts src/lib/pnl/api.test.ts src/lib/decimal.ts src/lib/decimal.test.ts vite.config.ts --max-warnings 0
cd dashboard && nix develop ..#ci-dashboard -c bun run build
```

Results:

```text
focused backend PnL tests: 53 passed
focused frontend PnL/decimal tests: 15 passed
scoped frontend lint: passed
production dashboard build: passed
```

Full dashboard `check` / full dashboard `lint` still report generated DTO drift outside the PnL path. The scoped PnL checks above pass on the current PR head.

## Anything else

Current limitations are surfaced in the dashboard Docs panel:

- PnL is realized closed-lot PnL from persisted fills plus tracked explicit costs/revenues
- this is not full NAV-return accounting
- percentage return and baseline/NAV drift require persisted portfolio state, mark, and cash-flow snapshots
- USD and USDC are treated as equivalent reporting currency for this view
- account-level costs/revenues are included at aggregate level and not artificially allocated to symbols
- open inventory and unmatched offchain-origin exposure are current replay state, not historical as-of-date snapshots
- market-session classification is based on the close timestamp of realized lots

The endpoint is dashboard/reporting-only and does not change hot-path trading behavior.

## Summary by CodeRabbit

* **New Features**
  * Added a new PnL view with richer breakdowns, including realized PnL, tracked costs/revenue, per-asset details, and daily session summaries.
  * Dashboard filtering now supports market-session and counter-trading views, plus improved date and snapshot handling.
  * PnL data now loads directly from the backend, and the dashboard/docs were updated accordingly.

* **Bug Fixes**
  * Improved error handling for invalid PnL requests and non-JSON responses.
  * Reduced unnecessary health/error banners in mock mode.
  * Added safer test infrastructure to limit concurrent local test nodes.


0xgleb (approved):

src/dashboard/pnl/state.rs
```diff
@@ -0,0 +1,161 @@
...
+pub(crate) struct SampleStatsAcc {
+    pub(crate) first_at: Option<String>,
+    pub(crate) last_at: Option<String>,
+    pub(crate) onchain_fill_count: usize,
+    pub(crate) offchain_fill_count: usize,
+}
+
+pub(crate) struct SummaryAndSymbols {
+    pub(crate) summary: PnlSummary,
+    pub(crate) symbols: Vec<PnlSymbolSummary>,
```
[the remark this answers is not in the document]
data4rtist: Not applying this one as written. `num_decimal::Num` is intentionally confined to `src/dashboard/pnl/`, which is a backend reporting/replay layer that emits serialized dashboard DTOs. It does not feed values back into the core trading, inventory, or on-chain domain state, where `rain_math_float::Float` remains canonical. The reporting path needs exact decimal multiplication/aggregation for displayed volume and notional, and the boundary stays inside dashboard PnL DTO construction.

src/dashboard/pnl/replay.rs
```diff
@@ -0,0 +1,855 @@
...
+    _warnings: &mut Vec<String>,
+) -> Result<Option<Fill>, PnlError> {
+    let Some(filled) = nested_record(&row.payload, "OnChainOrderFilled") else {
+        return Err(malformed_position_payload(
+            row,
+            "missing OnChainOrderFilled",
+        ));
+    };
+
+    let amount = persisted_decimal_field(row, filled, "amount")?;
```
JuaniRios, line 86: should fix: the fill parser accepts any decimal for amount, shares_filled and price and feeds it straight into FIFO replay, so a zero or negative value from a malformed or legacy event silently corrupts realized PnL. A negative quantity actually grows a lot instead of closing it, since match_fill_against_lots assumes positive inventory, and a zero price becomes a zero-cost lot. Validate that amount, shares and price are strictly positive right after parsing and error otherwise, with tests for the zero and negative cases on both venues.

src/dashboard/pnl/costs.rs
```diff
@@ -0,0 +1,785 @@
...
+        .flatten();
+    let occurred_at = bridged.and_then(|payload| {
+        if bridge_key == "Bridged" {
+            text_field(payload, "minted_at")
+        } else {
+            text_field(payload, "recovered_at")
+        }
+    });
+
+    let (Some(fee_collected), Some(occurred_at)) = (fee_collected, occurred_at) else {
```
JuaniRios, line 522: should fix: when a CCTP bridge or mint cost event is missing fee_collected or its timestamp, this warns and returns Ok(None), so the cost is dropped from tracked costs and net PnL while the report still succeeds. fee_collected is a required field on these events, so an absent value is data corruption, not an optional case, and treating it as zero understates net PnL. Return a malformed-payload error for the missing required field instead of skipping, and cover missing fee_collected and missing timestamp with tests.

src/dashboard/pnl/costs.rs
```diff
@@ -0,0 +1,785 @@
...
+            .unwrap_or_default()
+            .cmp(&activity_timestamp(right).unwrap_or_default())
+            .then_with(|| left.id.cmp(&right.id))
+    });
+
+    sorted
+        .iter()
+        .enumerate()
+        .filter_map(|(idx, activity)| {
+            let Some(occurred_at) = activity_timestamp(activity) else {
```
JuaniRios: should fix: Alpaca activity rows with a missing timestamp, missing or invalid net_amount, or a non-USD currency get turned into a warning and skipped while net PnL still returns as valid. These are external ledger rows carrying fees, interest, dividends and credits, so dropping them understates cost and revenue with no hard signal. Either fail the report on these parse failures or surface an explicit uncovered-rows count the frontend treats as incomplete, and add tests for each skip path.

src/dashboard/pnl/costs.rs
```diff
@@ -0,0 +1,785 @@
...
+    if let Some(date) = activity.date {
+        return Some(date.format("%Y-%m-%d").to_string());
+    }
+
+    activity
+        .created_at
+        .map(|created_at| created_at.to_rfc3339())
+}
+
+fn classify_activity(
```
JuaniRios, line 622: should fix: classify_activity decides cost vs revenue for every fee, interest, dividend and credit purely from the sign of net_amount, but that convention is backed only by hand-written fixtures, not a captured Alpaca response or a cited doc. If the sign convention is backwards, net PnL flips by twice the total fees. Pin it down with a real captured activities response in a test, or cite the Alpaca account-activities spec in a comment, before merging.

src/dashboard/pnl/costs.rs
```diff
@@ -0,0 +1,785 @@
...
+    activity_type: &str,
+    signed_amount: &Num,
+) -> Option<(CostCategory, AccountingBucket, AccountingEffect)> {
+    if FEE_ACTIVITY_TYPES.contains(&activity_type) {
+        return Some((
+            CostCategory::BrokerFee,
```
JuaniRios, line 632: should fix: FEE and PTC activities are aggregated into a cost bucket for the breakdown and coverage panel, but these codes also carry rebates and pass-through credits that are revenue, not cost. Counting a credit as a cost overstates the cost breakdown even when the top-line net is fine. Branch on the sign so credit-side FEE/PTC rows land in revenue, and add a test with a positive-amount FEE row.

src/dashboard/pnl/state.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub(crate) side: LotSide,
+    pub(crate) remaining_shares: Num,
+    pub(crate) price: Num,
+    pub(crate) opened_at: String,
+    pub(crate) opened_rowid: i64,
+    pub(crate) opened_venue: Venue,
+}
+
+#[derive(Debug, Clone, Default)]
+pub(crate) struct SummaryAcc {
```
JuaniRios, line 88: minor: this accumulator models USD amounts as num_decimal::Num, which reintroduces num-decimal into the root crate against the docs/float.md guidance to use the Usd newtype. That loses the currency type safety the rest of the codebase relies on. Consider using Usd for these fields.

src/dashboard/pnl/costs.rs
```diff
@@ -0,0 +1,785 @@
...
+            } else {
+                warnings.push(format!(
+                    "Skipped unsafe tokenization cost symbol in backend PnL response: {symbol}"
+                ));
+            }
+        }
+        return Ok((None, false));
+    }
+
+    let terminal_key = match row.event_type.as_str() {
```
JuaniRios, line 440: should fix: when a mint is recovered after TokensReceived, the terminal-key match can count the tokenization fee twice, once for the original mint and once for the recovery. The stuck-mint recovery flow is operationally common, so this can silently double the tokenization cost on a real headline number. De-duplicate by mint identity so a recovered mint contributes its fee once, and add a test replaying TokensReceived followed by a recovery.

src/dashboard/pnl/sessions.rs
```diff
@@ -0,0 +1,146 @@
...
+use super::response::PnlEntry;
+
+pub(crate) fn date_key(iso: &str) -> String {
+    parse_timestamp(iso).map_or_else(
+        || iso.get(..10).unwrap_or(iso).to_owned(),
+        |parsed| parsed.with_timezone(&New_York).date_naive().to_string(),
+    )
+}
+
+pub(crate) fn market_session_for_iso(iso: &str) -> &'static str {
```
JuaniRios: minor: market_session_for_iso returns a static str for what is really a closed set of session states, so callers compare and branch on raw strings. An enum would make the states exhaustive and prevent typos at the comparison sites. Consider returning a MarketSession enum.

src/dashboard/pnl/costs.rs
```diff
@@ -0,0 +1,785 @@
...
+use super::response::{PnlCostCoverage, PnlCostSummary, PnlSummary};
+use super::state::CostEventRow;
+
+const FEE_ACTIVITY_TYPES: &[&str] = &["FEE", "PTC"];
+const INTEREST_ACTIVITY_TYPES: &[&str] = &["INT"];
+const DIVIDEND_ACTIVITY_TYPES: &[&str] = &[
+    "DIV", "DIVCGL", "DIVCGS", "DIVNRA", "DIVROC", "DIVTXEX", "CGD",
+];
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
JuaniRios, line 37: should fix: this module adds #[allow(dead_code)] and #![allow(clippy::module_name_repetitions)] without a stated justification, which the quality-control policy flags as needing explicit reasoning. If the dead code is genuinely unused it should be removed, and if the naming repetition is intentional a short comment should say why. Please drop the suppressions or justify them in a comment.

src/dashboard/pnl/state.rs
```diff
@@ -0,0 +1,173 @@
...
+            Self::Onchain => "onchain",
+            Self::Offchain => "offchain",
+            Self::Manual => "manual",
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct Fill {
+    pub(crate) rowid: i64,
```
JuaniRios, line 66: minor: the replay model carries symbol as a raw String throughout, where the codebase has a Symbol newtype for exactly this. String loses the validation and lets a symbol be confused with any other string. Consider threading Symbol through the replay state.

src/dashboard/pnl/costs.rs
```diff
@@ -0,0 +1,785 @@
...
+    }
+
+    match category {
+        CostCategory::TokenizationFee => summary.tokenization_fees_usd += amount,
+        CostCategory::CctpFee => summary.cctp_fees_usd += amount,
+        CostCategory::ConversionSlippage => summary.conversion_slippage_usd += amount,
+        CostCategory::OracleWrite => summary.oracle_write_cost_usd += amount,
+        CostCategory::BrokerFee => summary.broker_fees_usd += amount,
+        CostCategory::RegulatoryFee => summary.regulatory_fees_usd += amount,
+        CostCategory::MarginInterest => summary.margin_interest_usd += amount,
```
JuaniRios: should fix: margin interest and broker fee amounts are accumulated as absolute values, since abs_decimal strips the sign before they are summed, so a credit and a charge of equal size collapse to the wrong total and the net direction is lost. For margin interest a rebate and an interest charge should not both add as positive cost. Sum the signed amounts and present magnitude only at display time.

src/dashboard/pnl/source.rs
```diff
@@ -0,0 +1,216 @@
...
+    ];
+    let symbols = query.symbol_filter(&mut warnings)?;
+    let mut tx = pool.begin().await?;
+    let as_of_rowid = effective_as_of_rowid(&mut tx, query).await?;
+    let effective_query = PnlQuery {
+        as_of_rowid: Some(as_of_rowid),
+        ..query.clone()
+    };
+
+    let event_rows = load_position_events(&mut tx, &symbols, as_of_rowid).await?;
```
JuaniRios, line 32: minor: the asOfRowid snapshot bounds the event rows by rowid but reads position_view at its current state, so a historical report mixes as-of fills with live reconciliation data. On a read-only dashboard this yields a subtly inconsistent snapshot rather than a true point-in-time view. Consider bounding position_view by the same rowid, or documenting that reconciliation is always live.

crates/execution/src/lib.rs
```diff
@@ -24,9 +24,9 @@ pub mod mock;
 pub mod order;
 
 pub use alpaca_broker_api::{
-    AlpacaAccountId, AlpacaBrokerApi, AlpacaBrokerApiCtx, AlpacaBrokerApiError,
-    AlpacaBrokerApiMode, ConversionDirection, CryptoOrderOutcome, JournalResponse, JournalStatus,
-    TimeInForce,
+    AccountActivitiesQuery, AccountActivity, AlpacaAccountId, AlpacaBrokerApi, AlpacaBrokerApiCtx,
```
JuaniRios: minor: this re-exports the Alpaca account-activity DTOs such as AccountActivity and AccountActivitiesQuery through the execution crate root, which the crate's no-leaky-abstractions rule pushes against, since consumers should go through the Executor trait rather than raw vendor DTOs. Consider keeping these types internal and exposing a domain-shaped accessor instead.

src/dashboard/pnl/response.rs
```diff
@@ -0,0 +1,293 @@
...
+    pub(crate) offchain_fill_count: usize,
+    pub(crate) matched_lot_count: usize,
+    pub(crate) open_lot_count: usize,
+    pub(crate) unmatched_offchain_fill_count: usize,
+}
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct PnlSymbolSummary {
+    pub(crate) symbol: String,
```
JuaniRios: minor: the PnL response DTO uses String for symbol fields where the DTO convention is the Symbol newtype. This is the boundary the frontend consumes, so keeping Symbol here preserves the type contract. Consider using Symbol in the response struct.

src/api.rs
```diff
@@ -158,6 +162,69 @@ async fn health() -> Json<HealthResponse> {
     })
 }
 
+async fn pnl(
```
JuaniRios, line 169: should fix: the Alpaca account-activity fetch runs separately from the DB event snapshot it is merged into, so the two can observe different points in time and a report can include activities that postdate its fill snapshot or vice versa. That produces a report where costs and fills disagree on their as-of boundary. Fetch and bound both against a single consistent cutoff, or document the tolerated skew.

src/dashboard/pnl/builder.rs
```diff
@@ -0,0 +1,195 @@
...
+        match row.event_type.as_str() {
+            "PositionEvent::OnChainOrderFilled" => {
+                if let Some(fill) = parse_onchain_fill(&row, &mut warnings)? {
+                    apply_onchain_fill(book, &fill, &mut entries, &mut warnings);
+                }
+            }
+            "PositionEvent::OffChainOrderPlaced" => {
+                apply_offchain_placement(book, &row, &mut warnings);
+            }
+            "PositionEvent::OffChainOrderFilled" => {
```
JuaniRios, line 133: should fix: tokenization fees come from persisted TokenizedEquityMint events, and separately Alpaca FEE/PTC account activities may represent the same broker-side charge, with no de-duplication between the two sources. If both fire for the same economic fee the cost is counted twice. Please confirm whether these sources can overlap, and if so key them so each fee is counted once.

src/dashboard/pnl/replay.rs
```diff
@@ -0,0 +1,855 @@
...
+    symbol: &str,
+    book: &mut SymbolBook,
+    position_nets: &HashMap<String, Num>,
+    warnings: &mut Vec<String>,
+    position_replay_deltas: &mut Vec<PositionReplayDelta>,
+) {
+    finalize_lots(&mut book.summary, &book.long_lots);
+    finalize_lots(&mut book.summary, &book.short_lots);
+
+    let epsilon = Num::from_str(EPSILON).unwrap_or_default();
```
JuaniRios: minor: Num::from_str(EPSILON).unwrap_or_default() silently falls back to zero if the constant ever fails to parse, which the no-silent-defaults rule for financial values discourages, and a zero epsilon changes lot-closing comparison behavior. Since EPSILON is a compile-time constant, parse it with expect and a message, or use a checked constructor so a bad constant fails loudly.

dashboard/src/lib/pnl/api.test.ts
```diff
@@ -9,22 +9,213 @@ describe('buildPnlParams', () => {
...
   })
 
-  it('omits optional filters when all symbols and all days are selected', () => {
+  it('omits optional filters when all symbols and all sessions are selected', () => {
     const params = buildPnlParams({
       limit: 100,
       offset: 0,
       symbols: new Set(),
-      dayFilter: 'all'
+      marketSessionFilter: 'all',
```
JuaniRios, line 27: nit: fetchCompletePnlReport guards against the total changing between pages, but no test covers that path, so a regression there would silently return a truncated or inconsistent report. Consider adding a test where the reported total shifts mid-pagination.

src/dashboard/pnl/tests.rs
```diff
@@ -0,0 +1,1946 @@
...
+    assert_eq!(report.summary.tracked_revenue_usd, "0.12");
+    assert_eq!(report.summary.net_realized_pnl_usd, "0.12");
+    assert_eq!(report.costs.broker_fees_usd, "0.12");
+    assert_eq!(report.costs.generic_costs_usd, "0");
+}
+
+#[test]
+fn matches_legacy_frontend_sql_fixture_for_stable_report_fields() {
+    let report = report_with(
+        vec![
```
JuaniRios, line 1967: nit: this end-to-end golden test asserts around fifty fields in one block, so a failure points at the whole report rather than the specific number that drifted. Splitting the assertions by concern, such as realized PnL, costs, and per-symbol, would make failures diagnosable. Not blocking.

src/dashboard/pnl/builder.rs
```diff
@@ -0,0 +1,195 @@
...
+        merge_symbol_replay_exposure(filtered.symbols, replay_symbols.into_iter());
+    let symbols_with_costs = with_direct_symbol_costs(
+        reset_symbol_costs(symbols_with_exposure),
+        &filtered_cost_entries,
+    )?;
+
+    let mut symbol_universe: BTreeSet<String> = position_symbols.into_iter().collect();
+    symbol_universe.extend(books.keys().cloned());
+    symbol_universe.extend(symbols_with_costs.iter().map(|row| row.symbol.clone()));
+    let symbol_universe: Vec<_> = symbol_universe.into_iter().collect();
```
JuaniRios, line 239: nit: symbol_universe is collected and then cloned into the response where the original is no longer used, so it can be moved instead of cloned.

src/dashboard/pnl/replay.rs
```diff
@@ -0,0 +1,855 @@
...
+                .entry(front_lot.trade_id.clone())
+                .or_default();
+            *matched += &matched_shares;
+        }
+
+        let opening_direction = lot_side_to_direction(front_lot.side);
+        let closing_direction = fill.direction;
+        let opening_price_text = fmt_decimal(&front_lot.price);
+        let closing_price_text = fmt_decimal(&fill.price);
+        let onchain_direction = if front_lot.opened_venue == Venue::Onchain {
```
JuaniRios: nit: the same three-arm venue-select branch is repeated across five fields here. Factoring it into one helper that selects by venue would remove the repetition.

src/dashboard/pnl/parsing.rs
```diff
@@ -0,0 +1,278 @@
...
+
+pub(crate) fn optional_persisted_decimal_field(
+    row: &PositionEventRow,
+    payload: &Value,
+    key: &'static str,
+) -> Result<Option<Num>, PnlError> {
+    optional_persisted_decimal_value(row.rowid, "Position", row.event_type.clone(), payload, key)
+}
+
+pub(crate) fn persisted_decimal_value(
```
JuaniRios, line 65: nit: persisted_decimal_value and its sibling differ only in null handling, so they are near-duplicates. Consider one parser parameterized on whether null is allowed.

src/dashboard/pnl/parsing.rs
```diff
@@ -0,0 +1,278 @@
...
+            .and_then(|placed| text_field(placed, "placed_at")),
+        "PositionEvent::ManualPositionAdjusted" => {
+            nested_record(&row.payload, "ManualPositionAdjusted")
+                .and_then(|adjusted| text_field(adjusted, "adjusted_at"))
+        }
+        _ => None,
+    }
+}
+
+pub(crate) fn ordered_position_events(
```
JuaniRios, line 146: nit: ordered_position_events sorts fills by timestamp, but no test exercises two fills at the same timestamp across venues, where the tiebreak decides replay order and thus realized PnL. Add a same-timestamp cross-venue ordering test to lock the tiebreak down.

crates/execution/src/alpaca_broker_api/activity.rs
```diff
@@ -0,0 +1,415 @@
...
+
+use chrono::{DateTime, NaiveDate, Utc};
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+use super::AlpacaBrokerApiError;
+use super::auth::AlpacaBrokerApiCtx;
+use super::client::AlpacaBrokerApiClient;
+
+const PNL_ACTIVITY_TYPES: &[&str] = &[
```
JuaniRios, line 20: minor: PNL_ACTIVITY_TYPES and the status set are assumed rather than backed by the Alpaca activities spec, so a missing or renamed code would be silently ignored. Cite the spec, or add a test decoding a real activities response that covers each type.

crates/execution/src/alpaca_broker_api/activity.rs
```diff
@@ -0,0 +1,415 @@
...
+    let mut rows = Vec::new();
+    let mut page_token: Option<String> = None;
+
+    for _ in 0..MAX_ACCOUNT_ACTIVITIES_PAGES {
+        let page = fetch_account_activities_page(client, query, page_token.as_deref()).await?;
+        if page.is_empty() {
+            return Ok(rows);
+        }
+
+        let last_id = page.last().map(|row| row.id.clone());
```
JuaniRios, line 104: nit: pagination uses the last row id as the page_token cursor, which assumes Alpaca's activities endpoint pages by id in that direction. If the cursor semantics differ, paging could loop or skip rows. Worth confirming against the API docs or a captured multi-page response.

JuaniRios (changes requested):

data4rtist:
@JuaniRios I pushed `92209c7f` with the follow-up fixes for your re-review comments. I re-checked the full comment list against the final diff and believe all items are covered.

**Should-fix items addressed**
- Strictly reject non-positive onchain/offchain fill quantities and prices, with zero/negative tests for both venues.
- Treat missing TokenizedEquityMint/CCTP fee fields and timestamps as malformed payloads instead of silently dropping costs, with tests.
- Treat malformed Alpaca ledger rows as hard report failures for missing timestamp, missing/invalid `net_amount`, non-USD currency, unsupported activity type, and unsupported status.
- Document the Alpaca `net_amount` signed broker-cash convention and fail closed for unknown activity/status values rather than silently undercounting.
- Classify positive FEE/PTC rows as revenue and net broker fee / margin-interest category totals by signed amount.
- De-duplicate tokenization recovery fees by mint aggregate and de-duplicate overlapping Alpaca broker fees against persisted tokenization fees.
- Document the accepted live/snapshot caveat for `position_view` reconciliation and live Alpaca activity ingestion.
- Removed the root execution re-export of Alpaca account-activity DTOs.
- Parse `EPSILON` loudly instead of silently defaulting.

**Minor/nit items addressed**
- Kept `Num` confined to report-only PnL accumulators and documented that boundary; no values flow back into core trading/inventory/onchain state.
- Threaded `Symbol` through replay state and PnL response DTOs.
- Added `MarketSession` enum.
- Removed unjustified lint suppressions.
- Moved `symbol_universe` instead of cloning it.
- Factored the repeated venue text selection helper.
- Parameterized the decimal parser null-policy helper.
- Added same-timestamp cross-venue replay-order coverage.
- Added frontend pagination total-change coverage.
- Split the large legacy fixture assertion block by concern.

**Alpaca docs pinned in code**
- Activity type filtering / enum: https://docs.alpaca.markets/us/reference/getaccountactivitiesbytype
- Account activities and `page_token` pagination: https://docs.alpaca.markets/us/reference/getaccountactivities

**Validation rerun locally**
- `cargo fmt --check`
- `git diff --check`
- `nix develop .#ci-backend -c cargo test --lib dashboard::pnl::tests` — 75 passed
- `nix develop .#ci-backend -c cargo test -p st0x-execution alpaca_broker_api::activity --lib` — 6 passed
- `nix develop .#ci-backend -c cargo clippy -p st0x-hedge --lib --tests --all-features`
- `nix develop ..#ci-dashboard -c bun run test:run src/lib/pnl/api.test.ts` — 8 passed

Could you re-review the latest commit when you get a chance? If it looks good, I will resolve the covered threads and merge once CI is green.

JuaniRios (approved):

JuaniRios:
* **#926** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/926?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/926?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=926).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #931, opened 2026-06-25
feat: order aggregate for extended-hours counter-trading

JuaniRios:
## What

- Bring the `OffchainOrder` event-sourced aggregate to its full extended-hours shape: `Cancelling` / `Cancelled` states, `CancelOrder` / `ConfirmCancellation` / `UpdatePartialFill` commands+events, `CounterTradeOrderKind`, and the shared `terminal_position_finalization` / `position_command_for_finalization` finalization helpers.
- Two-phase cancellation is fail-closed: a pre-cancel reconcile fetches broker state first and, if that fetch fails or the order has positive unpriced fills, blocks and retries rather than cancelling blind.
- Reconcile broker outcomes in the poll loop (`poll_status`): record partial fills, recover broker-side (unrequested) cancellations, confirm requested cancellations; `handle_rejection` / `reconcile_fill` thread broker timestamps through terminal transitions.
- Confirmed cancellations finalize the owning `Position` inline in the poll path: `PollOrderStatusCtx` carries a `position_store`, and `confirm_cancellation` + the broker-side-cancel recovery dispatch the position command through the shared finalization helpers — fail-closed on an unpriced fill (leave the position pending, never clear it).
- `Position` gains the `OffChainOrderCancelled` event/command (distinct from `Failed`); `api` adds `Cancelling` to the pending-orders query.

## Why

- Second PR of the stack (#916 -> **#931** -> #932). Lands the state machine and outcome reconciliation on top of the execution layer, before the hedge / cancel-and-replace policy wiring in #932.

## How

- `src/offchain/order.rs` + `poll_status.rs` / `handle_rejection.rs` / `reconcile_fill.rs`, `src/position.rs`, `src/api.rs` at the feature version. `cli/trading` and the event projections (`performance/projection`, `performance/reliability`, `dashboard/*`, `rebalancing/trigger`) are also brought up (additive arms for the new events).
- Consumer glue, conservative and finalized in #932: `conductor` and `hedge` handle the new `Cancelling` / `Cancelled` states and pass `kind: CounterTradeOrderKind::Market` on `Place`; conductor's orphan recovery is refactored to route terminal states through the shared `position_command_for_finalization` helper so #932's cancel-and-replace sweep cannot drift from it.

## Testing

[...]

PR #916, opened 2026-06-24
feat(execution): extended-hours execution-layer primitives

JuaniRios:
## What

- Add the `st0x-execution` crate groundwork for extended-hours counter-trading: a `MarketSession` (Regular/Extended/Closed) classification from the broker calendar, `place_limit_order` + `cancel_order` (-> `CancellationOutcome`) + `fetch_latest_trade_price` on the `Executor` trait, and the `LimitOrder` / `OrderPlacement` extensions.
- Represent broker order outcomes faithfully: `OrderState::PartiallyFilled` / `Cancelled`, `OrderUpdate.shares_filled`, and `client_order_id` idempotency; parse `filled_qty` and map broker `PartiallyFilled` / `Canceled` statuses instead of collapsing them; drive status updates from broker timestamps (`filled_at` / `canceled_at`) with a terminal-timestamp fallback.
- `OrderPlacement` now records `extended_hours` and `limit_price` -- on a duplicate-`client_order_id` 422 the *adopted* order's terms are reported (for both market and limit placements) so a later convergence sweep sees broker reality, falling back to the request's `extended_hours` / `limit_price` only when the broker lookup omits them (market adoption defaults an omitted `extended_hours` to `false`).
- Adapt the main crate at the execution-API boundary so the workspace still compiles: `poll_status` dispatch, the trading CLI (`place_alpaca_limit_order` + new `OrderState` arms), the telemetry `Executor` decorator, and the executor test mocks.

## Why

- First implementation PR of the extended-hours counter-trading stack (#915 -> **#916** -> #931 -> #932). Isolating the broker-facing primitives keeps the execution layer independently reviewable before the `OffchainOrder` aggregate (#931) and the hedge/sweep wiring (#932) build on it.

## How

- All of `crates/execution/**` at the feature version. `Executor` gains `place_limit_order` / `cancel_order` as newly required methods; `market_session` / `fetch_latest_trade_price` have default impls.
- Boundary glue is deliberately conservative and behaviour-preserving (finalized in #931/#932): `poll_status` re-polls `PartiallyFilled` exactly as the pre-feature `Submitted` path did and treats a broker `Cancelled` as a rejection; the CLI renders/handles the new states and the test mocks forward the new trait methods.

## Testing

[...]

PR #915, opened 2026-06-24
feat: groundwork for extended-hours counter-trading

JuaniRios:
## What

- Spec the extended-hours counter-trading feature (`SPEC.md`) and add its inert config + DB surface so the implementation PRs stack cleanly with no behaviour change.
- Add the required per-asset `extended_hours_counter_trading: OperationMode` field to `EquityAssetConfig` (no default — must be set explicitly), plus `is_extended_hours_enabled(symbol)` / `any_extended_hours_enabled()` helpers.
- Tighten `MAX_COUNTER_TRADE_SLIPPAGE_BPS` 10_000 -> 9_999 (10_000 bps = 100% would zero sell-side limit prices).
- Add `Cancelling` / `Cancelled` CASE arms to the generated offchain-order `status` view column, ahead of the code that produces those states.

## Why

- First PR of the stack (**#915** -> #916 -> #931 -> #932). Lands the config + schema groundwork so the feature can be built and reviewed in independent slices. The only behaviour change is config validation (see below); no trading behaviour is wired.

## How

- `crates/config/src/loader.rs`: new required field + helpers (`pub`, so no dead-code before callers land) + slippage-bound tightening, with config-parse and helper unit tests.
- `migrations/20260610135812_add_cancelling_cancelled_to_offchain_order_view.sql`: drops and recreates the `offchain_order_view` projection table so its generated `status` column gains `Cancelling` / `Cancelled` CASE arms (the table rebuilds from events at startup, so DROP+CREATE is safe); must precede any code that emits those states.
- `config/{prod,staging}/st0x-hedge.toml`, `example.config.toml`, `e2e/config.toml`: field added as `"disabled"`.
- Test fixtures that build `EquityAssetConfig` (`conductor`, `rebalancing/trigger`, `integration_tests/*`, `test_utils`, `vault_registry`, e2e infra) gain the now-required field as `Disabled`.

## Testing

- `crates/config` unit tests: field is required (parse fails when absent), helper returns the configured value, slippage rejects 10_000 bps and accepts up to 9_999. Workspace compiles `--all-targets`; clippy + fmt clean.

## Anything else

- Behaviour change is config validation only: `extended_hours_counter_trading` is now a required per-asset field (parse fails if absent) and `counter_trade_slippage_bps = 10000` is rejected (was the prior upper bound). No trading code consumes the field yet; it is `"disabled"` everywhere.
- Migration drops/recreates the `offchain_order_view` projection table (rebuilt from events at startup) — no source-of-truth data change.
[...]

PR #917, opened 2026-06-24
chore: enable extended-hours counter-trading for all assets except SPCX

JuaniRios:
## What

- Flip the per-asset `extended_hours_counter_trading` flag to `enabled` for every equity except SPCX, in both the prod and staging configs.

## Why

- Activates the extended-hours counter-trading feature (implemented in #916) in the live configs. SPCX stays disabled.

## How

- Config-value-only change: 17 assets per file flipped `disabled` -> `enabled` across prod and staging; SPCX left `disabled`.

## Testing

- No code change; configs are validated by the `all_repo_config_tomls_are_valid` loader test.

## Anything else

- Top of the 3-PR stack: #915 (groundwork) -> #916 (implementation) -> #917 (this). With this applied, the stack's code tree is byte-identical to the original single-PR branch #673 (superseded).
- This is the on-switch; revertable by flipping the values back.

## Summary by CodeRabbit

* **Configuration Updates**
  * Enabled extended-hours counter-trading for several equity symbols in both staging and production environments.
  * Applied the updated trading behavior consistently across the affected stock/ETF configurations.
  * Left the SPCX equity setting unchanged (still disabled).

PR #932, opened 2026-06-25
feat: wire extended-hours counter-trading (hedge, position-check, conductor)

JuaniRios:
## What

- Wire the extended-hours policy on top of the aggregate: session-aware hedge placement (`apply_slippage` limit pricing with Alpaca penny/sub-penny rounding, Market-vs-`ExtendedHoursLimit` selection by `MarketSession` re-checked at perform time), the level-triggered cancel-and-replace sweep that converges still-live extended-hours limit orders to market orders at the regular open, and the inline extended-hours enqueue path.
- Replace the conservative placeholders from #916/#931 with the real logic: `hedge` (in `PlaceHedge::perform`) derives `CounterTradeOrderKind` from the session re-checked at perform time, while `conductor` keeps inline regular-hours placement as `Market` and, for an extended-hours fill, enqueues an immediate `PlaceHedge` job instead of placing inline. Cancellations route through the shared finalization helpers.

## Why

- Final implementation PR of the stack (#915 -> #916 -> #931 -> **#932**). Turns the state machine + execution primitives into the actual extended-hours behaviour. Ships inert (config `"disabled"` everywhere) until enabled per asset in a later config flip.

## How

- `src/trading/offchain/hedge.rs`: session recheck at perform time (run unconditionally — see the post-review fix below), `apply_slippage`, order-kind selection with a per-symbol extended-hours gate, submission lock.
- `src/position_check.rs`: `request_extended_hours_cancellations` sweep gated by `is_extended_hours_enabled`, plus `finalize_terminal_pending_positions`.
- `src/conductor.rs` + `src/conductor/builder.rs`: inline extended-hours enqueue, the hedge `OrderPlacer` wired as a session source, `hedge_queue` wiring, and orphan recovery routed through the shared `position_command_for_finalization` helper.
- `src/onchain/accumulator.rs`, `src/trading/onchain/trade_accountant.rs`, `src/conductor/monitor/executor_maintenance.rs`, `src/integration_tests/arbitrage.rs`.

## Testing

- Unit tests for session-aware order-kind selection / `apply_slippage`, the cancel-and-replace sweep, and the conductor extended-hours enqueue path (`extended_hours_trade_enqueues_immediate_hedge_job`); hedge tests cover both decoupling failure modes (a stale Regular job skips when the live venue is closed; a disabled-symbol job skips during an Extended session) and the terminal-pending recovery no-re-poll path; `integration_tests/arbitrage.rs` updated. Verified locally on the stack tip: `cargo check --all-targets`, `cargo clippy --all-features`, `cargo fmt --check` clean. CI runs the full suite per PR.

## Anything else

- Post-review blocker fix ([RAI-1173](https://linear.app/makeitrain/issue/RAI-1173)): the perform-time market-session recheck was gated by the global `order_placer` flag, which was `Some` only when some asset enabled extended hours. That let a stale Regular job submit a market order after the venue closed (when all assets were disabled) or place an orphan-able extended-hours limit for a per-symbol-disabled symbol. Fix: the `OrderPlacer` is now **always** constructed (a pure session source, decoupled from the feature flag), so the recheck always runs; per-symbol eligibility is carried in `HedgeCtx.assets` and the `Extended` arm skips symbols that are not enabled. `recover_pending_poll_status` now warns instead of silently returning `Ok` for a terminal pending order, and the unreachable `OrderPlacerNotConfigured` error was removed.
- Inert until enabled per asset: with every asset `"disabled"`, the always-present `OrderPlacer` never reaches a limit-order placement because the per-symbol `is_extended_hours_enabled` gate skips every Extended placement, and the cancel-and-replace sweep (`request_extended_hours_cancellations`) skips every symbol — so behaviour is unchanged (market orders during regular hours only).
- One unconditional behaviour change: the `CheckPositions` tick now always runs `finalize_terminal_pending_positions` (finalizes terminal pending offchain orders into position commands, with `Cancelled` orders finalized via cancel/complete rather than failure semantics) regardless of the flag — active even with the feature disabled.
- This stack applies the extended-hours feature on top of current master and deliberately preserves master's operator-recovery / equity-redemption (schema 4) work -- the original single-PR branch predated those and this stack does not revert them.
- Stack: #915 -> #916 -> #931 -> **#932**.

## Summary by CodeRabbit

* **New Features**
  * Added extended-hours session-aware trade handling, including slippage-aware counter-trading with limit orders and automatic hedge job routing.
  * Enhanced periodic position monitoring to finalize terminal pending states and sweep/cancel still-live extended-hours orders when enabled.

* **Bug Fixes**
  * Improved startup/orphan order recovery to avoid silent undecided states and to ensure cancelled orders and unpriced terminal partial fills are handled consistently.
  * Adjusted cancelled-state dispatch so it no longer surfaces as a generic failure during post-placement processing.

* **Tests**
  * Updated and expanded coverage for extended-hours hedging, recovery, cancellation routing, and retry behavior.

### DIFF
--- a/src/conductor.rs
+++ b/src/conductor.rs
@@ -42,7 +42,7 @@ use st0x_evm::{USDC_BASE, Wallet};
 use st0x_execution::{
     AlpacaBrokerApi, AlpacaWalletService, ClientOrderId, CounterTradePreflight,
     CounterTradeReservation, CounterTradeSkipReason, ExecutionError, Executor, FractionalShares,
-    MarketOrder, Symbol, TryIntoExecutor,
+    MarketOrder, MarketSession, Symbol, TryIntoExecutor,
 };
 use st0x_raindex::{RaindexService, RaindexVaultId};
 use st0x_wrapper::WrapperService;
@@ -60,8 +60,9 @@ use crate::inventory::{
 use crate::offchain::order::{
     ExecutorOrderPlacer, OffchainOrder, OffchainOrderId, OffchainOrderPlacement, OrderPlacer,
     PollOrderStatus, PollOrderStatusJobQueue, TerminalPositionFinalization,
-    client_order_id_for_placement, place_offchain_order_at_broker,
-    position_command_for_finalization, terminal_position_finalization,
+    client_order_id_for_placement, finalize_cancelled_position_or_log_unpriced,
+    place_offchain_order_at_broker, position_command_for_finalization,
+    terminal_position_finalization,
 };
 #[cfg(test)]
 use crate::offchain::order::{OffchainOrderCommand, noop_order_placer};
@@ -215,6 +216,10 @@ pub(crate) struct TradeProcessingCqrs {
     /// hook the order stays `Submitted` forever -- the system has no other
     /// trigger to ask the broker about an in-flight order.
     pub(crate) poll_status_queue: PollOrderStatusJobQueue,
+    /// Used to enqueue an immediate `PlaceHedge` for extended-hours fills, which
+    /// the inline path cannot place itself (it has no `OrderPlacer` for the
+    /// limit-order price lookup). `CheckPositions` is the backstop.
+    pub(crate) hedge_queue: crate::trading::offchain::hedge::HedgeJobQueue,
 }
 
 /// Orchestrates the bot's runtime by composing long-running supervised tasks
@@ -1808,7 +1813,6 @@ pub(crate) async fn recover_orphaned_pending_offchain_orders(
     Ok(())
 }
 
-#[allow(clippy::cognitive_complexity)]
 async fn recover_single_orphaned_order(
     position: &Arc<Store<Position>>,
     offchain_order: &Arc<Store<OffchainOrder>>,
@@ -1948,14 +1952,23 @@ async fn resolve_terminal_claimed_order(
 
 /// Maps an orphaned offchain order to the position command that finalizes
 /// its owning position, or `None` when the position must be left pending.
-/// Routes terminal states through the shared finalization helpers so the
-/// mapping cannot drift from the cancel-and-replace sweep / rejection paths.
+/// Every `None` path logs its reason, so the caller's early return is never
+/// silent.
+#[allow(clippy::cognitive_complexity)]
 fn orphaned_order_position_command(
     order: &OffchainOrder,
     symbol: &Symbol,
     order_id: OffchainOrderId,
 ) -> Option<PositionCommand> {
     match terminal_position_finalization(order) {
+        // Mirrors `CheckPositionsCtx::finalize_position_for_terminal_order`:
+        // the fill cannot be recorded without a price, so leave the position
+        // pending instead of aborting the whole startup recovery sweep. The
+        // order is terminal, so its data never changes -- NO automated path
+        // resolves this; the every-tick finalize sweep only re-logs it, and
+        // operator intervention is required to record or discard the fill.
+        // Aborting here would instead crash-loop the bot and block hedging of
+        // every other symbol.
         Some(TerminalPositionFinalization::UnpricedFill { shares_filled }) => {
             error!(
                 %symbol, %order_id, %shares_filled,
@@ -1965,15 +1978,35 @@ fn orphaned_order_position_command(
             None
         }
 
+        // Non-terminal: the order is still in progress (Submitted /
+        // PartiallyFilled / Cancelling / Pending) -- nothing to finalize yet.
         None => {
-            if let OffchainOrder::Pending { .. } = order {
-                warn!(%symbol, %order_id, "Orphaned offchain order still Pending at startup -- may never have been submitted to the broker");
-            } else {
-                debug!(%symbol, %order_id, state = ?order, "Orphaned offchain order still in progress");
+            // A `Pending` order at startup was claimed by the position but never
+            // submitted to the broker, and this recovery sweep has no path to
+            // re-drive it -- surface it at warn so the anomaly is visible. The
+            // other in-flight states (Submitted/PartiallyFilled/Cancelling) are
+            // owned by the poll loop and are expected here.
+            match order {
+                OffchainOrder::Pending { .. } => {
+                    warn!(%symbol, %order_id, "Orphaned offchain order still Pending at startup -- may never have been submitted to the broker");
+                }
+                OffchainOrder::Submitted { .. }
+                | OffchainOrder::PartiallyFilled { .. }
+                | OffchainOrder::Cancelling { .. } => {
+                    debug!(%symbol, %order_id, state = ?order, "Orphaned offchain order still in progress");
+                }
+                OffchainOrder::Filled { .. }
+                | OffchainOrder::Failed { .. }
+                | OffchainOrder::Cancelled { .. } => {
+                    debug!(%symbol, %order_id, state = ?order, "Orphaned terminal order had no position finalization");
+                }
             }
             None
         }
 
+        // Complete and both NoFill outcomes map through the shared helper so
+        // the terminal-state -> position-command mapping cannot drift from
+        // the cancel-and-replace sweep and rejection paths.
         Some(finalization) => {
             let command = position_command_for_finalization(finalization, order_id);
             info!(%symbol, %order_id, ?command, "Orphaned order terminal -- recovering");
@@ -2240,10 +2273,9 @@ pub(crate) async fn execute_acknowledge_fill(
 }
 
 /// Marks the trade acknowledged on the `OnChainTrade` aggregate after the
-/// position write succeeded, completing the exactly-once pair (ADR 0005).
-/// A re-driven marker (`AlreadyAcknowledged`) is idempotent;
-/// infrastructure errors propagate so the apalis retry re-drives the
-/// idempotent acknowledge pair.
+/// position write succeeded. A re-driven marker (`AlreadyAcknowledged`) is
+/// idempotent; infrastructure errors propagate so the apalis retry re-drives
+/// the idempotent accounting steps.
 pub(crate) async fn execute_mark_acknowledged(
     onchain_trade: &Store<OnChainTrade>,
     trade_id: &OnChainTradeId,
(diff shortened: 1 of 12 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/932?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#932** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/932?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#917](https://github.com/ST0x-Technology/st0x.liquidity/pull/917) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/917?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#941](https://github.com/ST0x-Technology/st0x.liquidity/pull/941) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/941?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/932?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#931** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/931?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#916** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/916?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#915** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/915?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=932).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

JuaniRios:
@coderabbitai full review

src/conductor.rs
```diff
@@ -2173,6 +2195,57 @@ where
...
+    // OrderPlacer for the limit-order price lookup). Instead of waiting for the
+    // next CheckPositions scan (~1 min), enqueue an immediate PlaceHedge job,
+    // which has the OrderPlacer and submits the limit order. CheckPositions
+    // remains the backstop. Preflight + clamp here, mirroring the scan path, so
+    // we never enqueue a hedge that would exceed buying power.
+    if execution.market_session == MarketSession::Extended {
+        let _counter_trade_submission_guard = cqrs.counter_trade_submission_lock.lock().await;
+
+        match preflight_counter_trade_submission(executor, &execution, None).await? {
+            CounterTradeSubmissionCheck::Skipped => return Ok(None),
```
0xgleb, line 2531: should fix: the new extended-hours block in process_queued_trade has its own preflight `Skipped => return Ok(None)` (this line) and `clamp_shares_to_reservation(&mut execution, ...)` just below, and the clamped `execution.shares` is what gets written into the enqueued PlaceHedge payload. The only test that reaches this branch (extended_hours_trade_enqueues_immediate_hedge_job) provisions 10 AAPL shares + $10k buying power against a 5-share hedge, so the preflight returns Allowed and the clamp never fires; it asserts job.shares == 5.0 un-clamped. The two skip tests (trade_above_threshold_skips_counter_trade_without_offchain_inventory, ..._still_skips_with_zero_inventory) run with a bare MockExecutor (Regular session), so they only hit the regular-path Skipped branch below, never this one. This is the financial-safety gate for the extended path and it's a separate copy of the regular branch that can drift. Please add two extended-session tests: one where the preflight skips (assert Ok(None), zero PlaceHedge jobs enqueued, position left unclaimed) and one where the reservation is smaller than the requested shares (assert the enqueued PlaceHedge.shares equals the clamped quantity).

src/trading/offchain/hedge.rs
```diff
@@ -47,6 +137,12 @@ pub(crate) struct PlaceHedge {
     pub(crate) executor: SupportedExecutor,
     pub(crate) threshold: ExecutionThreshold,
     pub(crate) offchain_order_id: OffchainOrderId,
+    #[serde(default = "default_market_session")]
```
0xgleb, line 138: should fix: PlaceHedge is a durable apalis job persisted as JSON, and `#[serde(default = "default_market_session")]` is the entire mechanism that keeps in-flight jobs enqueued by the pre-deploy binary (which lack this field) deserializable after rollout. Right now nothing tests that path: the hedge.rs tests all build PlaceHedge via struct literal with the field set, and the only from_slice in the suite (the conductor enqueue test) round-trips a freshly-serialized job that already has the field. The codebase already has this exact kind of legacy-payload test elsewhere (rebalancing/equity/job.rs, offchain/order.rs, usdc_rebalance.rs). Add a small unit test that serde_json::from_value/from_str on a PlaceHedge JSON object lacking market_session succeeds and yields MarketSession::Regular, so a dropped attribute or wrong default variant (both of which compile clean) gets caught.

src/onchain/accumulator.rs
```diff
@@ -76,20 +82,37 @@ fn check_asset_enabled(asset_enabled: bool, symbol: &Symbol) -> bool {
...
+    match session {
+        MarketSession::Regular => Ok(Some(session)),
+
+        MarketSession::Extended if extended_hours_enabled => {
+            debug!(target: "hedge", %symbol, "Extended hours session, will use limit order");
+            Ok(Some(session))
+        }
 
-    Ok(is_open)
+        MarketSession::Extended | MarketSession::Closed => {
```
0xgleb, line 105: should fix: this defer arm now folds in the new Extended-with-symbol-disabled case, which is reachable in production (trading enabled, extended_hours_counter_trading disabled, broker in an Extended session lands here and returns None). Every accumulator test uses MockExecutor::new() (Regular) or .with_market_closed() (Closed); none drives an Extended session, so the Closed subcase is covered but the Extended-disabled subcase is exercised nowhere (the Extended-enabled->Some path is only covered indirectly via the conductor test). If this arm were ever miswired to return Some, a disabled symbol would get hedged with a market order during extended hours. Add a check_execution_readiness test with MockExecutor::new().with_market_session(MarketSession::Extended) and extended_hours_counter_trading Disabled asserting None, and optionally a companion asserting Some(Extended) when enabled.

src/trading/offchain/hedge.rs
```diff
@@ -209,7 +447,32 @@ impl Job<HedgeCtx> for PlaceHedge {
...
 
-            Some(Filled { .. } | Pending { .. } | Cancelled { .. }) | None => {}
+            // Cancelled here is unexpected -- a freshly Placed order should
+            // not be Cancelled by the time we re-load it. It is still a
+            // broker-confirmed cancellation, so route it through the shared
+            // terminal finalization: recording it as a failure would anchor
+            // the next attempt's client_order_id to the cancelled key (the
+            // broker would dedupe against a dead order) and drop any partial
+            // fill the cancellation retained.
+            Some(cancelled @ Cancelled { .. }) => {
```
0xgleb: should fix: this new arm routes a freshly-placed-then-Cancelled order through terminal_position_finalization + position_command_for_finalization, and the Complete path debits a retained partial fill from the position. The conductor twin dispatch_post_place_state got both a no-fill and a partial-fill Cancelled test in this PR, but the parallel copy here in PlaceHedge has none: every placer in the hedge suite yields Submitted (or Failed for the rejecting placer) after Place, never Cancelled, so this arm never runs. Since it's a hand-mirrored copy of the conductor logic that touches money, the two can drift undetected. Either drive a placer/aggregate setup that yields Cancelled at reload (asserting CompleteOffChainOrder debits net for a priced partial fill and CancelOffChainOrder for no-fill, mirroring the conductor tests), or factor the shared send-or-log wrapper so a single tested helper backs both sites (see the duplication note on the conductor arm).

src/conductor.rs
```diff
@@ -2499,7 +2572,31 @@ async fn dispatch_post_place_state(
             Ok(None)
         }
 
-        Some(Pending { .. } | Filled { .. } | Cancelling { .. } | Cancelled { .. }) => {
+        // Cancelled directly after Place is unexpected but is a genuine
+        // broker-confirmed cancellation: route through the shared terminal
+        // finalization instead of recording a failure, which would anchor the
+        // next attempt's client_order_id to the cancelled key and drop any
+        // retained fill. Mirrors the matching arm in `PlaceHedge::perform`.
+        Some(cancelled @ Cancelled { .. }) => {
```
0xgleb, line 3074: nit: this arm and the matching one in PlaceHedge::perform are the same ~12-line wrapper: `terminal_position_finalization(&cancelled).and_then(|f| position_command_for_finalization(f, id))`, then `if let Some(command) { position.send(...).await? } else { error!("...operator intervention required") }`, with a byte-identical operator-facing error string. The PR even self-documents it with "Mirrors the matching arm in PlaceHedge::perform." The money-critical decision (which command to emit) already lives in the two shared helpers, so this is low-stakes, but a small shared async helper (e.g. finalize_or_log_unpriced(position, symbol, id, &cancelled)) next to those helpers would keep the send + unpriced-fill error message in one place and let a single test cover both call sites.

src/trading/offchain/hedge.rs
```diff
@@ -47,6 +137,12 @@ pub(crate) struct PlaceHedge {
     pub(crate) executor: SupportedExecutor,
     pub(crate) threshold: ExecutionThreshold,
     pub(crate) offchain_order_id: OffchainOrderId,
+    #[serde(default = "default_market_session")]
+    pub(crate) market_session: st0x_execution::MarketSession,
```
0xgleb: nit, two small things on this field. (1) MarketSession is already imported unqualified at the top of the file and used unqualified in the production match arms, so `st0x_execution::MarketSession` here (and in default_market_session's return type/body) is redundant prefix noise; drop it to match the rest of the file. (2) self.market_session is read in perform only for the staleness log comparison (current_session vs self.market_session); the operative session is always re-fetched into current_session and that is what derives order_kind, so this serialized field can never affect a trade decision. A one-line doc comment noting it's diagnostic-only would stop a future reader from assuming it gates the session-dependent order_kind logic.

src/trading/onchain/trade_accountant.rs
```diff
@@ -237,6 +237,25 @@ pub(crate) enum TradeAccountingError {
...
+        source: Box<dyn std::error::Error + Send + Sync>,
+    },
+    #[error("Failed to fetch latest trade price for {symbol}")]
+    LimitPriceFetch {
+        symbol: st0x_execution::Symbol,
+        #[source]
+        source: Box<dyn std::error::Error + Send + Sync>,
+    },
+    #[error(
+        "Executor does not support fetching trade prices for \
```
0xgleb: nit: this message split across two lines with a backslash continuation collapses to about 76 chars (`#[error("Executor does not support fetching trade prices for {symbol}")]`), well under the 100-char limit, and the neighboring MarketSessionCheck/LimitPriceFetch variants are single-line (MarketSessionCheck's message is actually longer). cargo fmt won't reformat literals inside #[error], so collapse it to one line for grep-ability and consistency.

0xgleb (changes requested):

0xgleb (approved):

PR #935, opened 2026-06-25
fix: close the acknowledge->mark double-count with a bounded pending-ack set

0xgleb:
## What

Closes RAI-1157, and closes the cross-process double-count window it exposed.
Two things:

- `process-tx` accounts each fill exactly once: it refuses any fill already
  recorded in the `OnChainTrade` log and records a previously-unrecorded fill
  through the same `witness -> enrich -> acknowledge -> mark -> confirm` sequence
  the automated pipeline uses.
- The shared exactly-once pipeline now closes the `acknowledge -> mark` crash
  window: `Position` rejects an out-of-order / cross-process re-drive via a
  bounded, event-sourced pending-acknowledgement set (ADR 0010, supersedes ADR
  0005 in part).

## Why

Fill accounting de-duplicated against `Position`'s single-slot
`last_acknowledged_trade_id`. That slot is correct only under serialized,
in-order processing -- the property ADR 0005 relied on (the bot's
`concurrency(1)` worker plus a per-symbol in-process lock). `process-tx` runs in
a **separate process** from the bot, so that lock cannot serialize the two, and
re-running it -- on a fill the bot already accounted, or twice on a bot-missed
one -- slipped past the slot and double-counted the net position (and placed a
duplicate hedge). The same window is reachable for any concurrency > 1.
Double-counting a fill is a direct financial-integrity hazard.

## How

- **Bounded pending-acknowledgement set.** `AcknowledgeOnChainFill` records the
  applied `trade_id` in a `Position`-side set, in the same atomic batch as the
  position move, via a dedicated `OnChainFillApplied` event. The dedup rejects a
  fill whose `trade_id` is still in the set, so a re-drive is rejected regardless
  of how many fills intervened or how long it was delayed -- the set is
  event-sourced, survives crashes, and is visible cross-process. A fifth
  pipeline step (`ConfirmOnChainFillAcknowledged`, after the marker is durable)
  prunes it; the set is empty at rest and bounded by in-flight jobs.
- **Replay-safe by construction.** The set-insert is a dedicated event with zero
  pre-ADR-0010 occurrences, so the schema bump's full replay rebuilds the set
  empty instead of re-accumulating every historical `trade_id`.
- **Zero-migration upgrade bridge.** The single slot is retained: a fill applied
  under old code but unmarked at the deploy restart has no `OnChainFillApplied`
  event, so the slot (not the set) guards its re-drive. No SQL migration.
- **process-tx** records unrecorded fills through the full sequence and refuses
[...]

PR #928, opened 2026-06-25
feat: surface reconciled transfers distinctly in the transfer DTO

JuaniRios:
## What

- Adds a distinct `Reconciled` status to `EquityMintStatus`, `EquityRedemptionStatus`, and `UsdcBridgeStatus` DTOs carrying `reconciled_at`, `failure_reason`, and `reconcile_reason`; previously all three mapped aggregate `Reconciled` → DTO `Completed`, silently dropping failure context
- Mint `failure_reason` is `String` (non-optional — mint `Failed` always has a reason); redemption and USDC use `Option<String>` (timeout/pre-existing-field replay can produce `None`)
- `is_terminal()` updated to include `Reconciled` across all three transfer types
- `UsdcRebalance::Reconciled` aggregate state gains `failure_reason: Option<String>` with `#[serde(default)]`; `apply(OperatorReconciled)` derives it from the prior failed state (not stored in the event) so historical replays recover the reason without a migration
- `ReconcileReason` gains a `Display` impl to emit human-readable `reconcile_reason` strings in the DTO
- SPEC.md updated to document the new distinct `Reconciled` DTO status
- Dashboard: `statusStyle` returns amber for `reconciled` (substring, case-insensitive — also covers `OperatorReconciled` event name in the timeline); `isTerminalStatus` includes `reconciled` so `upsertTransfer` routes active→recent; recovery commands suppressed for reconciled transfers

## Why

[RAI-1156](https://linear.app/makeitrain/issue/RAI-1156/surface-reconciled-equityusdc-transfers-distinctly-in-the-transfer-dto) — Reconciled transfers previously rendered as `Completed` in the DTO, making operator-resolved stuck transfers indistinguishable from genuine successes on the dashboard during incident response.

## How

- `failure_reason` is intentionally absent from the `OperatorReconciled` event; `apply()` derives it from the pre-reconciliation state (e.g. `DepositFailed.reason`). This covers historical events persisted before the field existed — replaying an old `OperatorReconciled` still recovers the correct reason from the state that preceded it.
- `UsdcRebalance::SCHEMA_VERSION` bumped 5 → 6 because a stale `Reconciled` snapshot would silently deserialize `failure_reason` as `None` via `#[serde(default)]`, bypassing the derivation. The bump forces rebuild from events so the field is populated correctly.
- The `OperatorReconciled` evolve arm uses an exhaustive match (no wildcard) over all `UsdcRebalance` variants, forcing any future variant to be explicitly classified at compile time.

## Testing

[...]

PR #818, opened 2026-06-11
feat: force-fail mints stuck at MintRequested

JuaniRios:
## What
Implements [RAI-999](https://linear.app/makeitrain/issue/RAI-999) — Step 10 (final gap) of [RAI-978](https://linear.app/makeitrain/issue/RAI-978). `transfer fail --kind mint` can now force-fail a mint stuck at `MintRequested`.

## Why
A mint stuck at `MintRequested` (requested at the provider but never accepted) had no operator force-fail path — the CLI bailed with "cannot fail before acceptance," and the aggregate returned `AlreadyInProgress`. RAI-978's coverage matrix flagged this as the last gap.

## How
`FailAcceptance` is now valid from `MintRequested` (emitting `MintAcceptanceFailed`), and that event's `evolve` is broadened to start from `MintRequested` as well as `MintAccepted` → `Failed`. No new event or schema bump (reuses `MintAcceptanceFailed`). The CLI maps a `MintRequested` mint to `FailAcceptance` instead of bailing. CQRS aggregate-command flow throughout.

## Testing
Aggregate test: `FailAcceptance` from `MintRequested` → `MintAcceptanceFailed` → `Failed`. Adjusted an existing evolve test (the case it guarded, `MintAcceptanceFailed` from `MintRequested`, is now intentionally valid) to assert rejection from a terminal `Failed` state instead. 179 mint/cli tests pass; clippy + fmt clean.

## Anything else
Closes the recovery-CLI epic [RAI-978](https://linear.app/makeitrain/issue/RAI-978) coverage matrix.

PR #937, opened 2026-06-26
fix: process-tx fails closed on already-recorded fills

JuaniRios:
## What

Closes [RAI-1157](https://linear.app/makeitrain/issue/RAI-1157/process-tx-must-fail-closed-on-an-already-recorded-tx-hash-log-index).

- `process_found_trade` (`src/cli/trading.rs`) rewritten to follow the ADR-0005 witness/acknowledge sequence and fail closed when a fill is already fully accounted.
- New `position_fill_already_recorded` guard (`src/conductor.rs`): queries `PositionEvent::OnChainOrderFilled` events by `(tx_hash, log_index)` and is consulted on every path before `execute_acknowledge_fill`, closing both the crash-window and legacy-fill double-count holes.
- New `reconcile_post_place_state` in the CLI: mirrors `dispatch_post_place_state` from the normal pipeline; sends `PositionCommand::FailOffChainOrder` on broker failure so `pending_offchain_order_id` is cleared and the normal pipeline can re-hedge.
- `block_number` added as a stored field on `OnchainTrade`; `OnchainTradeBuilder` extended with `with_block_number` / `with_block_timestamp` helpers (`src/test_utils.rs`).
- `execute_witness_trade`, `execute_enrich_trade`, `execute_acknowledge_fill`, `execute_mark_acknowledged` promoted to `pub(crate)` for direct CLI reuse.
- SPEC.md documents the ADR-0005 exactly-once protocol for `process-tx` and its operational precondition.

## Why

- The old CLI bypassed the `OnChainTrade` aggregate and called `execute_acknowledge_fill` directly.
- The `Position` aggregate's single-slot `last_acknowledged_trade_id` guard only deduplicates the *immediately preceding* fill — once a newer fill advances the slot, a CLI retry for an older fill bypasses the guard and counts it twice.
- Two concrete holes: **crash-window** (fill A witnessed+acknowledged in Position, `mark_acknowledged` crashed, fill B advanced slot, CLI retry re-applied A) and **legacy-fill** (fill A in Position via old path, no `OnChainTrade` record, slot advanced by fill B, `None` branch re-applied A).

## How

- `process_found_trade` loads the `OnChainTrade` aggregate first: `Acknowledged` → fail closed (early return, no hedge); `Witnessed` → crash-recovery (skip re-witness, enrich if needed, proceed to acknowledge); `None` → new fill (hard bail if `block_number` or `block_timestamp` absent, witness, enrich, then acknowledge).
- After the branch, a single authoritative `position_fill_already_recorded` check runs on **all three paths** before `execute_acknowledge_fill` — the durable guard that closes both holes.
- `reconcile_post_place_state` reloads the `OffchainOrder` aggregate after placement and sends `PositionCommand::FailOffChainOrder` when the order landed `Failed` (or unexpected state).

## Testing

12 new unit tests in `src/cli/trading.rs`: fail-closed on acknowledged fill (no second event, no hedge); resume of witnessed-but-unacknowledged fill (exactly 1 fill event); witness+acknowledge of a new fill; CLI-then-normal-pipeline integration guard (`process_queued_trade` skips already-acknowledged fill); missing `block_number` and missing `block_timestamp` both produce explicit errors; failed broker placement clears `pending_offchain_order_id`; concurrent-witnessed resumes cleanly; concurrent-acknowledged fails closed; crash-window regression `process_tx_does_not_double_count_witnessed_fill_after_newer_fill_acknowledged`; legacy-fill `None`-path regression `process_tx_none_path_does_not_recount_legacy_position_fill`; successful broker submission sets `pending_offchain_order_id`.

## Anything else

**Operational precondition**: `process-tx` must be run while the live bot is stopped (or after the apalis accounting job for the fill has drained). The durable guard's `SELECT` and the CQRS `apply` are separate transactions — a concurrent live-bot delivery of the same fill can slip through the TOCTOU window. Documented in SPEC.md.

## Summary by CodeRabbit

* **New Features**
  * Improved trade fill accounting with exactly-once semantics, including fail-closed handling, crash-window resume, and stricter deduplication for recorded fills.
  * Refined offchain hedge/placement lifecycle handling with clearer reconciliation of in-flight vs failed vs cleared states.

* **Bug Fixes**
  * Prevented duplicate fill/event recording during retries and concurrent witness/ack races.
  * Improved resilience when block timestamp data is missing by using fallback resolution.

* **Documentation**
  * Updated the operational specification with a required “stop live bot / drain accounting safely” workflow.

* **Tests**
  * Expanded unit and integration coverage, including updated event ordering expectations.

### DIFF
--- a/src/cli/trading.rs
+++ b/src/cli/trading.rs
@@ -2,7 +2,6 @@
 
 use alloy::primitives::TxHash;
 use alloy::providers::Provider;
-use anyhow::Context;
 use async_trait::async_trait;
 use sqlx::SqlitePool;
 use std::io::Write;
@@ -20,8 +19,8 @@ use st0x_execution::{
 };
 
 use crate::conductor::{
-    execute_acknowledge_fill, execute_enrich_trade, execute_mark_acknowledged, execute_settle_fill,
-    execute_witness_trade,
+    FillAccountingOutcome, account_for_onchain_fill, execute_mark_acknowledged,
+    execute_settle_fill, is_expected_place_offchain_order_rejection,
 };
 use crate::offchain::order::{
     OffchainOrder, OffchainOrderId, OrderPlacementResult, OrderPlacer,
@@ -33,7 +32,6 @@ use crate::onchain::{OnChainError, OnchainTrade, TradeValidationError};
 use crate::onchain_trade::{OnChainTrade, OnChainTradeId};
 use crate::position::{Position, PositionCommand};
 use crate::symbol::cache::SymbolCache;
-use st0x_config::ExecutionThreshold;
 use st0x_config::{BrokerCtx, Ctx};
 use st0x_float_serde::format_float_with_fallback;
 
@@ -530,44 +528,69 @@ pub(super) async fn process_found_trade<W: Write>(
 
     writeln!(stdout, "🔄 Processing trade with TradeAccumulator...")?;
 
+    let trade_id = OnChainTradeId {
+        tx_hash: onchain_trade.tx_hash,
+        log_index: onchain_trade.log_index,
+    };
+
+    // Build stores. CLI path: per-invocation instances are fine (single-instance
+    // rule applies to the server path only; see AGENTS.md).
+    let onchain_trade_store = StoreBuilder::<OnChainTrade>::new(pool.clone())
+        .build(())
+        .await?;
     let (position_store, position_projection) = StoreBuilder::<Position>::new(pool.clone())
         .build(())
         .await?;
     let (offchain_order_store, _) = StoreBuilder::<OffchainOrder>::new(pool.clone())
         .build(())
         .await?;
-    let onchain_trade_store = StoreBuilder::<OnChainTrade>::new(pool.clone())
-        .build(())
-        .await?;
 
-    // Refuse a fill the bot has already recorded (witnessed or acknowledged),
-    // then account this unrecorded fill through the same exactly-once
-    // witness -> acknowledge -> mark -> settle sequence the automated pipeline
-    // uses, so it is recorded in the OnChainTrade log and a re-run is refused
-    // rather than double-counted. process-tx runs in a separate process from the
-    // bot, so an in-process lock cannot serialize them; the persisted
-    // pending-acknowledgement set (ADR 0010) makes the position itself reject an
-    // out-of-order cross-process re-drive, and the recovery-CLI no-concurrent-bot
-    // contract remains the operational backstop.
-    refuse_if_fill_already_recorded(
+    let Some(block_number) = onchain_trade.block_number else {
+        anyhow::bail!("Fill {trade_id}: missing block_number, cannot witness fill");
+    };
+
+    let FillAccountingOutcome::Accounted { trade_id } = account_for_onchain_fill(
+        pool,
         &onchain_trade_store,
         &position_store,
         &onchain_trade,
-        stdout,
-    )
-    .await?;
-    account_fill_exactly_once(
-        &onchain_trade_store,
-        &position_store,
-        &onchain_trade,
+        block_number,
         ctx.execution_threshold,
     )
-    .await?;
+    .await?
+    else {
+        writeln!(
+            stdout,
+            "Fill {trade_id} is already fully accounted. Nothing to do; \
+             the normal pipeline will hedge any unhedged position exposure."
+        )?;
+        return Ok(());
+    };
 
     let executor_type = ctx.broker.to_supported_executor();
     let base_symbol = onchain_trade.symbol.base();
 
-    // CLI test command uses MockExecutor (market always open)
+    match reconcile_existing_pending_order(
+        &offchain_order_store,
+        &position_store,
+        base_symbol,
+        stdout,
+    )
+    .await?
+    {
+        CliPendingReconciliation::NoPending | CliPendingReconciliation::Cleared => {}
+        CliPendingReconciliation::InFlight => {
+            mark_and_settle_fill(
+                &onchain_trade_store,
+                &position_store,
+                &trade_id,
+                &onchain_trade,
+            )
+            .await?;
+            return Ok(());
+        }
+    }
+
     let trading_enabled = ctx.is_trading_enabled(base_symbol);
 
     if !trading_enabled {
@@ -575,6 +598,13 @@ pub(super) async fn process_found_trade<W: Write>(
             stdout,
             "Trading disabled by configuration for {base_symbol}"
         )?;
+        mark_and_settle_fill(
+            &onchain_trade_store,
+            &position_store,
+            &trade_id,
+            &onchain_trade,
+        )
+        .await?;
         return Ok(());
     }
 
@@ -597,6 +627,13 @@ pub(super) async fn process_found_trade<W: Write>(
             stdout,
             "   (Waiting to accumulate enough shares for a whole share execution)"
         )?;
+        mark_and_settle_fill(
+            &onchain_trade_store,
+            &position_store,
+            &trade_id,
+            &onchain_trade,
+        )
+        .await?;
         return Ok(());
     };
 
@@ -607,34 +644,6 @@ pub(super) async fn process_found_trade<W: Write>(
         "Trade triggered execution for {executor_type:?} (ID: {offchain_order_id})"
     )?;
 
-    if let Err(error) = position_store
-        .send(
-            &params.symbol,
-            PositionCommand::PlaceOffChainOrder {
-                offchain_order_id,
-                shares: params.shares,
-                direction: params.direction,
-                executor: params.executor,
-                threshold: ctx.execution_threshold,
-            },
-        )
-        .await
-    {
-        error!(
-            %offchain_order_id,
-            symbol = %params.symbol,
-            "Failed to execute Position::PlaceOffChainOrder: {error}"
-        );
-    }
-
-    // Reuse a prior failed attempt's stashed OffchainOrderId as the broker-side
-    // idempotency anchor (mirroring the automated `execute_create_offchain_order`
-    // path) so a CLI retry after a transient broker failure dedupes instead of
-    // placing a second order. Fail fast on a load error rather than placing under
-    // a fresh key: a transient load failure while an anchor exists would submit a
-    // SECOND live broker order for shares the prior attempt already placed -- a
-    // double-hedge the fail-fast rule exists to prevent. Re-running the CLI
-    // reloads the anchor.
     let anchor = position_store
         .load(&params.symbol)
         .await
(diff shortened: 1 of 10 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/937?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#937** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/937?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/937?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#928** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/928?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#818** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/818?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#815** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/815?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#935](https://github.com/ST0x-Technology/st0x.liquidity/pull/935) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/935?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#808** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/808?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=937).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/cli/trading.rs
```diff
@@ -618,88 +741,102 @@ pub(super) async fn process_found_trade<W: Write>(
...
+            // Order submitted to the broker. The CLI does not enqueue a live
+            // polling job; the order will be reconciled to a terminal state by
+            // the order-status recovery sweep on the next bot startup.
+            writeln!(stdout, "Trade processing completed!")?;
+        }
+        // Pending/Filled: lifecycle bug — Place always emits Placed+Submitted or
+        // Placed+Failed atomically. None: Place silently failed (infrastructure
+        // error logged above). Both cases use the same cleanup; the reason string
+        // differs so the operator can distinguish them in logs.
+        unexpected_state @ (Some(OffchainOrder::Pending { .. } | OffchainOrder::Filled { .. })
```
0xgleb: should fix: this `unexpected_state` arm (the None / Pending / Filled case) is the only branch of `reconcile_post_place_state` with no test. The Failed and Submitted arms each got one, but the None sub-case (Place failing below the aggregate layer) is the most likely in practice, and it's the safety net that stops a position getting permanently stuck with a dangling `pending_offchain_order_id`. Worth a test that drives this arm (delete the OffchainOrder row before reconcile, or a store whose load returns None) and asserts the pending id ends up None.

src/cli/trading.rs
```diff
@@ -618,88 +741,102 @@ pub(super) async fn process_found_trade<W: Write>(
...
+            let error = match unexpected_state {
+                Some(OffchainOrder::Pending { .. } | OffchainOrder::Filled { .. }) => {
+                    "Offchain order in unexpected state directly after Place"
+                }
+                None => "Offchain order missing after Place",
+                Some(
+                    OffchainOrder::Failed { .. }
+                    | OffchainOrder::Submitted { .. }
+                    | OffchainOrder::PartiallyFilled { .. },
+                ) => unreachable!("covered by outer match arms"),
```
0xgleb: should fix: collapsing `Some(Pending | Filled) | None` into one arm and then re-matching to recover the variant means this `unreachable!` is hand-asserted, not compiler-verified. A future edit that moves Failed/Submitted/PartiallyFilled into the catch-all turns it into a real panic in a production CLI path, the same family as the unwrap/expect the repo bans. Splitting into two outer arms (Some(Pending | Filled) and None) with a shared cleanup helper removes both the re-match and the unreachable.

src/cli/trading.rs
```diff
@@ -526,12 +528,134 @@ pub(super) async fn process_found_trade<W: Write>(
...
+    // before touching the position. Decision tree (mirrors process_queued_trade):
+    //   Acknowledged -> fail closed (fill fully accounted).
+    //   Witnessed    -> crash-recovery: resume from acknowledge.
+    //   None (new)   -> validate block_timestamp, witness, enrich, then acknowledge.
+    //
+    // PRECONDITION: run while the live bot is stopped (or the apalis accounting
+    // job for this fill has been drained). The durable dedup guard below is
+    // best-effort: its SELECT and the CQRS apply are separate transactions, so a
+    // concurrent live-bot delivery of the same fill can slip through the window.
+    let block_timestamp = match onchain_trade_store.load(&trade_id).await? {
```
0xgleb: minor: this witness/resume/acknowledge decision tree is a near-duplicate of the same ADR-0005 logic in `process_queued_trade`, and the four `execute_*` primitives were widened to pub(crate) only to enable the copy. The two copies have already diverged (this one adds a reload-after-AlreadyFilled sub-branch the pipeline lacks), so correctness now depends on keeping two exactly-once implementations in lockstep. Consider extracting one shared orchestrator returning something like `enum FillRecord { AlreadyAccounted, Proceed { block_timestamp } }` called by both, which also re-privatizes the four primitives

src/cli/trading.rs
```diff
@@ -526,12 +528,134 @@ pub(super) async fn process_found_trade<W: Write>(
...
+    if !skip_acknowledge {
+        execute_acknowledge_fill(
+            &position_store,
+            &onchain_trade,
+            ctx.execution_threshold,
+            block_timestamp,
+        )
+        .await?;
+    }
+    execute_mark_acknowledged(&onchain_trade_store, &trade_id).await?;
```
0xgleb: minor: mark_acknowledged runs here, before PlaceOffChainOrder sets the pending id and before reconcile_post_place_state. If reconcile then errors (load fails, or the FailOffChainOrder send is rejected), the pending id is left set while the trade is already Acknowledged, so a CLI retry short-circuits at 'already fully accounted' and is_ready_for_execution blocks the pipeline too, until the startup sweep clears it. Moving mark_acknowledged to after a clean reconcile makes a retry re-drive the whole hedge-placement section instead of dead-ending

src/cli/trading.rs
```diff
@@ -1282,4 +1464,1004 @@ mod tests {
...
+            .unwrap();
+
+        let position = position_store
+            .load(&Symbol::new("AAPL").unwrap())
+            .await
+            .unwrap()
+            .expect("Position must exist after fill accounting");
+
+        assert!(
+            position.pending_offchain_order_id.is_some(),
```
0xgleb: minor: this asserts `pending_offchain_order_id.is_some()` but never checks an order was actually placed, and the repo's 'assertions must be specific' rule discourages is_some(). A bug that set an arbitrary non-None id without placing an order would still pass. Unwrap and validate the id, and add the `COUNT(*) ... OffchainOrderEvent >= 1` / `OnChainOrderFilled = 1` assertions the other ADR-0005 tests use.

src/conductor.rs
```diff
@@ -2059,6 +2062,35 @@ async fn execute_mark_acknowledged(
...
 
+/// Returns `true` when the `Position` aggregate already has a durable
+/// `OnChainOrderFilled` event for the given `(tx_hash, log_index)`.
+///
+/// Used by the CLI's `process-tx` resume path to guard against
+/// double-counting a witnessed-but-unacknowledged fill. The Position's
+/// single-slot `last_acknowledged_trade_id` guard only deduplicates the
+/// immediately preceding fill; if a newer fill has since advanced the
+/// slot, a CLI retry would re-apply the old fill without this durable check.
+pub(crate) async fn position_fill_already_recorded(
```
0xgleb, line 2257: minor: this guard hard-codes the event-type literal and the json_extract paths ($.OnChainOrderFilled.trade_id.tx_hash / .log_index) and leans on TxHash::to_string() matching the serde output. It's the linchpin of the exactly-once fix but is silently coupled to PositionEvent's serde tagging and TradeId's field layout, so a future serde rename/retag would make it return false and reopen the double-count hole with no compile error. A test pinning the serialized OnChainOrderFilled JSON shape, and/or deriving the event-type string from PositionEvent::event_type(), would make a refactor break loudly.

src/cli/trading.rs
```diff
@@ -526,12 +528,134 @@ pub(super) async fn process_found_trade<W: Write>(
...
+        }
+    };
+
+    // Authoritative durable dedup guard — consulted once on every path that
+    // reaches the acknowledge step. The single-slot Position guard is insufficient
+    // for CLI retries after a newer fill has advanced the slot (ADR-0005 crash
+    // window). This also guards the None path (fresh witness) against legacy fills
+    // whose Position record already exists but whose OnChainTrade record was not
+    // previously created.
+    let skip_acknowledge = position_fill_already_recorded(pool, &trade_id).await?;
```
0xgleb: nit: binding the positive predicate to `skip_acknowledge` and then using `if !skip_acknowledge` is a double negative (the trailing `// skip when already recorded` comment is a tell). Inlining it as `if !position_fill_already_recorded(...).await? { ... }`, or binding `already_recorded`, reads more directly.

src/cli/trading.rs
```diff
@@ -576,7 +700,7 @@ pub(super) async fn process_found_trade<W: Write>(
         "Trade triggered execution for {executor_type:?} (ID: {offchain_order_id})"
     )?;
 
-    if let Err(error) = position_store
+    position_store
         .send(
             &params.symbol,
             PositionCommand::PlaceOffChainOrder {
```
0xgleb, line 664: minor: this Place send uses .await? so a benign rejection (e.g. already-pending) propagates and makes process-tx exit with an error after the fill is already accounted, whereas the pipeline this mirrors treats the same rejection as a graceful skip. In practice is_ready_for_execution gates this out unless the live bot races a write (which the run-while-stopped precondition forbids), so it's mostly a divergence-from-pipeline note rather than a live bug. Matching the pipeline (log and skip instead of ?) would avoid a confusing operator-facing failure.

Katomoto (commented):
Logic looks correct. One suggestion: the early return on line prevents the cleanup function from running — worth adding a finally block.

0xgleb (changes requested):

src/cli/trading.rs
JuaniRios: Fixed in 9f0812546e94972b717b3113370b66832a6902e8.

src/cli/trading.rs
JuaniRios: Fixed in 9f0812546e94972b717b3113370b66832a6902e8.

src/cli/trading.rs
JuaniRios: Fixed in 9f0812546e94972b717b3113370b66832a6902e8.

src/cli/trading.rs
JuaniRios: Fixed in 9f0812546e94972b717b3113370b66832a6902e8.

src/cli/trading.rs
JuaniRios: Fixed in 9f0812546e94972b717b3113370b66832a6902e8.

src/conductor.rs
JuaniRios: Fixed in 9f0812546e94972b717b3113370b66832a6902e8.

src/cli/trading.rs
JuaniRios: Fixed in 9f0812546e94972b717b3113370b66832a6902e8.

src/cli/trading.rs
JuaniRios: Fixed in 9f0812546e94972b717b3113370b66832a6902e8.

0xgleb (approved):

PR #940, opened 2026-06-26
feat: durably record in-flight CCTP burn tx to prevent double-burn

JuaniRios:
## What

- Closes [RAI-1168](https://linear.app/makeitrain/issue/RAI-1168).
- Durably records an in-flight CCTP `depositForBurn` tx hash so a per-attempt timeout or a crash between broadcasting the burn and confirming its receipt can never trigger a second burn (USDC burned twice but minted once = direct loss).

## Why

- The pre-existing crash-recovery scan (`find_recent_burn`) is mempool-blind: it adopts only a **mined** burn. A burn broadcast but not yet mined when the per-attempt timeout fired was invisible to the scan, so the redrive re-burned it.
- A double-burn is unrecoverable loss, so the redrive path must distinguish "broadcast, possibly in flight" from "never broadcast" before deciding whether to burn.

## How

- **Two-phase burn.** Split the burn into `submit_burn` (broadcasts, returns the tx hash *without* awaiting the receipt) and `confirm_burn` (awaits + validates the `MessageSent` receipt). Between them, `RecordPendingBurn` durably commits the hash onto `BridgingSubmitting` (new `pending_burn_tx` field). `ClearPendingBurn` is sent before *every* (re)broadcast so a retry can't confirm against a stale hash.
- **Uncancellable critical section.** The `submit_burn -> RecordPendingBurn` sequence runs on a detached `tokio::spawn` task bounded by `BURN_BROADCAST_TIMEOUT`, so the job timeout can't cancel between broadcast and record.
- **Status-check-before-scan on resume.** Resume checks the recorded tx's `burn_status` before falling back to the scan: mined adopts (re-validated via `confirm_burn`), pending redrives (never reburns), reverted falls through to scan-or-reburn, and **dropped** pages the operator (terminal `BurnTxDropped`, no auto-reburn — a load-balanced RPC can misreport a still-pending burn as dropped). If nothing was ever recorded and the scan is empty, resume fails closed (`BurnSubmitInconclusive`) rather than reburning.
- **Drop policy mirrors the wallet.** `burn_status` reports `Dropped` only after a grace window plus a consecutive-mempool-absence threshold (`BurnDropConfig`), matching `wait_for_receipt`, so a still-pending tx is never misclassified.
[...]

PR #927, opened 2026-06-25
fix: reconcile zombie transfer jobs that latch rebalancing

JuaniRios:
## What

- Add **zombie reconciliation** to the USDC and equity rebalancing dedupe guards (`in_flight_usdc_transfer`, `in_flight_equity_transfer`): a retryable `Failed` transfer job whose aggregate is already terminal is marked `Killed` (race-safe) and skipped, so a dead job can't latch rebalancing — while a genuinely live retryable job still suppresses duplicates.
- Fix a latent runtime bug: the `job` payload is a JSON BLOB (apalis `JsonCodec`), now read as `Vec<u8>` + `serde_json::from_slice` (decoding as `String` errors `BLOB` vs `TEXT`).
- Add `TokenizedEquityMint::is_terminal()` / `EquityRedemption::is_terminal()`; USDC reuses `UsdcRebalance::holds_rebalance_guard()`.
- Log before the guard's conservative early-returns (store-not-wired -> `debug`, aggregate-not-found -> `warn`) so a stuck guard is attributable, not silent.
- Document apalis `Jobs`-table status/payload semantics in `docs/conductor.md`.

## Why

- A dead apalis `Jobs` row latched USDC rebalancing off in **both** directions in prod (~23.5h). [RAI-1155](https://linear.app/makeitrain/issue/RAI-1155/usdc-rebalancing-dedupe-guard-counts-a-finished-apalis-job-as-in)
- A `Failed AND attempts < max_attempts` row is **not** terminal — apalis `fetch_next.sql` re-fetches it (ignoring `done_at`), so it is a live job. The latch was a *zombie*: such a row whose aggregate was already completed via the `resume` operator, while the fail-stop circuit breaker kept the worker from resolving it.
- The issue's original idea (drop the `Failed` branch) is **unsafe**: on restart apalis re-drives the old job (original aggregate id) while a new check enqueues a second transfer (new id) — a duplicate-transfer race with no cross-id idempotency.

## How

- Keep `OR (status = 'Failed' AND attempts < max_attempts)` in both guards (`transfer_live_job_for_id` unchanged).
- On the Failed path: load the aggregate; terminal (`!holds_rebalance_guard()` / `is_terminal()`) -> `kill_zombie_job` (`... WHERE id=? AND status='Failed' AND attempts < max_attempts`, race-safe no-op if apalis grabbed it) -> next candidate; non-terminal -> still in-flight.
[...]

PR #920, opened 2026-06-24
fix: standing CCTP allowance so burns don't race a fresh approve

JuaniRios:
## What

Fixes a prod incident: a dRPC load-balancing race between a just-submitted USDC
`approve` and the subsequent `depositForBurn` pre-flight `eth_call` caused CCTP
burns to revert with "ERC20: transfer amount exceeds allowance", permanently
latching USDC rebalancing. [RAI-1151](https://linear.app/makeitrain/issue/RAI-1151)

## Why

- The previous code approved the exact burn amount immediately before each
  `depositForBurn`. Under dRPC load balancing, a lagging node could serve the
  `depositForBurn` pre-flight `eth_call` before it saw the approve block, reading
  allowance = 0 and reverting.
- A CCTP burn revert with no retry caused the apalis job to fail permanently,
  latching `usdc_in_progress` and blocking all future rebalances until manual
  operator intervention.

## How

- **Standing `U256::MAX` allowance** (`ensure_standing_allowance`): the burn path
  checks allowance against a `U256::MAX / 2` threshold and only submits a new
  `approve(TokenMessenger, U256::MAX)` when it drops below that line. On the
  steady-state hot path (allowance >> threshold) no approve tx is sent.
- **`wait_for_node_sync` gate after approve**: after the cold-path approve
  confirms, the bridge polls `eth_blockNumber` through the load-balanced endpoint
  until it observes a block at or above the approve block, shrinking the race
  window before the burn.
- **One-shot retry** (`retry_burn_if_revert`): if `depositForBurn` still reverts,
  the bridge re-runs `ensure_standing_allowance`, re-queries the Circle
  fast-transfer fee (to avoid a stale fee from a spike during the ~30 s sync
  window), then retries exactly once. Non-revert errors (transport timeouts,
  connection failures) are NOT retried. The `#[cfg(test)]`
  `deposit_for_burn_with_allowance_retry` helper exercises the endpoint-level
  path without a fee re-query; production callers always use `burn_internal` which
  does re-query via `retry_burn_if_revert`.
- **`EvmError::is_revert()` + `CctpError::is_revert()`**: exhaustive
  compile-enforced classifiers. A new `EvmError` variant forces an explicit
  revert/non-revert decision before the code compiles. `CctpError::is_revert()`
  delegates to `EvmError::is_revert()` for `Evm` variants and checks
  `as_revert_data()` for bare `Contract` variants.
- **`apply_sync_result`**: pure function that maps sync failure to the original
[...]

PR #948, opened 2026-06-29
fix: redrive inconclusive Alpaca withdrawals

JuaniRios:
## What

- Closes [RAI-1177](https://linear.app/makeitrain/issue/RAI-1177/alpaca-usdc-withdrawal-poll-timeout-false-negative-re-withdraws-and).
- Makes Alpaca->Base USDC withdrawals durable when Alpaca polling is inconclusive: `poll_and_confirm_withdrawal` returns `WithdrawalPollInconclusive` instead of emitting `FailWithdrawal`, so `UsdcRebalance::Withdrawing` and the USDC guard stay held.
- Re-polls the same Alpaca transfer ID through `TransferUsdcToMarketMaking` redrives, with a 4h operator alert deadline and slower 30m post-deadline cadence.
- Re-arms stranded `Withdrawing{AlpacaToBase}` and `WithdrawalComplete{AlpacaToBase}` states on startup and timeout sweep instead of clearing the guard.
- Updates manual `stox transfer resume` retry behavior and the operator runbook for inconclusive withdrawal polls.

## Why

- A withdrawal poll timeout used to look terminal even when Alpaca may have completed the transfer, which released the guard and let the next rebalance tick start another withdrawal.
- Prod incident 2026-06-26 (~14:13-14:28 UTC) stranded ~7,321 USDC in the bridging wallet, drained Alpaca USD, and blocked USDC rebalancing for ~28h until manual recovery.
- Alpaca's explicit `TransferStatus::Failed` is the determinate failure signal; API errors, timeouts, and missing UUIDs are not proof that funds never moved.
- This is the Alpaca-withdrawal counterpart to [RAI-1168](https://linear.app/makeitrain/issue/RAI-1168), which made CCTP burn tracking durable.

## How

- Adds `UsdcTransferError::WithdrawalPollInconclusive { id, initiated_at, source }` and maps indeterminate Alpaca poll errors to it while preserving the existing `TransferStatus::Failed` -> `FailWithdrawal` path.
- Handles `WithdrawalPollInconclusive` in `TransferUsdcToMarketMaking::perform` by enqueueing a delayed redrive and returning `Ok(())`, so apalis retry budget is not consumed for safe re-polls.
- Extends `UsdcRebalance::is_resumable_mid_flight_data` so `Withdrawing{AlpacaToBase}` and `WithdrawalComplete{AlpacaToBase}` are resumable, while `WithdrawalSubmitting{AlpacaToBase}` remains excluded because it has no persisted Alpaca transfer ID.
- Changes `RebalancingService` timeout/startup recovery to use `transfer_live_job_for_id` and `RearmPolicy::AlpacaToBaseIdempotentRedrive`, blocking duplicate redrives for live apalis rows including retryable `Failed` rows.
- Threads `Withdrawing.initiated_at` through resume errors and uses the withdrawal initiation timestamp for the 4h alert deadline.

## Testing

- `nix run .#ci` passed end to end: workspace checks, `cargo nextest run --workspace --all-features` (`3037` passed, `5` skipped, `2` flaky passed on retry), clippy, fmt, DTO generation, dashboard lint, and `svelte-check`.
- Added coverage for indeterminate Alpaca poll errors, `TransferNotFound`, explicit Alpaca `Failed`, complete-after-inconclusive, persisted `initiated_at`, deadline alert behavior, startup/sweep re-arm behavior, live-job gating, and CLI redrive retry behavior.

## Screenshots

N/A.

## Anything else

- `UsdcRebalance::SCHEMA_VERSION` bumps from 5 to 6 so stale snapshots rebuild with the corrected `Withdrawing.initiated_at` semantics.
- A permanently absent Alpaca UUID still requires direct operator recovery after verifying no funds moved; the runbook documents this limitation and keeps the guard held rather than risking a re-withdrawal.

---
<a href="https://app.blacksmith.sh/ST0x-Technology/codesmith/st0x.liquidity/pr/948"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785329075&installation_id=125569124&pr_number=948&repository=ST0x-Technology%2Fst0x.liquidity&return_to=https%3A%2F%2Fgithub.com%2FST0x-Technology%2Fst0x.liquidity%2Fpull%2F948&signature=b55a233d7084c81a3a530ce5088403a3a2eaf9a4f690d8a37c4c822315823728"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

* **New Features**
  * USDC rebalancing now idempotently re-arms and resumes Alpaca→Base withdrawals after restarts, including reconciling when a CCTP burn is already present.
  * Indeterminate Alpaca withdrawal polling continues via delayed redrive, with operator paging after 4 hours.

* **Bug Fixes**
  * Prevents immediate failures for inconclusive polling outcomes, including cases where Alpaca reports a failure with an on-chain tx hash.
  * Corrects withdrawal initiation/timing semantics so resumed retries use the original initiation time.

* **Documentation**
  * Updated the CLI Operations Guide with manual recovery steps and expected outcomes for stuck withdrawals.

### DIFF
--- a/docs/cli-ops.md
+++ b/docs/cli-ops.md
@@ -260,6 +260,46 @@ Not covered by `transfer recheck` yet:
   per-id filter; each succeeds or fails independently and failures are reported
   as counts with a non-zero exit). Requires the bot to be running.
 
+### Withdrawal poll inconclusive (Alpaca->Base stuck at `Withdrawing`)
+
+An Alpaca withdrawal poll that returns an indeterminate error (timeout, network
+failure, non-retried API error) leaves the aggregate in `Withdrawing` with the
+guard held. It does NOT send `FailWithdrawal` and does NOT release the guard.
+The job schedules an unbounded delayed redrive that re-polls the same Alpaca
+transfer ID (idempotent -- never re-initiates the withdrawal).
+
+**Operator alert after 4 hours**: if polling remains inconclusive for more than
+4 hours from `Withdrawing.initiated_at` (a durable aggregate timestamp -- the
+countdown survives bot restarts), the job begins paging the operator on every
+redrive while still keeping the guard held and continuing to re-poll
+automatically. The alert message includes the transfer UUID and the elapsed
+time. Normal transient outages (< 4 hours) never fire an alert.
+
+When alerted, first check whether Alpaca connectivity/credentials are intact:
+
+    # Check the withdrawal status directly via the Alpaca API or dashboard.
+    # To manually re-poll immediately:
+    stox transfer resume --kind usdc --id <uuid> --direction to-raindex
+
+This re-polls Alpaca for the recorded transfer and proceeds normally if the
+withdrawal has completed (the common case), or emits `FailWithdrawal` if Alpaca
+reports Failed with no tx hash. If Alpaca reports Failed with a tx hash, polling
+stays inconclusive and the guard remains held. The `--direction` must be
+`to-raindex` for AlpacaToBase.
+
+**Known limitation -- permanent `TransferNotFound`**: if `transfer resume`
+consistently reports inconclusive and Alpaca's dashboard confirms the withdrawal
+UUID was never initiated or is genuinely absent from Alpaca's records, the
+automatic re-poll will loop indefinitely. This is intentional and safe: the
+guard stays held, no re-withdrawal occurs, and funds are not stranded. However,
+rebalancing remains blocked until resolved. Distinguish this from a transient
+Alpaca outage (where re-poll will eventually succeed) by verifying directly on
+the Alpaca dashboard or API that no withdrawal with the recorded UUID exists. A
+force-fail recovery verb for this exact case is a tracked follow-up. No current
+CLI command clears a `Withdrawing` aggregate whose Alpaca UUID is genuinely
+absent; escalate to the on-call engineer for direct recovery after confirming no
+funds moved.
+
 ### Clearing a pre-burn guard latch
 
 Use `fail-usdc-transfer` when a USDC rebalance is stranded at
--- a/src/rebalancing/trigger/mod.rs
+++ b/src/rebalancing/trigger/mod.rs
@@ -556,6 +556,17 @@ enum UsdcTimeoutCleanup {
         tracking: usdc::UsdcRebalanceTracking,
         elapsed: Duration,
     },
+    /// The timed-out aggregate is in an Alpaca-to-Base pre-burn state whose
+    /// source withdrawal may already have moved funds, with no live apalis job.
+    /// The caller must re-enqueue a fresh `TransferUsdcToMarketMaking` redrive
+    /// rather than clearing the guard. Resume is idempotent for these states:
+    /// `Withdrawing` only re-polls Alpaca, and `WithdrawalComplete` scans for an
+    /// existing burn before submitting one. The guard stays held throughout.
+    ReArmAlpacaToBaseWithdrawal {
+        tracking: usdc::UsdcRebalanceTracking,
+        elapsed: Duration,
+        amount: Usdc,
+    },
 }
 
 /// How [`RebalancingService::rearm_stranded_transfers`] gates re-arm for a
@@ -575,6 +586,13 @@ enum RearmPolicy {
     /// means the redrive budget is exhausted with no driver -- strand it for an
     /// operator alert rather than silently resetting the counter to zero.
     MidFlightPreBurn,
+    /// Alpaca-to-Base states after the source withdrawal may have moved funds:
+    /// re-arm whenever no LIVE job row exists. Unlike `MidFlightPreBurn`, a
+    /// terminal (exhausted/Killed) job row does NOT block re-arm. Resume is
+    /// idempotent from these states and never starts a second Alpaca withdrawal,
+    /// so job-budget exhaustion means the chain lost its driver, not that the
+    /// transfer is unrecoverable.
+    AlpacaToBaseIdempotentRedrive,
 }
 
 /// A stranded USDC rebalance to re-arm on startup, with the [`RearmPolicy`] that
@@ -964,20 +982,35 @@ impl RebalancingService {
         };
 
         for id in candidate_ids {
-            // A timed-out transfer whose apalis row is still in flight is NOT
+            // A timed-out transfer whose apalis row is still live is NOT
             // abandoned: the job will resume and may have an irreversible on-chain
-            // withdraw/burn already submitted. Skip the cleanup entirely so we
-            // neither clear the in-progress guard / inventory inflight nor mark it
-            // timed-out (which would make the reactor ignore its eventual terminal
-            // event and latch the guard forever). Let apalis drive it to terminal.
-            if self.transfer_in_flight_for_id(&id).await {
-                warn!(
-                    target: "rebalance",
-                    aggregate_id = %id,
-                    "USDC transfer exceeded its timeout but its apalis row is still in \
-                     flight; deferring cleanup to the job rather than clearing the guard"
-                );
-                continue;
+            // withdraw/burn already submitted. "Live" includes retryable Failed
+            // rows (`attempts < max_attempts`), because apalis will re-fetch them
+            // on its own. Skip the cleanup entirely so we neither clear the
+            // in-progress guard / inventory inflight nor mark it timed-out (which
+            // would make the reactor ignore its eventual terminal event and latch
+            // the guard forever). Let apalis drive it to terminal.
+            match self.transfer_live_job_for_id(&id).await {
+                Ok(true) => {
+                    warn!(
+                        target: "rebalance",
+                        aggregate_id = %id,
+                        "USDC transfer exceeded its timeout but its apalis row is still live; \
+                         deferring cleanup to the job rather than clearing the guard"
+                    );
+                    continue;
+                }
+                Ok(false) => {}
+                Err(error) => {
+                    warn!(
+                        target: "rebalance",
+                        aggregate_id = %id,
+                        ?error,
+                        "Failed to query live USDC transfer job row; treating transfer as live \
+                         for this sweep tick to avoid unsafe timeout cleanup"
+                    );
+                    continue;
+                }
             }
 
             let Some(cleanup) = self.cleanup_timed_out_usdc_rebalance(&id, now).await? else {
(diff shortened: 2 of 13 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/948?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#948** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/948?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/948?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#940** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/940?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#927** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/927?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#921** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/921?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#920** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/920?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=948).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/rebalancing/trigger/mod.rs
```diff
@@ -972,18 +990,20 @@ impl RebalancingService {
...
-            // timed-out (which would make the reactor ignore its eventual terminal
-            // event and latch the guard forever). Let apalis drive it to terminal.
-            if self.transfer_in_flight_for_id(&id).await {
+            // withdraw/burn already submitted. "Live" includes retryable Failed
+            // rows (`attempts < max_attempts`), because apalis will re-fetch them
+            // on its own. Skip the cleanup entirely so we neither clear the
+            // in-progress guard / inventory inflight nor mark it timed-out (which
+            // would make the reactor ignore its eventual terminal event and latch
+            // the guard forever). Let apalis drive it to terminal.
+            if self.transfer_live_job_for_id(&id).await? {
```
0xgleb: should fix: swapping `transfer_in_flight_for_id` (a `bool` that logged and returned `true` on any sqlx/serde error) for `transfer_live_job_for_id(...).await?` quietly changes the failure mode. The old version kept a per-candidate fail-safe and let the loop `continue`; the new `?` propagates a transient SQLite error all the way out and aborts the whole `expire_stuck_usdc_rebalances` loop for that tick, so every remaining USDC candidate misses its timeout handling. It is safe for `Withdrawing{AlpacaToBase}` (the guard just stays held), but it degrades cleanup for other pre-burn states whenever the DB hiccups, and the error is only logged a level up so it is invisible. Suggest matching the error here, logging, and treating it as live with `continue` instead of `?`-propagating.

src/rebalancing/usdc/manager.rs
```diff
@@ -1445,16 +1462,51 @@ impl<
...
-                    .await?;
-                return Err(UsdcTransferError::AlpacaWallet(error));
+                // CONSERVATIVE FAIL-CLOSED: every `AlpacaWalletError` returned by
+                // `poll_transfer_until_complete` (including ApiError{4xx/5xx},
+                // TransferTimeout, Reqwest, ParseError, TransferNotFound, and
+                // InvalidStatusTransition) is treated as INDETERMINATE -- the
+                // withdrawal may have already succeeded on Alpaca's side even if
+                // the poll did not confirm it. This is intentional: Alpaca's
+                // documented determinate terminal failure is delivered as
+                // `Ok(transfer)` with `status == TransferStatus::Failed`, which is
```
0xgleb: should fix: this comment is where the PR's one remaining auto-release path rests. It treats `Ok(transfer)` with `status == Failed` as a determinate "funds never left" terminal and emits `FailWithdrawal` just below (the `status != Complete` branch), which releases the rebalance guard. But the determinacy claim is not pinned: the only citation is the get-by-id endpoint URL, not Alpaca docs stating a FAILED outgoing crypto-wallet transfer guarantees no on-chain broadcast, and the test fixture is hand-written `status:FAILED, tx_hash:null`. A FAILED transfer carrying a non-null `tx` would still release the guard and let the next tick start a second withdrawal. Pin it with a versioned lifecycle-doc link or a recorded real FAILED payload, or gate the release on FAILED-plus-null-tx.

src/rebalancing/usdc/manager.rs
```diff
@@ -1445,16 +1462,51 @@ impl<
...
+                // Recovery for a permanently-absent UUID is operational (see
+                // docs/cli-ops.md "Withdrawal poll inconclusive"), not automated.
+                //
+                // Alpaca Broker API: GET /v1/wallets/transfers/{transfer_id}
+                // (docs.alpaca.markets/reference/getbrokerwallettransferbyid).
+                // The transfer lifecycle terminates in COMPLETE or FAILED status
+                // delivered as a Transfer payload. HTTP 4xx/5xx responses are
+                // transient (auth, network, Alpaca-side load) and do not indicate
+                // the transfer's ultimate fate -- the same transfer ID must be
+                // re-polled to determine the outcome.
```
0xgleb: should fix: two things here. First, the cited doc (`getbrokerwallettransferbyid`) is the by-id endpoint, but `get_transfer_status` actually GETs the list endpoint `/v1/accounts/{account_id}/wallets/transfers` and filters client-side by UUID, so the citation points at the wrong contract. Second, reclassifying `TransferNotFound` as indeterminate (guard held, re-poll forever) is only safe if that list endpoint returns the complete transfer history unpaginated. If Alpaca paginates or caps it, an older in-flight UUID falls outside the window, produces a false `TransferNotFound`, and latches all USDC rebalancing until an operator steps in. The existing 4-element test only proves client-side filtering, not the absence of a cap. Cite the list endpoint's pagination contract and add a real-response test large enough to exercise any cap, or handle pagination.

src/rebalancing/usdc/job.rs
```diff
@@ -1129,6 +1235,57 @@ mod tests {
...
+                initiated_at: Utc::now()
+                    - chrono::Duration::from_std(WITHDRAWAL_POLL_ALERT_DEADLINE).unwrap()
+                    - chrono::Duration::seconds(1),
+            }
+        }
+
+        /// Exactly at the deadline boundary: `elapsed == WITHDRAWAL_POLL_ALERT_DEADLINE`.
+        /// Pins the `>=` comparison -- if it were accidentally changed to `>`, the
+        /// exact-boundary test fails while `after_deadline` (elapsed + 1s) still passes.
+        fn at_deadline() -> Self {
```
0xgleb: minor: this comment claims `at_deadline()` pins the `>=` comparison so a change to `>` would fail the exact-boundary test. It cannot: `at_deadline()` sets `initiated_at = now - DEADLINE` at one instant, but the handler computes `elapsed` at a strictly later instant, so `elapsed` is always strictly greater than the deadline, never equal. The alert fires under both `>=` and `>`, so the test passes either way and cannot catch that regression. Either drop the claim, or test the comparison as a pure function with `elapsed == deadline` directly.

src/rebalancing/trigger/mod.rs
```diff
@@ -3666,25 +3697,41 @@ impl RebalancingService {
...
+                            UsdcRebalance::BridgingFailed { .. } => RearmPolicy::RecoverableFailure,
+                            UsdcRebalance::Bridging { .. }
+                            | UsdcRebalance::AwaitingAttestation { .. }
+                            | UsdcRebalance::Attested { .. } => RearmPolicy::PostBurnResumable,
+                            // `resumable_post_burn_transfer` returns Some for exactly
+                            // BridgingFailed{BaseToAlpaca,burn_tx=Some}, Bridging,
+                            // AwaitingAttestation, and Attested -- all handled above.
+                            // Because resumable_post_burn_transfer itself uses `_ => None`,
+                            // new variants silently return None and never reach here.
+                            _ => unreachable!(
```
0xgleb: minor: this post-burn policy match uses `_ => unreachable!(...)`, while the adjacent mid-flight match just below names every non-resumable variant so a forgotten policy is a compile error rather than a runtime panic. The `BridgingFailed { .. }` arm also catches every shape although only `BridgingFailed{BaseToAlpaca, burn_tx=Some}` reaches it. It is safe today because of the outer `if let Some(..)` guard, but extending `resumable_post_burn_transfer` to return `Some` for a new shape would silently assign a policy (or panic) instead of failing the build. Worth mirroring the exhaustive mid-flight match and narrowing the `BridgingFailed` arm.

docs/cli-ops.md
```diff
@@ -260,6 +260,44 @@ Not covered by `transfer recheck` yet:
   per-id filter; each succeeds or fails independently and failures are reported
   as counts with a non-zero exit). Requires the bot to be running.
 
+### Withdrawal poll inconclusive (Alpaca->Base stuck at `Withdrawing`)
+
+An Alpaca withdrawal poll that returns an indeterminate error (timeout, network
+failure, non-retried API error) leaves the aggregate in `Withdrawing` with the
+guard held. It does NOT emit `WithdrawalFailed` and does NOT release the guard.
```
0xgleb: nit: this line says "does NOT emit `WithdrawalFailed`" (the event/state name) while a few lines down the same section says "emits `FailWithdrawal`" (the command). SPEC.md uses `FailWithdrawal` consistently in this construction. Mixing the command and state names in one runbook section is confusing mid-incident, since they are distinct CQRS artifacts. Suggest "does NOT send `FailWithdrawal`" here to match.

0xgleb (changes requested):

0xgleb (approved):

PR #950, opened 2026-06-29
RAI-1183: USDC rebalance withdrawable-cash cap should always apply

JuaniRios:
## What

Implements [RAI-1183](https://linear.app/makeitrain/issue/RAI-1183/usdc-rebalance-withdrawable-cash-cap-skipped-when-no-reserve).

This updates Alpaca-to-Base USDC trigger sizing so Alpaca withdrawable cash always caps the transfer amount, even when no cash reserve is configured.

## Why

When `reserved = None`, the trigger previously skipped the withdrawable-cash cap and sized transfers from gross offchain imbalance, bounded only by the operational limit. In production on 2026-06-29, that allowed the bot to size a transfer above Alpaca's actual withdrawable cash, causing the USD->USDC conversion path to fail and leaving the USDC rebalance guard stuck until manual intervention.

## How

- Computes Alpaca-to-Base withdrawable capacity for every offchain USDC excess, treating an omitted reserve as `0`.
- Keeps missing withdrawable cash as a hard skip because the transfer cannot be safely bounded.
- Splits the below-minimum cases so reserve-exhausted capacity and naturally low withdrawable cash report different skip reasons.
- Adds boundary coverage for exactly `$51` withdrawable cash and for withdrawable cash binding below the configured operational limit.

## Testing

- `nix develop .#ci-backend -c cargo check -p st0x-hedge`
- `nix develop .#ci-backend -c cargo clippy -p st0x-hedge --all-targets`
- `nix develop .#ci-backend -c cargo nextest run -p st0x-hedge alpaca_to_base`
- `nix develop .#ci-backend -c cargo nextest run -p st0x-hedge usdc`
- `nix run .#ci`
- GitHub CI for PR #950: green

## Screenshots

N/A

## Anything else

PR #950 was submitted through Graphite as v2.

### DIFF
--- a/src/rebalancing/trigger/usdc.rs
+++ b/src/rebalancing/trigger/usdc.rs
@@ -25,6 +25,25 @@ pub(super) enum UsdcRebalanceOperation {
     BaseToAlpaca { amount: Usdc },
 }
 
+/// Internal decision built inside the inventory read-lock, coupling each
+/// imbalance variant with its pre-computed capacity data. This eliminates
+/// the outer `Option<Option<_>>` and makes the post-lock dispatch exhaustive.
+enum UsdcImbalanceDecision {
+    NoImbalance,
+    TooMuchOffchain {
+        excess: Usdc,
+        withdrawable_capacity: WithdrawableCapacity,
+    },
+    TooMuchOnchain {
+        excess: Usdc,
+    },
+}
+
+enum WithdrawableCapacity {
+    Missing,
+    Available(Usdc),
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub(crate) enum UsdcTrackingEvent {
     Initiated,
@@ -236,13 +255,17 @@ const USDC_TRANSFER_MAX_DECIMAL_PLACES: u8 = 6;
 pub(crate) enum UsdcTriggerSkip {
     /// No USDC imbalance detected.
     NoImbalance,
-    /// Alpaca did not report settled withdrawable cash, so outbound
-    /// Alpaca-to-Base capacity cannot be proven reserve-safe.
+    /// Alpaca did not report settled withdrawable cash, so the withdrawable cap
+    /// cannot be applied and the transfer amount cannot be safely bounded.
     MissingWithdrawableCash,
-    /// Withdrawable cash is at or below the configured reserve, so no
-    /// Alpaca-to-Base capacity is available. Operators should investigate;
-    /// the reserve may need to be lowered or the broker rebalanced.
+    /// Withdrawable cash, after subtracting the configured reserve, leaves
+    /// capacity below the Alpaca minimum withdrawal. Lower the reserve or
+    /// rebalance the broker.
     ReserveCapacityExhausted,
+    /// Alpaca's withdrawable cash itself is below the Alpaca minimum withdrawal.
+    /// No reserve is configured; the broker simply has too little settled cash.
+    /// No operator action is needed; this clears once the broker settles more.
+    WithdrawableBelowMinimum,
     /// Imbalance exists but amount is below Alpaca's minimum withdrawal ($51).
     BelowMinimumWithdrawal { excess: Usdc },
     /// Arithmetic error during imbalance calculation.
@@ -301,7 +324,7 @@ pub(super) async fn check_imbalance_and_build_operation(
     usdc_limit: Option<Usdc>,
     reserved: Option<Usd>,
 ) -> Result<UsdcRebalanceOperation, UsdcTriggerSkip> {
-    let imbalance_check = {
+    let decision = {
         let inventory = inventory.read().await;
         let imbalance = inventory
             .check_usdc_imbalance_with_gross_offchain(threshold, reserved)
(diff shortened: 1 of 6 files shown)


JuaniRios:
* **#950** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/950?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/950?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=950).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/rebalancing/trigger/usdc.rs
```diff
@@ -313,16 +335,18 @@ pub(super) async fn check_imbalance_and_build_operation(
...
+        // configured. `alpaca_to_base_usdc_capacity` treats `None` reserve
+        // as zero, so "no reserve" means the full withdrawable cash is
+        // available as capacity. Capacity reads parse `withdrawable_cash_cents`,
+        // so binding the call to TooMuchOffchain only keeps Base-to-Alpaca
+        // rebalancing independent of withdrawable cash (which it must be).
+        // Holding one read guard for both values preserves TOCTOU safety.
+        let decision = match imbalance {
+            None => UsdcImbalanceDecision::NoImbalance,
+            Some(Imbalance::TooMuchOffchain { excess }) => {
+                let withdrawable_capacity = inventory
```
0xgleb, line 350: should fix: capacity is now computed for every `TooMuchOffchain` imbalance, so `cash_withdrawable` becomes the sole binding cap on no-reserve transfers, and a `None` value skips with `MissingWithdrawableCash` even when no reserve is set. The cents encoding is parse-tested, but the semantic that account settled USD cash actually bounds USDC/CCTP withdrawals to Base is asserted only in field comments, not a cited Alpaca doc, and the new tests set withdrawable directly so they never exercise the wire path. Worth pinning it with a doc citation on `AccountDetailsResponse.cash_withdrawable` or a recorded `/account` response integration test.

src/rebalancing/trigger/usdc.rs
```diff
@@ -1302,8 +1335,8 @@ mod tests {
         let result = check_imbalance_and_build_operation(&threshold, &inventory, None, None).await;
 
         assert!(
-            matches!(result, Ok(UsdcRebalanceOperation::AlpacaToBase { amount }) if !amount.inner().lt(float!(51)).unwrap_or(true)),
-            "Expected AlpacaToBase with amount >= 51, got {result:?}"
+            matches!(result, Ok(UsdcRebalanceOperation::AlpacaToBase { amount }) if amount == Usdc::new(float!(200))),
```
0xgleb: minor: the Ok-path assertions use `assert!(matches!(result, Ok(...) if amount == ...))` while the error-path tests in this same module use `assert_eq!`. `UsdcRebalanceOperation` derives `PartialEq`, so `assert_eq!(result, Ok(UsdcRebalanceOperation::AlpacaToBase { amount: Usdc::new(float!(N)) }))` works directly, prints both sides on failure, and matches the repo's 'assert_eq! with exact values' convention. Applies here and to the new no-reserve Ok-path tests.

src/rebalancing/trigger/usdc.rs
```diff
@@ -25,6 +25,24 @@ pub(super) enum UsdcRebalanceOperation {
...
+/// the outer `Option<Option<_>>` and makes the post-lock dispatch exhaustive.
+enum UsdcImbalanceDecision {
+    NoImbalance,
+    /// `withdrawable_capacity`:
+    /// - `None`      -> broker omitted withdrawable cash (-> MissingWithdrawableCash)
+    /// - `Some(cap)` -> withdrawable minus the configured reserve (reserve
+    ///   treated as $0 when unconfigured)
+    TooMuchOffchain {
+        excess: Usdc,
+        withdrawable_capacity: Option<Usdc>,
```
0xgleb: nit: `withdrawable_capacity: Option<Usdc>` uses `None` to mean specifically 'broker did not report withdrawable cash', but that meaning lives only in the doc comment. Since the whole point of this refactor is replacing `Option<Option<_>>` with a type that makes invalid states unrepresentable, a small `enum WithdrawableCapacity { Missing, Available(Usdc) }` would finish the job and let the compiler enforce it. You can wrap the capacity call with `capacity.map_or(Missing, Available)` so no inner plumbing changes.

src/rebalancing/trigger/usdc.rs
```diff
@@ -1723,6 +1756,161 @@ mod tests {
...
+
+        assert_eq!(
+            result,
+            Err(UsdcTriggerSkip::WithdrawableBelowMinimum),
+            "Withdrawable below $51 minimum with no reserve must return WithdrawableBelowMinimum, got {result:?}"
+        );
+    }
+
+    #[tokio::test]
+    async fn alpaca_to_base_transfers_full_excess_when_withdrawable_ample_no_reserve() {
```
0xgleb: nit: this looks functionally identical to `test_excess_above_minimum_triggers_alpaca_to_base` after its assertion was tightened to `== 200` in this PR: same 100/500 USDC, $500 withdrawable, no reserve, no limit, same expected 200. The extra `.with_offchain_gross_usd_cents(50_000)` doesn't change the exercised path because gross equals net offchain here. Consider dropping it, or if the intent is to guard against over-clamping, add a one-line comment to the existing test instead.

0xgleb (approved):

PR #969, opened 2026-07-01
feat: rebalance via the shared RaindexInventory

hardyjosh:
## What

Closes [RAI-1196](https://linear.app/makeitrain/issue/RAI-1196/liquidity-bot-rebalance-via-the-shared-raindexinventory).

- Adds a shared-**`RaindexInventory`** settlement path: in `managed` mode, rebalance `deposit4`/`withdraw4` route through the inventory (`rainlanguage/raindex.governance`) instead of the Raindex orderbook, and the read path becomes owner-aware for the vault-ownership cutover.
- Settlement mode is an explicit **`InventoryMode::Legacy | Managed`** enum, so a deploy can't silently conflate "no distinct inventory yet" with "inventory misconfigured". **prod and staging ship `legacy`** (orderbook settlement, unchanged behavior) until the RAI-1198 cutover.

> **Breaking config change:** every `[raindex]` section now requires `inventory_mode` (`"legacy"` | `"managed"`) and `vault_owner`. `inventory` is required only for `managed` and forbidden for `legacy` (validated at startup, no fallback).

## Why

- One-pool model: the Bebop hook and univ4 hook both hold `OPERATOR_ROLE` on the same inventory and draw from the same vaults. If the bot kept settling on the orderbook directly, its capital moves would bypass the inventory and desync from the venue adapters.
- `OPERATOR_ROLE` on the inventory is the trust boundary: the bot EOA holds the role and the inventory forwards `deposit4`/`withdraw4` to Raindex.

## How

- **Settlement mode:** `legacy` settles on the orderbook (bot-owned vaults, no preflight); `managed` settles through the `inventory` address. Replaces a fragile `inventory == orderbook` equality check.
- **Owner-aware reads:** the required, explicit `vault_owner` scopes every `vaultBalance2` read, vault-registry entry, and ClearV3/TakeOrderV3 order-owner match. Pre-migration it is the `[wallet]` EOA; flipped to the inventory at the cutover. The `(orderbook, vault_owner)` pair travels in a named `VaultRegistryId` struct so it can't be positionally swapped. Signer-specific uses stay on `order_owner()`.
- **Crash recovery** (`find_recent_withdrawal`): union scan over `OperatorWithdraw` (current) + `WithdrawV2` (pre-migration), newest-first, so a resume across the cutover adopts an in-flight withdraw instead of double-spending. **Fails closed** on any anomalous/undecodable log from a trusted address — skipping one could re-issue an irreversible withdraw.
- **Managed preflight:** verifies the bot holds `OPERATOR_ROLE` before rebalancing (retrying `hasRole` to absorb an RPC node lagging a fresh grant), else fails fast; skipped in `legacy`. Also best-effort revokes the stale bot→orderbook ERC20 allowance.
- **Typed reverts:** `InsufficientVaultLiquidity` is decoded from a revert across every shape it arrives in (call error, gas-estimation, mined replay); the USDC manager surfaces it as a distinct terminal error (operator reconciliation, no blind redrive) rather than an opaque wrap.
- New flake input `raindex-governance` exports `ST0X_RAINDEX_INVENTORY_ABI`.

## Testing

- `cargo nextest run` across config + evm + raindex + hedge (incl. e2e, `--all-features`): all pass. `cargo clippy --workspace --all-targets --all-features` and `cargo fmt --check`: clean.
- New unit/integration tests cover: fail-closed and same-vault cutover-ordering scans, revert decoding across all shapes, the `OPERATOR_ROLE` retry, the preflight legacy-skip/managed-attempt branches, `EvmCtx` field-mapping + mode validation, and `vault_owner != snapshot.owner` polling.
- A live Base-fork run against a *distinct* inventory is still pending — tracked in [RAI-1248](https://linear.app/makeitrain/issue/RAI-1248/test-coverage-distinct-inventory-rebalancing-path-preflight-e2e).

## Anything else

- **Ships inert.** prod and staging are `legacy` until the [RAI-1198](https://linear.app/makeitrain/issue/RAI-1198/redeploy-raindexinventory-owned-by-turnkey-migrate-vaults-regrant) cutover — the bot doesn't hold `OPERATOR_ROLE` on the inventory yet, so `managed` would fail the preflight and block boot. At cutover: redeploy the inventory under Turnkey, migrate vaults, grant the role, then flip `inventory_mode`→`managed`, set `inventory`, and flip `vault_owner`. The prod/staging TOMLs document the sequence (staging still needs its own inventory deploy).
- Follow-up: [RAI-1248](https://linear.app/makeitrain/issue/RAI-1248/test-coverage-distinct-inventory-rebalancing-path-preflight-e2e).
- Stack: [#970](https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/970) (RAI-1197 cross-venue fill watcher) is on top.

## Summary by CodeRabbit

* **New Features**
  * Added support for shared-inventory Raindex settlement with configurable legacy or managed modes.
  * Exposed the vault owner setting throughout the app for order, vault, and balance lookups.
  * Added clearer runtime reporting for inventory-related operations.

* **Bug Fixes**
  * Improved startup validation for Raindex configuration to catch invalid settlement settings early.
  * Updated withdrawal handling to surface insufficient vault liquidity more clearly.
  * Made allowance cleanup and role checks more robust during rebalancing startup.


hardyjosh:
* **#970** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/970?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#969** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/969?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/969?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=969).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/config/src/evm.rs
```diff
@@ -38,6 +38,23 @@ impl std::fmt::Display for IngestionCutoff {
...
+    /// `vaultBalance2` read, vault-registry entry, and order-owner fill match is
+    /// scoped by. Once the shared-inventory migration (RAI-1198) moves the
+    /// orders/vaults under the inventory contract, `msg.sender` to Raindex is the
+    /// inventory, so it becomes the vault owner and this must be set to the
+    /// `inventory` address. Left unset (the pre-migration default) it falls back
+    /// to the `[wallet]` address, matching the bot-EOA-owned vaults that exist
+    /// today. This is the ops cutover lever: flip it to the inventory address in
+    /// the same change that grants the bot `OPERATOR_ROLE` and migrates vaults.
+    #[serde(default)]
+    pub vault_owner: Option<Address>,
```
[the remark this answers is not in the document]
hardyjosh: Fixed: em dashes replaced with ASCII `--` on all lines this PR adds (swept the whole diff, not just these two lines).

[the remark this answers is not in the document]
hardyjosh: Fixed: `vault_owner` is now a required field with no fallback -- a missing value fails config parsing at startup, per the no-silent-defaults policy. Prod/staging/example/e2e configs set it explicitly (pre-migration = the wallet address; the migration cutover flips it to the inventory address in the same change that grants OPERATOR_ROLE).

crates/config/src/loader.rs
```diff
@@ -1074,13 +1065,33 @@ impl Ctx {
...
+    /// `RaindexInventory.Operator{Deposit,Withdraw}` events (and the `sender`
+    /// in pre-migration `WithdrawV2` events). It is NOT necessarily the address
+    /// that owns the Raindex vaults — see [`Self::vault_owner`].
+    ///
+    /// Named `order_owner` for historical reasons: before the shared-inventory
+    /// migration the signing wallet also owned the orders/vaults, so the two
+    /// concepts coincided.
     pub fn order_owner(&self) -> Address {
         self.order_owner
     }
```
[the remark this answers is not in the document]
hardyjosh: Fixed: replaced with ASCII `--`.

crates/config/src/loader.rs
```diff
@@ -1074,13 +1065,33 @@ impl Ctx {
...
+    ///
+    /// Every `vaultBalance2` read, vault-registry entry, and ClearV3/TakeOrderV3
+    /// order-owner fill match is scoped by this address. Post the shared-inventory
+    /// migration (RAI-1198) the inventory contract is `msg.sender` to Raindex, so
+    /// it owns the vaults; `[raindex].vault_owner` is set to the inventory address
+    /// at that cutover. Until then it falls back to the signing wallet, matching
+    /// the bot-EOA-owned vaults that exist pre-migration.
+    pub fn vault_owner(&self) -> Address {
+        self.evm.vault_owner.unwrap_or(self.order_owner)
+    }
```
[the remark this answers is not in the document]
hardyjosh: Fixed in tandem with the field definition: the `unwrap_or(self.order_owner)` fallback is gone; `Ctx::vault_owner()` now returns the required config value directly.

crates/raindex/src/lib.rs
```diff
@@ -43,6 +60,31 @@ pub enum RaindexError {
...
+    /// could not cover the requested amount (e.g. a concurrent Raindex clear
+    /// drained it first). Unlike the orderbook's `withdraw4`, which returns
+    /// `min(target, balance)`, the inventory settles atomically — it reverts
+    /// rather than forward a short fill. Retryable once the vault is refunded,
+    /// and distinguishable so the caller can reconcile instead of blindly
+    /// re-submitting the same reverting withdraw.
+    #[error(
+        "inventory vault under-funded for {token}: requested {requested}, vault could cover {received}"
+    )]
+    InsufficientVaultLiquidity {
```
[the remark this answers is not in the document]
hardyjosh: Fixed: wrapped with backslash continuation to match `MissingOperatorRole`.

crates/raindex/src/service.rs
```diff
@@ -55,30 +60,56 @@ const SCAN_FINALITY_MARGIN: u64 = 2;
...
+/// capital pool. The inventory's `deposit4`/`withdraw4` have signatures
+/// identical to `IRaindexV6` -- calldata is the same shape, only the target
+/// contract differs.
+///
+/// Reads (`vaultBalance2`, historical `WithdrawV2` scans) target the orderbook
+/// directly because the vault balances live in the orderbook's storage; the
+/// inventory is a proxy that owns orders/vaults on the orderbook, not a store of
+/// its own. Post-migration the balances are keyed under the inventory address —
+/// that owner key is supplied by the caller (see `Ctx::vault_owner`), so the
+/// read still hits the orderbook but under the correct owner.
```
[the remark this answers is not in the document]
hardyjosh: Fixed: replaced with ASCII `--`.

crates/raindex/src/service.rs
```diff
@@ -276,56 +349,113 @@ impl<E: Evm> RaindexService<E> {
...
-                {
-                    debug!(target: "orderbook", %tx_hash, from_block, "Found existing withdrawal during resume");
-                    return Ok(Some((tx_hash, event.withdrawAmountUint256)));
+                let block_after_from = log.block_number.is_some_and(|b| b > from_block);
+                if !block_after_from {
+                    continue;
+                }
+                let Some(tx_hash) = log.transaction_hash else {
+                    continue;
+                };
```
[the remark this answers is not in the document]
hardyjosh: Fixed: added a `warn!` (with the offending log) before the `continue`, matching the decode-failure branches.

crates/raindex/src/service.rs
```diff
@@ -357,6 +487,44 @@ impl<E: Evm> RaindexService<E> {
...
+        if has_role {
+            Ok(())
+        } else {
+            Err(RaindexError::MissingOperatorRole {
+                inventory: self.inventory_address,
+                wallet: self.owner,
+            })
+        }
+    }
+
```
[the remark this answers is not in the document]
hardyjosh: Fixed: added `verify_operator_role` pass/fail tests via the Asserter-mock pattern (mocked `OPERATOR_ROLE()` + `hasRole()` responses) and `revoke_orderbook_allowance` no-op / revoke / idempotency tests on the Anvil LocalEvm.

example.config.toml
```diff
@@ -29,6 +19,20 @@ beneficiary_entity_name = "Acme Corp"
...
+# OPERATOR_ROLE on this address for rebalancing to succeed (a startup preflight
+# checks this and refuses to start otherwise).
+inventory = "0x6b7B523fADd1677413AD92c9404C8F0796bACF6F"
+# Optional. Owner of the Raindex orders/vaults on-chain — scopes every
+# vaultBalance2 read, vault-registry entry, and ClearV3/TakeOrderV3 order-owner
+# fill match. Leave unset while the vaults are owned by the bot EOA (the
+# pre-inventory-migration state; it falls back to [wallet].address). Once the
+# vaults are migrated under the inventory contract, set this to the `inventory`
+# address so balance reads and fill matching query the right owner:
+#   vault_owner = "0x6b7B523fADd1677413AD92c9404C8F0796bACF6F"
```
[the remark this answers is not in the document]
hardyjosh: Fixed: `vault_owner` is now required (see the companion thread on `crates/config/src/evm.rs`) and this example sets it explicitly with a comment describing the pre-/post-migration values.

nix/raindex-governance.nix
```diff
@@ -0,0 +1,66 @@
...
+      };
+    in
+    ''
+      mkdir -p "$out/${dep.name}-${dep.version}"
+      unzip -q ${archive} -d "$out/${dep.name}-${dep.version}"
+    '';
+
+  soldeerDeps = pkgs.runCommand "raindex-governance-soldeer-deps" {
+    nativeBuildInputs = [ pkgs.unzip ];
+  } ("mkdir -p $out\n" + pkgs.lib.concatMapStrings unpackDep soldeerLock.dependencies);
```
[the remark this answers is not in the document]
hardyjosh: Skipping this one deliberately: the Soldeer vendoring block is duplicated in exactly two nix files, both wrapping a stable upstream pattern that has not changed since introduction. Extracting a shared `mkSoldeerDeps` helper couples the rain-orderbook ABI derivation (which feeds prod builds and is otherwise untouched by this PR) to a refactor with no behavioral payoff. If a third Soldeer consumer appears, the helper becomes worth it.

nix/raindex-governance.nix
```diff
@@ -0,0 +1,66 @@
...
+  };
+
+in
+rec {
+  inherit abi soldeerDeps;
+
+  abiEnv = {
+    ST0X_RAINDEX_INVENTORY_ABI = "${abi}/RaindexInventory.sol/RaindexInventory.json";
+  };
+}
```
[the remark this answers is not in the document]
hardyjosh: Fixed: dropped the unnecessary `rec`.

src/cli/rebalancing.rs
```diff
@@ -92,10 +93,17 @@ async fn build_equity_transfer_services(
...
 
-    let raindex = Arc::new(RaindexService::new(base_caller, ctx.evm.orderbook, wallet));
+    let raindex = Arc::new(RaindexService::new(
+        base_caller,
+        RaindexContracts {
+            inventory: ctx.evm.inventory,
+            orderbook: ctx.evm.orderbook,
+        },
+        wallet,
+    ));
```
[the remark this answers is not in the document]
hardyjosh: Fixed: added `onchain::raindex_contracts(&EvmCtx)` as the single conversion point; all four ctx-based construction sites use it now (test-only sites keep explicit literals since they construct from local constants).

src/cli/trading.rs
```diff
@@ -430,7 +367,9 @@ pub(super) async fn process_tx_with_provider<W: Write, P: Provider + Clone + 'st
 ) -> anyhow::Result<()> {
     let evm_ctx = &ctx.evm;
     let pyth_feed_ids = PythFeedIds::new(ctx.pyth_feed_ids());
-    let order_owner = ctx.order_owner();
+    // Matches ClearV3/TakeOrderV3 fills to our Raindex orders — owned by the
+    // inventory contract post-migration, the bot EOA before it.
+    let order_owner = ctx.vault_owner();
```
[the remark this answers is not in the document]
hardyjosh: Fixed: replaced with ASCII `--`.

src/conductor.rs
```diff
@@ -1284,6 +1257,94 @@ fn build_usdc_notifier(
...
+             configured RaindexInventory",
+        )?;
+    info!(
+        target: "inventory",
+        inventory = %ctx.evm.inventory,
+        "OPERATOR_ROLE preflight passed",
+    );
+    revoke_stale_orderbook_allowances(raindex_service, ctx).await;
+    Ok(())
+}
```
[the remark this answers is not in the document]
hardyjosh: Fixed: reworded to describe the shared-inventory cutover without the ticket reference (also stripped the same reference from the other doc comments this PR added).

crates/raindex/src/service.rs
```diff
@@ -757,6 +1000,318 @@ mod tests {
...
+            b256!("0x0000000000000000000000000000000000000000000000000000000000000009");
+        // Ascending block order (as get_logs returns): stale WithdrawV2 first,
+        // matching OperatorWithdraw last (newer).
+        let stale = withdraw_log_with(
+            Address::ZERO,
+            USDC_BASE,
+            other_vault,
+            from_block + 1,
+            U256::from(5u64),
+        );
```
0xgleb, line 1192: should fix: the stale WithdrawV2 here uses `other_vault`, so the per-branch vault filter alone excludes it and the newest-first ordering this test describes is never actually exercised. If `logs.iter().rev()` regressed to oldest-first, this test would still pass. Make the stale log use `TEST_VAULT_ID` with a matching sender and an older block, so only the ordering can prevent adopting the stale amount.

crates/raindex/src/service.rs
```diff
@@ -276,56 +349,118 @@ impl<E: Evm> RaindexService<E> {
...
+        // (event type mismatch → decode fails → skip).
         let filter = Filter::new()
             .from_block(from_block)
-            .address(self.orderbook_address)
-            .event_signature(IRaindexV6::WithdrawV2::SIGNATURE_HASH);
+            .address(vec![self.inventory_address, self.orderbook_address])
+            .event_signature(vec![
+                IRaindexInventory::OperatorWithdraw::SIGNATURE_HASH,
+                IRaindexV6::WithdrawV2::SIGNATURE_HASH,
+            ]);
```
0xgleb, line 406: should fix: there's no test where a WithdrawV2 and an OperatorWithdraw both match the same (token, vaultId, owner) in one scan window, which is the exact cutover scenario this union filter exists for. A regression in the iteration order or the else-if structure would return the stale pre-migration amount, and during crash recovery that feeds the burn reconciliation. Worth adding a case with both events on the same vault differing only by block number, plus a variant with the OperatorWithdraw in the earlier block.

crates/raindex/src/service.rs
```diff
@@ -467,7 +668,7 @@ impl<W: Wallet> Raindex for RaindexService<W> {
     ) -> Result<TransactionReceipt, RaindexError> {
         let receipt = self.evm.confirm::<OpenChainErrorRegistry>(tx_hash).await?;
 
-        info!(target: "orderbook", tx_hash = %receipt.transaction_hash, "Transaction confirmed");
+        info!(target: "inventory", tx_hash = %receipt.transaction_hash, "Transaction confirmed");
```
0xgleb, line 717: should fix: both withdraw submit paths got `.map_err(map_withdraw_revert)` in this PR but this confirm path didn't. If the node accepts the tx and a concurrent trade drains the vault before inclusion, the on-chain revert surfaces here as an opaque `RaindexError::Evm` instead of the typed `InsufficientVaultLiquidity`. The exposed caller is equity redemption (`submit_withdraw` + `confirm_tx_receipt`); the USDC manager is fine, its withdraws go through `withdraw()`, which waits for the receipt with the mapping applied. `map_withdraw_revert` passes non-matching errors through unchanged, so applying it here is safe for deposit confirmations too.

src/conductor.rs
```diff
@@ -1284,6 +1284,91 @@ fn build_usdc_notifier(
...
+/// hold `OPERATOR_ROLE` on the configured inventory or every rebalance
+/// deposit/withdraw reverts. Fails fast with a clear error rather than burning
+/// gas reverting on the first rebalance (the shared-inventory cutover grants
+/// the role before rebalancing runs).
+///
+/// Skipped when `inventory == orderbook`: that collapse means no distinct
+/// inventory contract is in play (integration tests and any not-yet-migrated
+/// setup), so there is no `OPERATOR_ROLE` to check and the orderbook allowance
+/// is the live one, not a stale artifact to revoke.
+async fn preflight_inventory_access<Chain: Wallet + Clone>(
```
0xgleb, line 1346: should fix: every fixture that actually reaches this preflight collapses inventory == orderbook (`create_test_ctx_with_order_owner` uses 0x1111... for both, the integration test uses TEST_ORDERBOOK for both), so the function always takes the early return and the real path (role check, allowance revoke, error context) has no coverage through the orchestration layer. A test with distinct addresses and mocked RPC responses could assert the Err on a missing role, Ok on a granted role, and that no role calls are issued in the collapsed case.

src/conductor.rs
```diff
@@ -1284,6 +1284,91 @@ fn build_usdc_notifier(
...
+    raindex_service: &RaindexService<Chain>,
+    ctx: &Ctx,
+) -> anyhow::Result<()> {
+    if ctx.evm.inventory == ctx.evm.orderbook {
```
0xgleb: should fix: address equality is doing the job of a mode discriminant here. A production misconfig that copies the orderbook address into `inventory` silently bypasses the preflight with only a debug log, precisely when the check matters most. An explicit enum on the ctx, something like `InventoryMode::Legacy | Managed { inventory }`, would make the two modes impossible to conflate. Fine as a tracked follow-up if you'd rather not grow this PR.

src/conductor.rs
```diff
@@ -1284,6 +1284,91 @@ fn build_usdc_notifier(
...
+
+/// Best-effort: revoke any stale pre-migration allowance the bot granted the
+/// orderbook directly. Deposits now approve the inventory instead, so a
+/// leftover orderbook allowance is dead capital-exposure surface. Idempotent
+/// (a no-op once zero) and non-fatal -- a failure here must not block startup.
+async fn revoke_stale_orderbook_allowances<Chain: Wallet + Clone>(
+    raindex_service: &RaindexService<Chain>,
+    ctx: &Ctx,
+) {
+    let mut revoke_tokens: Vec<Address> = vec![st0x_evm::USDC_BASE];
```
0xgleb: nit: the `mut` + `extend` pair can be one immutable expression: `std::iter::once(st0x_evm::USDC_BASE).chain(ctx.assets.equities.symbols.values().map(|equity| equity.tokenized_equity_derivative))`

crates/raindex/src/service.rs
```diff
@@ -130,34 +161,76 @@ impl<W: Wallet> RaindexService<W> {
...
     }
 
+    /// Idempotently revokes any stale ERC20 allowance the signing wallet granted
+    /// the **orderbook** for `token`. Before the shared-inventory migration the
+    /// bot approved the orderbook as the deposit spender; that spender is now the
+    /// inventory, so the orderbook allowance is dead capital-exposure surface
+    /// that would otherwise sit until the next (never-again) orderbook deposit
+    /// consumed it. A no-op when the allowance is already zero, so it is safe to
+    /// run on every startup. Returns `true` if a revoke tx was submitted.
+    pub async fn revoke_orderbook_allowance<Registry: IntoErrorRegistry>(
```
0xgleb, line 199: minor: returning `Result<bool>` where true means "revoke submitted" and false means "already zero" is the boolean-blindness pattern the guidelines warn about; the Ok(true)/Ok(false) match at the call site doesn't explain itself. A small `RevokeOutcome { Revoked, AlreadyZero }` enum would make it self-documenting.

crates/raindex/src/service.rs
```diff
@@ -357,6 +492,44 @@ impl<E: Evm> RaindexService<E> {
         Ok(self.evm.provider().get_block_number().await?)
     }
 
+    /// Startup preflight: verifies the signing wallet (`self.owner`) holds
+    /// `OPERATOR_ROLE` on the configured inventory, which every `deposit4` /
+    /// `withdraw4` requires. Returns [`RaindexError::MissingOperatorRole`] if
+    /// not, so the caller can fail fast rather than let the bot revert on the
+    /// first rebalance. `OPERATOR_ROLE` is read from the contract rather than
+    /// hardcoded, so a future role-scheme change stays in sync.
+    pub async fn verify_operator_role<Registry: IntoErrorRegistry>(
```
0xgleb, line 524: minor: this makes two independent eth_calls, so on a load-balanced RPC one node can serve OPERATOR_ROLE() while a lagging node serves hasRole() and returns a spurious false right after a grant, refusing startup at cutover time. Since keccak256("OPERATOR_ROLE") is a fixed constant you could hardcode it and drop the first call entirely, or add a small retry like the SCAN_ATTEMPTS pattern in find_recent_withdrawal.

crates/raindex/src/service.rs
```diff
@@ -757,6 +1000,318 @@ mod tests {
...
+            .unwrap();
+
+        assert_eq!(result, Some((expected_tx, op_amount)));
+    }
+
+    #[test]
+    fn map_withdraw_revert_decodes_insufficient_vault_liquidity() {
+        use alloy::hex;
+        use alloy::rpc::json_rpc::ErrorPayload;
+        use alloy::transports::TransportError;
```
0xgleb: nit: these function-body imports (and the SolCall one in push_operator_role_responses) belong at the mod tests header per the repo's import convention

crates/raindex/src/service.rs
```diff
@@ -276,56 +349,118 @@ impl<E: Evm> RaindexService<E> {
...
+                };
+
+                // A log that matches the filter's topic0/address set but fails
+                // to decode is skipped, not propagated: one malformed or
+                // unexpected log must not abort crash-recovery for the whole
+                // scan (which would turn a skippable log into a hard
+                // `ScanInconclusive` and block the resume). Under the current
+                // contracts this can't happen; the guard is defense against
+                // future ABI drift or an imposter log at the same address.
+                if let Some(topic) = log.topics().first() {
```
0xgleb: nit: the missing-tx-hash branch above warns, but a log with no topics, or an unrecognized topic0 falling through the else-if, is skipped silently here. Since provider drift is a named hazard in this scan, a warn in those cases would keep the defenses consistent.

crates/config/src/evm.rs
```diff
@@ -112,6 +135,44 @@ mod tests {
...
+
+        let error = result.unwrap_err();
+        assert!(
+            error.to_string().contains("vault_owner"),
+            "expected missing-field error for vault_owner, got: {error}"
+        );
+    }
+
+    #[test]
+    fn vault_owner_parses_when_present() {
```
0xgleb: nit: this round-trips a plain Address field through the derived Deserialize, which alloy already owns; the missing-field companion test carries the real coverage. Could be dropped, or repurposed to assert vault_owner actually flows through into EvmCtx.

src/conductor.rs
```diff
@@ -1307,14 +1392,13 @@ fn spawn_rebalancing_infrastructure<Chain: Wallet + Clone>(
...
 
-        let raindex_service = Arc::new(RaindexService::new(
-            base_wallet.clone(),
-            deps.ctx.evm.orderbook,
-            market_maker_wallet,
-        ));
+        let raindex_service =
+            build_rebalancing_raindex_service(&base_wallet, &deps.ctx, market_maker_wallet);
+
+        preflight_inventory_access(&raindex_service, &deps.ctx).await?;
```
0xgleb, line 1440: should fix: when inventory != orderbook this hard-fails Conductor startup in Rebalancing mode. Prod/staging already point at the real inventory but the bot wallet does not hold OPERATOR_ROLE yet, so merging and deploying blocks the service until ops grants the role. The PR says ships inert but this is a full startup failure, not just reverts on rebalance. Confirm deploy ordering with ops, or gate the preflight until cutover.

crates/raindex/src/lib.rs
```diff
@@ -43,6 +60,33 @@ pub enum RaindexError {
...
+    /// drained it first). Unlike the orderbook's `withdraw4`, which returns
+    /// `min(target, balance)`, the inventory settles atomically -- it reverts
+    /// rather than forward a short fill. Retryable once the vault is refunded,
+    /// and distinguishable so the caller can reconcile instead of blindly
+    /// re-submitting the same reverting withdraw.
+    #[error(
+        "inventory vault under-funded for {token}: requested {requested}, \
+         vault could cover {received}"
+    )]
+    InsufficientVaultLiquidity {
```
0xgleb, line 101: should fix: map_withdraw_revert surfaces this typed error, but usdc manager only wraps it as UsdcTransferError::Vault. Once inventory withdraws are live, an atomic InsufficientVaultLiquidity revert leaves the aggregate at WithdrawalSubmitting and apalis retries blindly. Worth matching this variant in withdraw_from_vault and failing or reconciling with context instead of a generic Vault wrap.

tests/e2e/full_system.rs
```diff
@@ -364,6 +364,8 @@ beneficiary_entity_name = "Simulate Failures"
 
 [raindex]
 orderbook = "{orderbook}"
+inventory = "{orderbook}"
```
0xgleb: minor: e2e collapses inventory to orderbook, so CI never hits the OPERATOR_ROLE preflight or distinct-inventory write path that prod uses. Combined with the pending Base fork noted in the PR, the prod-shaped config is outside automated verification. A test with inventory != orderbook would cover the actual deploy gate.

src/rebalancing/trigger/mod.rs
```diff
@@ -473,6 +473,9 @@ pub(crate) struct RebalancingService {
     config: RebalancingServiceConfig,
     vault_registry: Arc<Store<VaultRegistry>>,
     orderbook: Address,
+    /// On-chain owner of the Raindex vaults, used to key vault-registry lookups.
+    /// The inventory contract post-migration, the bot EOA before it; production
+    /// sources this from `Ctx::vault_owner`.
```
0xgleb: nit: this field is now sourced from Ctx::vault_owner() but still named order_owner, which is the opposite of Ctx::order_owner(). Easy footgun during the cutover. Consider renaming while you are here.

0xgleb (changes requested):

src/conductor.rs
JuaniRios: critical: with the prod and staging TOMLs included in this PR, this preflight fails the whole bot at boot, not just rebalancing. Both configs set inventory != orderbook so this skip doesn't apply, the bot's EOA doesn't hold OPERATOR_ROLE on 0x6b7B yet (hasRole returns false on Base as of today), and the error propagates through spawn_rebalancing_infrastructure -> PositionAndRebalancing::setup -> Conductor::run with bare ?s, so hedging and trading go down on every restart until the role is granted. Either grant the role before this deploys, or keep inventory == orderbook in the configs until it's granted so this skip branch applies, and flip the config as an explicit step with the RAI-1198 cutover.

src/inventory/polling.rs
```diff
@@ -251,7 +259,7 @@ where
     ) -> Result<Option<VaultRegistry>, InventoryPollingError<Exe::Error>> {
         let vault_registry_id = VaultRegistryId {
             orderbook: self.snapshot_id.orderbook,
-            owner: self.snapshot_id.owner,
+            owner: self.vault_owner,
```
JuaniRios, line 262: should fix: the vault_owner vs snapshot_id.owner split is the core of this PR, but every test in this module passes the same address for both, so a regression that reads snapshot_id.owner here would still pass the whole suite. One test with two distinct addresses (registry keyed by vault_owner, snapshot_id.owner different) asserting the balance reads and registry lookup key on vault_owner would pin the actual behavior.

crates/raindex/src/service.rs
```diff
@@ -379,6 +552,32 @@ impl<E: Evm> RaindexService<E> {
...
+/// Re-maps a raw `withdraw4` revert into [`RaindexError::InsufficientVaultLiquidity`]
+/// when the inventory reverted because the vault could not cover the request.
+///
+/// The inventory's `InsufficientVaultLiquidity` custom error is not in the
+/// OpenChain selector registry, so `submit` surfaces it as an opaque
+/// `EvmError::Contract` carrying the raw revert bytes. We decode those bytes
+/// against the compiled `IRaindexInventory` binding here so callers can branch
+/// on the under-funded case (reconcile / retry after refund) instead of seeing
+/// a generic contract error. Any other error is returned unchanged.
+fn map_withdraw_revert(err: EvmError) -> RaindexError {
```
JuaniRios, line 604: should fix: this only matches EvmError::Contract, but a real InsufficientVaultLiquidity revert likely never arrives as that variant. If the revert hits at gas estimation it comes back as Transport(RpcError), and if the tx mines and reverts, the replay path decodes through the selector registry and yields DecodedRevert(Unknown(bytes)) because this selector isn't in OpenChain. Worth extracting revert bytes from those two shapes too, and adding a test that drives an actual reverting withdraw4 through submit rather than hand-constructing the Contract variant.

crates/config/src/evm.rs
```diff
@@ -87,6 +108,8 @@ impl EvmCtx {
         Self {
             rpc_url: secrets.rpc,
             orderbook: config.orderbook,
+            inventory: config.inventory,
```
JuaniRios: should fix: nothing asserts these three same-typed Address fields land in the right slots. The new tests only exercise EvmConfig deserialization and never call EvmCtx::new. A test with three distinct addresses asserting orderbook, inventory, and vault_owner each map to their own field would guard exactly the swap this pass-through mapping invites.

src/conductor.rs
JuaniRios, line 1346: should fix: neither this function nor revoke_stale_orderbook_allowances has test coverage, and the e2e config collapses inventory == orderbook so the non-skip path never runs in CI. At minimum, a test that the skip branch performs no role check and one that a MissingOperatorRole failure propagates out of spawn_rebalancing_infrastructure would cover the new wiring.

crates/raindex/src/service.rs
```diff
@@ -276,56 +349,118 @@ impl<E: Evm> RaindexService<E> {
...
+                // unexpected log must not abort crash-recovery for the whole
+                // scan (which would turn a skippable log into a hard
+                // `ScanInconclusive` and block the resume). Under the current
+                // contracts this can't happen; the guard is defense against
+                // future ABI drift or an imposter log at the same address.
+                if let Some(topic) = log.topics().first() {
+                    if *topic == IRaindexInventory::OperatorWithdraw::SIGNATURE_HASH {
+                        let Ok(decoded) = log.log_decode::<IRaindexInventory::OperatorWithdraw>()
+                        else {
+                            warn!(target: "inventory", %tx_hash, "Skipping undecodable OperatorWithdraw log during resume scan");
```
JuaniRios: should fix: the old scan aborted on undecodable logs, this now warns and skips. The imposter-log justification can't actually happen since the filter is pinned to the two trusted contract addresses, so the only realistic decode failure is ABI drift, and in that case the skipped log could be our own in-flight withdrawal. The scan then returns None and the caller re-issues the withdraw, which is the double-spend this scan exists to prevent. Failing closed with an error seems safer than failing open here.

crates/config/src/evm.rs
```diff
@@ -38,6 +38,23 @@ impl std::fmt::Display for IngestionCutoff {
...
+    /// `vaultBalance2` read, vault-registry entry, and order-owner fill match is
+    /// scoped by. Required and explicit (no fallback): this parameter determines
+    /// fund-routing correctness, so a missing value must fail at startup rather
+    /// than silently assume an owner. Pre-migration (vaults owned by the bot
+    /// EOA) set it to the `[wallet]` address. Once the shared-inventory
+    /// migration moves the orders/vaults under the inventory contract,
+    /// `msg.sender` to Raindex is the inventory, so it becomes the vault owner
+    /// and this must be flipped to the `inventory` address in the same change
+    /// that grants the bot `OPERATOR_ROLE` and migrates vaults.
+    pub vault_owner: Address,
```
JuaniRios, line 117: minor: the PR description says vault_owner is optional and falls back to the wallet EOA, but the field is required and the tests assert a missing value fails to parse. The code is right per the no-silent-fallback rule, just worth correcting the description (and any runbook text copied from it) so nobody deploys expecting the fallback.

crates/raindex/src/service.rs
```diff
@@ -757,6 +1000,318 @@ mod tests {
...
+            .find_recent_withdrawal(USDC_BASE, TEST_VAULT_ID, from_block)
+            .await
+            .unwrap();
+
+        assert_eq!(result, Some((expected_tx, op_amount)));
+    }
+
+    #[test]
+    fn map_withdraw_revert_decodes_insufficient_vault_liquidity() {
+        use alloy::hex;
```
JuaniRios: minor: these function-body imports can move to the top of mod tests next to the existing alloy imports. AGENTS.md only allows function-level use for enum variants, cfg-gated code, and macro hygiene, and none apply here since the whole test module is already cfg(test)-gated.

crates/config/src/evm.rs
```diff
@@ -38,6 +38,23 @@ impl std::fmt::Display for IngestionCutoff {
 #[serde(deny_unknown_fields)]
 pub struct EvmConfig {
     pub orderbook: Address,
+    /// Shared `RaindexInventory` (raindex.governance) that owns the Raindex
+    /// vaults. All venue adapters (Bebop hook, univ4 hook) and this bot's own
+    /// rebalancing path settle through it -- `deposit4`/`withdraw4` on this
+    /// address instead of on the orderbook. Fill events on the pooled vaults
+    /// are also surfaced here as `OperatorDeposit`/`OperatorWithdraw`.
+    pub inventory: Address,
```
JuaniRios: minor: SPEC.md still describes rebalancing as depositing straight into orderbook vaults and doesn't mention the inventory, OPERATOR_ROLE, or the vault_owner cutover. Since SPEC.md is the north star doc and this changes how capital moves, it should get a section on the shared-inventory routing. README is also missing the new flake input and required config fields.

crates/raindex/src/service.rs
JuaniRios: nit: this if-let plus if/else-if chain sits four levels deep inside the two loops. A let-else continue on the topic followed by a match on *topic flattens it a level, which is the shape the repo style guide prefers.

src/conductor.rs
```diff
@@ -1307,14 +1392,13 @@ fn spawn_rebalancing_infrastructure<Chain: Wallet + Clone>(
         let vault_lookup: Arc<dyn VaultLookup> = Arc::new(VaultRegistryLookup::new(
             deps.vault_registry_projection,
             deps.ctx.evm.orderbook,
-            market_maker_wallet,
+            deps.ctx.vault_owner(),
```
JuaniRios: minor: VaultRegistryLookup::new still takes orderbook and owner as adjacent positional Addresses, and this call now mixes a contract address with vault_owner(). Given the PR added RaindexContracts specifically to kill this class of silent swap, the same named-struct treatment here would be consistent.

src/conductor.rs
```diff
@@ -1359,7 +1443,7 @@ fn spawn_rebalancing_infrastructure<Chain: Wallet + Clone>(
             },
             deps.vault_registry,
             deps.ctx.evm.orderbook,
-            market_maker_wallet,
+            deps.ctx.vault_owner(),
```
JuaniRios: minor: same swap-risk shape here, RebalancingService::new declares orderbook followed by order_owner as bare positional Addresses. Bundling them into a named-field struct like RaindexContracts would make the call sites self-checking.

src/conductor.rs
JuaniRios: nit: the intermediate Vec isn't needed, iter::once(USDC_BASE).chain(ctx.assets.equities.symbols.values().map(...)) works directly in the for loop.

JuaniRios (changes requested):

crates/raindex/src/service.rs
hardyjosh: Fixed. Rewrote as `find_recent_withdrawal_prefers_newer_operator_withdraw_on_same_vault`: the stale WithdrawV2 now uses `TEST_VAULT_ID` with the matching sender and an older block, so both events pass their branch filter and only newest-first ordering prevents adopting the stale amount -- oldest-first would now fail the test.

crates/raindex/src/service.rs
hardyjosh: Fixed. Added both same-(token,vaultId) cases differing only by block: `..._prefers_newer_operator_withdraw_on_same_vault` (OperatorWithdraw newer) and `..._prefers_newer_withdrawv2_on_same_vault` (WithdrawV2 newer), each asserting the correct amount is adopted -- so an ordering or else-if regression fails.

crates/raindex/src/service.rs
hardyjosh: Fixed. Applied `.map_err(map_withdraw_revert)` to `confirm_tx_receipt` too, so the equity-redemption submit_withdraw + confirm path surfaces the typed `InsufficientVaultLiquidity`. As you noted, non-matching errors pass through unchanged, so it's safe for deposit confirmations.

src/conductor.rs
hardyjosh: The role-check and revoke are now unit-tested at the raindex layer (`verify_operator_role` pass/fail/retry, `revoke_orderbook_allowance` no-op/revoke/idempotency). The collapsed early-return is exercised by every existing conductor/e2e fixture (all collapse inventory==orderbook and start cleanly -- if the gate didn't skip, they'd hit hasRole on the orderbook and fail). Orchestration/e2e coverage of the distinct-inventory path needs a deployed RaindexInventory fixture on Anvil; filed as [RAI-1248](https://linear.app/makeitrain/issue/RAI-1248).

src/conductor.rs
hardyjosh: Agreed the address-equality is doing a mode discriminant's job. For this PR we've made the config ship genuinely inert (inventory == orderbook, explicitly commented), so the misconfig you describe is the intended inert state rather than a silent bypass. The explicit `InventoryMode` enum is the right end state -- filed as [RAI-1249](https://linear.app/makeitrain/issue/RAI-1249) rather than growing this PR (+ PR2's fill watcher).

src/conductor.rs
hardyjosh: Fixed -- `std::iter::once(USDC_BASE).chain(...)`.

crates/raindex/src/service.rs
hardyjosh: Fixed -- added `RevokeOutcome { Revoked, AlreadyZero }`; the call site matches on it and the tests assert on it.

crates/raindex/src/service.rs
hardyjosh: Fixed -- `hasRole` is now retried (`ROLE_CHECK_ATTEMPTS`), so a lagging load-balanced node serving a stale `false` right after a grant doesn't refuse startup at cutover; a single `true` wins immediately. Kept reading `OPERATOR_ROLE` from the contract (stays in sync with a future role-scheme change). Added `verify_operator_role_retries_past_a_lagging_false`.

crates/raindex/src/service.rs
hardyjosh: Fixed -- moved `hex`, `ErrorPayload`, `TransportError`, and `SolCall` to the `mod tests` header.

crates/raindex/src/service.rs
hardyjosh: Fixed -- a log with no topics and a log with an unrecognized topic0 both `warn!` before skipping now, consistent with the missing-tx-hash and decode-failure branches.

crates/config/src/evm.rs
hardyjosh: Fixed -- repurposed to `vault_owner_flows_from_config_into_ctx`, which builds an `EvmCtx` via `EvmCtx::new` and asserts vault_owner threads through and is distinct from both orderbook and inventory. (The missing-field test carries the required-ness coverage.)

src/conductor.rs
hardyjosh: Good catch -- you're right this is a full startup failure, not inert. Resolved by making the config genuinely inert: prod/staging now set `inventory = orderbook` (preflight skipped, writes stay on the orderbook = current behaviour). The RAI-1198 cutover is one config change flipping both `inventory` and `vault_owner` to the shared inventory, deployed with the OPERATOR_ROLE grant. Updated the config comments and the PR body.

crates/raindex/src/lib.rs
hardyjosh: Fixed -- `withdraw_from_vault` and `resume_withdrawal_submitting` route `withdraw_usdc` failures through `handle_vault_withdraw_error`, which on `InsufficientVaultLiquidity` emits `FailWithdrawal` (WithdrawalSubmitting -> WithdrawalFailed, so apalis stops blindly retrying the reverting withdraw) and returns terminal `UsdcTransferError::VaultUnderFunded` for reconciliation. Other errors keep the generic `Vault` wrap and retry semantics.

tests/e2e/full_system.rs
hardyjosh: The distinct-inventory e2e needs a deployed RaindexInventory fixture on Anvil; filed with the orchestration preflight test as [RAI-1248](https://linear.app/makeitrain/issue/RAI-1248).

src/rebalancing/trigger/mod.rs
hardyjosh: Fixed -- renamed the `RebalancingService` field and constructor param to `vault_owner`, with a doc note that it's the vault owner, not the signing wallet.

crates/raindex/src/service.rs
JuaniRios: Reworked to `find_recent_withdrawal_prefers_newer_operator_withdraw_over_older_withdrawv2_same_vault`: the stale WithdrawV2 now uses `TEST_VAULT_ID` with a matching sender and an older block, so only newest-first ordering (not the vault filter) can prevent adopting it. Addressed in 3a06081c.

crates/raindex/src/service.rs
JuaniRios: Added same-vault cutover tests in both directions (`OperatorWithdraw` newer vs `WithdrawV2` newer, differing only by block), plus a fail-closed undecodable-log test. 3a06081c.

crates/raindex/src/service.rs
JuaniRios: Applied `.map_err(map_withdraw_revert)` to `confirm_tx_receipt`. 3a06081c.

src/conductor.rs
JuaniRios: Added `preflight_skips_role_check_in_legacy_mode` and `preflight_attempts_role_check_in_managed_mode` at the conductor layer (mocked wallet-backed `RaindexService`); the role verification itself is covered by the service-level `verify_operator_role` tests. 3a06081c.

src/conductor.rs
JuaniRios: Implemented the enum rather than deferring it: `InventoryMode::Legacy | Managed { inventory }` on `EvmCtx`, driven by an explicit `inventory_mode` config field (managed requires `inventory`, legacy forbids it, validated at startup). The address-equality check is gone. 3a06081c.

src/conductor.rs
JuaniRios: Done -- `iter::once(USDC_BASE).chain(...)` directly in the loop. 3a06081c.

crates/raindex/src/service.rs
JuaniRios: Added `RevokeOutcome { Revoked, AlreadyZero }`. 3a06081c.

crates/raindex/src/service.rs
JuaniRios: `verify_operator_role` now reads the immutable `OPERATOR_ROLE` constant once (node-agnostic) and retries `hasRole` across `SCAN_ATTEMPTS` with backoff, so a node lagging a fresh cutover grant no longer refuses startup. 3a06081c.

crates/raindex/src/service.rs
JuaniRios: Moved to the `mod tests` header (the `SolCall` import, and the `map_withdraw_revert` test's `hex`/`ErrorPayload`/`TransportError`). 3a06081c.

crates/raindex/src/service.rs
JuaniRios: Went further than a warn: the scan now **fails closed** (`RaindexError::ScanAnomalousLog`) on any anomalous log from a trusted address -- no topics, unrecognized topic0, or undecodable. Since the filter is pinned to trusted contracts, an undecodable log means ABI drift and could be our own in-flight withdrawal; skipping it would report a false absence and re-issue an irreversible withdraw, so failing closed is the safer default. 3a06081c.

crates/config/src/evm.rs
JuaniRios: Repurposed into `evm_ctx_new_maps_config_fields_to_their_own_slots`, which drives three distinct addresses through `EvmCtx::new` and asserts each lands in its own slot, plus Legacy/Managed validation tests. 3a06081c.

src/conductor.rs
JuaniRios: Confirmed and handled: prod/staging now ship `inventory_mode = "legacy"` until the RAI-1198 cutover, so the preflight is skipped and the bot boots. The TOML comments spell out the cutover sequence (flip to `managed` + set `inventory` + flip `vault_owner` in the change that grants the role). 3a06081c.

crates/raindex/src/lib.rs
JuaniRios: The USDC manager now matches `InsufficientVaultLiquidity` and surfaces a distinct terminal `UsdcTransferError::InsufficientVaultLiquidity` (latched for operator reconciliation, no blind redrive) instead of the opaque `Vault` wrap. `map_withdraw_revert` was also broadened to decode the revert from the `Transport` and `DecodedRevert(Unknown)` shapes, not just `Contract`. 3a06081c.

tests/e2e/full_system.rs
JuaniRios: Agreed -- a true distinct-inventory e2e needs a deployed `RaindexInventory` with AccessControl (the pending Base fork), which the anvil harness doesn't have. Covered the managed path at every tractable layer (conductor preflight, config resolution, service `verify_operator_role`, polling); the full e2e/fork coverage is tracked in RAI-1248. 3a06081c.

src/rebalancing/trigger/mod.rs
JuaniRios: Rather than rename: the `order_owner` field is gone. The `(orderbook, vault_owner)` pair is now bundled into the existing `VaultRegistryId` named struct at both `RebalancingService::new` and `VaultRegistryLookup::new`, so the swap risk the field name warned about is eliminated at the type level. 3a06081c.

crates/raindex/src/service.rs
JuaniRios: Addressed. The stale `WithdrawV2` now uses `TEST_VAULT_ID` with a matching sender and an older block, so only the newest-first ordering can prevent adopting the stale amount — see `find_recent_withdrawal_prefers_newer_operator_withdraw_over_older_withdrawv2_same_vault` in `crates/raindex/src/service.rs`. A regression to oldest-first now fails the test.

crates/raindex/src/service.rs
JuaniRios: Addressed. Both cutover orderings are now covered on the same `(token, vaultId, owner)`, differing only by block: `find_recent_withdrawal_prefers_newer_operator_withdraw_over_older_withdrawv2_same_vault` and `find_recent_withdrawal_prefers_newer_withdrawv2_over_older_operator_withdraw_same_vault`.

crates/raindex/src/service.rs
JuaniRios: Addressed. `confirm_tx_receipt` now applies `.map_err(map_withdraw_revert)` too, so a concurrent drain that reverts at inclusion surfaces as the typed `InsufficientVaultLiquidity` on the equity-redemption path. Non-matching errors still pass through unchanged.

crates/raindex/src/service.rs
JuaniRios: Addressed. `revoke_orderbook_allowance` now returns `RevokeOutcome { Revoked, AlreadyZero }` (defined in `crates/raindex/src/lib.rs`) instead of a bare `bool`.

crates/raindex/src/service.rs
JuaniRios: Addressed. `verify_operator_role` now retries the `hasRole` read with the `SCAN_ATTEMPTS` / `SCAN_RETRY_BACKOFF` pattern and warns "OPERATOR_ROLE not yet visible; retrying (RPC node may lag a recent grant)", so a lagging node right after the cutover grant no longer refuses startup.

crates/raindex/src/service.rs
JuaniRios: Addressed. All test imports, including `SolCall`, now live at the `mod tests` header; no function-body `use` remains in `service.rs`.

crates/raindex/src/service.rs
JuaniRios: Addressed, and taken further than warning: a log with no topics or an unrecognized `topic0` now warns *and* fails closed with `RaindexError::ScanAnomalousLog` rather than being skipped, since a silently skipped log could be our own in-flight withdrawal.

crates/raindex/src/lib.rs
JuaniRios: Addressed. `classify_vault_withdrawal_error` in `src/rebalancing/usdc/manager.rs` now maps `RaindexError::InsufficientVaultLiquidity` onto a dedicated `UsdcTransferError::InsufficientVaultLiquidity` variant instead of the generic `Vault` wrap, wired into both withdrawal submit paths and covered by tests.

tests/e2e/full_system.rs
JuaniRios: Acknowledged, leaving as-is. A managed-mode e2e needs `OPERATOR_ROLE` granted on a Base fork, which is the pending-fork dependency this PR can't land ahead of. In the meantime the mode branch, the role check and its error context are unit-covered by `preflight_skips_role_check_in_legacy_mode` and `preflight_attempts_role_check_in_managed_mode` in `src/conductor.rs`. Tracking the real managed-mode e2e with the RAI-1198 cutover.

src/rebalancing/trigger/mod.rs
JuaniRios: Addressed. The field is now `registry_id: VaultRegistryId`, so the owner lives inside a named struct rather than a bare `order_owner: Address` that meant the opposite of `Ctx::order_owner()`.

src/conductor.rs
JuaniRios: Addressed. `preflight_skips_role_check_in_legacy_mode` (empty `Asserter`, proving no eth_call is issued) and `preflight_attempts_role_check_in_managed_mode` (asserting the error carries the preflight context) now cover both branches with distinct inventory and orderbook addresses.

src/conductor.rs
JuaniRios: Addressed in this PR rather than deferred. `InventoryMode { Legacy, Managed { inventory } }` now lives in `crates/config/src/evm.rs`, and the two contradictory configs are rejected at startup by `EvmConfigError::{ManagedWithoutInventory, LegacyWithInventory}`. Address equality no longer discriminates the mode.

src/conductor.rs
JuaniRios: Addressed. Now `std::iter::once(st0x_evm::USDC_BASE).chain(ctx.assets.equities.symbols.values().map(|equity| equity.tokenized_equity_derivative))`, consumed directly by the loop.

src/conductor.rs
JuaniRios: Good catch, and fixed. Both prod and staging now ship `inventory_mode = "legacy"`, which skips the preflight entirely, so this stack is inert on deploy. Flipping to `"managed"` is an explicit RAI-1198 cutover step done in the same change that grants `OPERATOR_ROLE`.

src/conductor.rs
JuaniRios: Resolved by the `InventoryMode` enum: prod and staging now set `inventory_mode = "legacy"` with no `inventory` address, so `preflight_inventory_access` early-returns and boot cannot fail before the role is granted. The config flip is now an explicit cutover step rather than an implicit consequence of merging.

src/inventory/polling.rs
JuaniRios: Addressed. `poll_onchain_keys_vault_registry_on_vault_owner_not_snapshot_owner` in `src/inventory/polling.rs` now seeds distinct `vault_owner` and `snapshot_id.owner`, keys the registry only under `vault_owner`, and asserts the balance reads and registry lookup follow `vault_owner`. Reading `snapshot_id.owner` here now fails.

crates/raindex/src/service.rs
JuaniRios: Addressed. `map_withdraw_revert` now normalizes through `err.revert_data()`, which covers all three shapes (`Contract`, JSON-RPC `Transport`, and `DecodedRevert(Unknown(..))`). Three dedicated tests pin each shape plus a pass-through case. Driving a real reverting `withdraw4` isn't possible in this harness — the test service points inventory at the orderbook, whose `withdraw4` clamps to the balance rather than reverting.

crates/config/src/evm.rs
JuaniRios: Addressed. `evm_ctx_new_maps_config_fields_to_their_own_slots` in `crates/config/src/evm.rs` calls `EvmCtx::new` with three distinct addresses and asserts each of `orderbook`, `inventory` and `vault_owner` lands in its own field.

src/conductor.rs
JuaniRios: Addressed. Both branches are now covered: the legacy skip issues no role call (empty `Asserter`), and a missing role propagates out as an `Err` carrying the preflight context. `preflight_inventory_access` is invoked with a bare `?` inside `spawn_rebalancing_infrastructure`, so that propagation is exercised.

crates/raindex/src/service.rs
JuaniRios: Agreed, and changed to fail closed. Undecodable logs now return `RaindexError::ScanAnomalousLog` instead of being skipped, with `find_recent_withdrawal_fails_closed_on_undecodable_log` asserting the exact variant. Failing open here risked re-issuing a withdrawal this scan exists to prevent.

crates/config/src/evm.rs
JuaniRios: Addressed — the description now states `vault_owner` is required and that `inventory` is "validated at startup, no fallback", matching what the code and tests enforce. No runbook text depends on the old wording.

crates/raindex/src/service.rs
JuaniRios: Addressed. All test imports now sit at the `mod tests` header; no function-body `use` remains.

crates/config/src/evm.rs
JuaniRios: Addressed. SPEC.md now carries a "Shared-Inventory Settlement" section covering the legacy/managed modes, the `OPERATOR_ROLE` preflight and the `vault_owner` cutover, and README documents the new required config fields and the `raindex-governance.nix` flake input.

crates/raindex/src/service.rs
JuaniRios: Addressed. The chain is now flattened with `let-else` guards on `transaction_hash` and `topics().first()`, dropping a nesting level. Kept `if/else-if` on the topic rather than a `match` because `SIGNATURE_HASH` are associated consts, not structural patterns.

src/conductor.rs
JuaniRios: Addressed. `VaultRegistryLookup::new` now takes a named `VaultRegistryId { orderbook, owner }` instead of two adjacent positional `Address` arguments.

src/conductor.rs
JuaniRios: Addressed. `RebalancingService::new` now takes the same named `VaultRegistryId` struct, so the call sites are self-checking.

src/conductor.rs
JuaniRios: Addressed. The intermediate `Vec` is gone; `once(USDC_BASE).chain(...)` feeds the `for` loop directly.

crates/config/src/evm.rs
JuaniRios: Addressed. The redundant round-trip test is gone; its coverage was repurposed into `evm_ctx_new_maps_config_fields_to_their_own_slots`, which asserts `vault_owner` actually flows through `EvmCtx::new`. The missing-field companion test remains.

config/staging/st0x-hedge.toml
```diff
@@ -32,6 +32,24 @@ beneficiary_entity_name = "T0 TRADE (BVI) LTD"
...
+# give staging its own inventory redeploy; at that cutover switch this to
+# "managed", set `inventory` to the staging inventory, and flip `vault_owner`
+# below to that same address so a staging bug can't touch prod-owned vaults.
+inventory_mode = "legacy"
+# REQUIRED. Owner of the Raindex orders/vaults on-chain (vaultBalance2 /
+# registry / fill match key). Currently the bot's Turnkey wallet
+# (= [wallet].address below) because the vaults are bot-EOA-owned
+# pre-migration; flip to the `inventory` address once RAI-1198 migrates the
+# vaults under it.
+vault_owner = "0xA9C16673F65AE808688cB18952AFE3d9658C808f"
```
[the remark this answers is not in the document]
JuaniRios: Good catch — real bug, fixed. Staging's `vault_owner` was prod's wallet address; it now correctly reads `0x16ca08a5825612aAe805D172a81B1a52b43574bf` (staging's `[wallet].address`). Deployed as written, staging would have read prod's vault balances and matched prod's fills as its own. Added `repo_config_vault_owner_matches_settlement_mode` in `crates/config/src/loader.rs`, which asserts the invariant across every checked-in config: `legacy` implies `vault_owner == [wallet].address`, `managed` implies `vault_owner == inventory`. Verified it fails against the old value.

example.config.toml
```diff
@@ -29,6 +29,28 @@ beneficiary_entity_name = "Acme Corp"
...
+# and refuses to start otherwise).
+inventory = "0x6b7B523fADd1677413AD92c9404C8F0796bACF6F"
+# REQUIRED. Owner of the Raindex orders/vaults on-chain -- scopes every
+# vaultBalance2 read, vault-registry entry, and ClearV3/TakeOrderV3 order-owner
+# fill match. No fallback: this determines fund routing, so it must be explicit
+# and a missing value fails startup. While the vaults are owned by the bot EOA
+# (the pre-inventory-migration state), set it to the [wallet].address value.
+# Once the vaults are migrated under the inventory contract, set it to the
+# `inventory` address so balance reads and fill matching query the right owner.
+vault_owner = "0xfcad0b19bb29d4674531d6f115237e16afce377c"
```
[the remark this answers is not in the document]
JuaniRios: Correct, fixed. `example.config.toml` shows `managed` mode, so `vault_owner` now equals the `inventory` address and the comment spells out which value each mode takes. The new `repo_config_vault_owner_matches_settlement_mode` test enforces this for every checked-in config, so the template can't drift back.

JuaniRios (approved):

src/rebalancing/usdc/manager.rs
```diff
@@ -143,6 +143,26 @@ impl<EthWallet: Wallet, BaseWallet: Wallet> UsdcBridgeHelper for CctpBridge<EthW
...
 }
 
+/// Classifies a failed vault `withdraw_usdc`. An atomic
+/// [`RaindexError::InsufficientVaultLiquidity`] revert means the vault could not
+/// cover the request; it withdrew nothing (atomic revert), so retrying only
+/// reverts again until the vault is refunded. It is surfaced as a distinct,
+/// contextful error the job latches for operator reconciliation (no auto-retry)
+/// rather than the opaque `Vault` wrap, which the job redrives. Any other error
+/// keeps the opaque wrap and the normal redrive path.
+fn classify_vault_withdrawal_error(error: RaindexError) -> UsdcTransferError {
```
0xgleb, line 153: **The documented semantics of this classification don't exist at the job layer.** This doc says the job "latches for operator reconciliation (no auto-retry)", and the variant's docs/`#[error]` text in `src/rebalancing/usdc/mod.rs` say the withdrawal aggregate "is failed for operator reconciliation before this surfaces". Neither happens:

- `job.rs` has no arm for `UsdcTransferError::InsufficientVaultLiquidity` — it falls into the generic terminal catch-all (`Err(error) =>`), which returns `Err` to apalis, so the job is retried up to the retry budget (that arm's own comment says up to 4 attempts).
- Each retry re-enters `resume_withdrawal_submitting`; the finality-gated scan finds nothing (the revert emitted no event), so it falls through and re-issues the reverting `withdraw4` on-chain.
- Nothing sends `FailWithdrawal` on this path, so the aggregate stays at `WithdrawalSubmitting` — not failed.

Net effect: the variant behaves identically to the opaque `Vault` wrap it exists to be distinct from, and the operator alert/error text misdescribe the aggregate state. Either add a job arm that actually latches (Ok-return + single alert, like `BurnTxDropped`) and/or fails the aggregate, or correct the three doc/log/error texts to match reality. Safety isn't compromised (atomic revert, bounded retries, alert fires) — the issue is the operator-facing contract being wrong.

src/rebalancing/usdc/mod.rs
```diff
@@ -58,6 +58,19 @@ pub(crate) enum UsdcTransferError {
...
+    /// The shared inventory reverted a `withdraw4` because the vault could not
+    /// cover the requested amount (a concurrent clear drained it). Distinct from
+    /// the opaque `Vault` wrap so it is not redriven blindly: retrying the same
+    /// withdraw reverts again until the vault is refunded, so the withdrawal
+    /// aggregate is failed for operator reconciliation before this surfaces.
+    #[error(
+        "inventory vault under-funded on withdraw: {source}; \
+         failed for operator reconciliation"
+    )]
+    InsufficientVaultLiquidity {
```
0xgleb, line 71: Two things here (see the main comment on `classify_vault_withdrawal_error` for context):

1. "the withdrawal aggregate is failed for operator reconciliation before this surfaces" (doc + `#[error]` text) is not true today — nothing sends `FailWithdrawal` on this path.
2. Minor: typing the field as the whole `RaindexError` permits invalid states — nothing prevents constructing this variant around, say, `ZeroAmount`; only the classifier's match guarantees the invariant. Carrying `token`/`requested`/`received` directly (as `RaindexError::InsufficientVaultLiquidity` does) would make the invalid state unrepresentable.

crates/config/src/loader.rs
```diff
@@ -3849,6 +3980,45 @@ mod tests {
...
+
+            // e2e/config.toml supplies its wallet out of band, so there is no
+            // [wallet].address to pin a legacy vault_owner against.
+            let Some(wallet_value) = config.wallet else {
+                continue;
+            };
+            let wallet: WalletMeta = wallet_value.try_into().unwrap();
+
+            match (config.raindex.inventory_mode, config.raindex.inventory) {
+                (InventoryModeTag::Legacy, _) => assert_eq!(
```
0xgleb: Minor: this Legacy arm wildcards `inventory`, and `all_repo_config_tomls_are_valid` only deserializes (never calls `EvmCtx::new`), so a checked-in `inventory_mode = "legacy"` config still carrying an `inventory` address passes CI and only fails at the deploy-time `validate-config` gate. Asserting `inventory.is_none()` here would catch the contradiction in CI, matching the fail-fast intent of `LegacyWithInventory`.

crates/raindex/src/lib.rs
```diff
@@ -43,6 +72,45 @@ pub enum RaindexError {
...
+    /// failed to ABI-decode. Impossible under the current contracts, so it means
+    /// ABI drift or an imposter at a trusted address. The log could be our own
+    /// in-flight withdrawal, so the scan fails closed rather than skip it and let
+    /// the caller re-issue an irreversible, already-submitted withdraw. Not
+    /// auto-retryable -- ABI drift needs a code fix, not another scan.
+    #[error(
+        "withdrawal scan hit an anomalous log ({reason}); failing closed to \
+         avoid re-issuing an already-submitted withdraw"
+    )]
+    ScanAnomalousLog { reason: &'static str },
```
0xgleb: Minor: the undecodable-log paths do `.inspect_err(...).map_err(|_| ScanAnomalousLog { reason: "..." })`, discarding the alloy decode error instead of carrying it via `#[source]`, and `reason: &'static str` is a stringly discriminator. A small reason enum + `#[source]` on the decode arm would preserve the error chain per the repo error-design rule.

crates/raindex/src/service.rs
```diff
@@ -357,6 +508,64 @@ impl<E: Evm> RaindexService<E> {
...
+                return Ok(());
+            }
+
+            if attempt < SCAN_ATTEMPTS {
+                warn!(
+                    target: "inventory",
+                    inventory = %self.inventory_address,
+                    wallet = %self.owner,
+                    attempt,
+                    "OPERATOR_ROLE not yet visible; retrying (RPC node may lag a recent grant)",
```
0xgleb, line 557: Minor: the thread reply mentioned a test `verify_operator_role_retries_past_a_lagging_false`, but it doesn't exist — the all-false failure path is covered, and this retry loop is implemented, but no test asserts that a `true` after intermediate `false`s succeeds. Worth adding since that's the exact scenario this loop exists for.

0xgleb (changes requested):

0xgleb (approved):

PR #970, opened 2026-07-01
feat: hedge cross-venue fills via RaindexInventory events

hardyjosh:
## What

Closes [RAI-1197](https://linear.app/makeitrain/issue/RAI-1197/liquidity-bot-hedge-on-cross-venue-fills-bebop-univ4-trade-events) and [RAI-684](https://linear.app/makeitrain/issue/RAI-684/extend-liquidity-bot-countertrading-to-bebop) (folded in — the venue-agnostic version covers Bebop).

Adds a third fill source to the `st0x.liquidity` fill watcher: settlements routed through the shared **`RaindexInventory`** on behalf of any venue adapter (Bebop hook, univ4 hook, or any future venue holding `OPERATOR_ROLE`) are now hedged alongside direct Raindex clears.

Stacked on [#969](https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/969).

## Why

- One pool, many consumers: Bebop RFQs and univ4 swaps against the same shared vaults are settled by draws on the inventory contract (`OperatorDeposit` + `OperatorWithdraw` per settlement). Before this PR the bot only hedged direct Raindex clears (`ClearV3`/`TakeOrderV3`), so every adapter fill flowed through the shared pool unhedged — a growing directional exposure.
- Venue-agnostic by construction: the vault deltas ARE the trade regardless of which adapter routed the settlement, so watching the single inventory contract covers every future venue for free (supersedes the Bebop-only RAI-684).

## How

- **Event model:** new `RaindexTradeEvent::InventoryTrade` variant carrying the paired `(OperatorDeposit, OperatorWithdraw)`. Direction/price/amount derive through the same `TradeDetails::try_from_io` as the ClearV3 path (USDC side prices, wrapped-equity side drives quantity). `TradeDetails` now carries the resolved `equity_symbol` + `equity_token`, so both parsers stop re-deriving the equity side from a raw `"USDC"` string compare.
- **Backfill:** a third `eth_getLogs` filter (address = `evm_ctx.inventory_address()`, topic0 ∈ `{OperatorDeposit, OperatorWithdraw}`). Deposits are bucketed by `tx_hash`; each withdraw pairs with its same-tx deposit into one `AccountForDexTrade`.
- **Skip the bot's own settlements:** since #969 the bot's rebalancing calls `deposit4`/`withdraw4` on the same inventory, emitting `Operator*` events with `operator == order_owner`. Those are filtered out before pairing (threaded from `BackfillRange::perform`), so routine rebalances don't fire the unpaired-withdraw path or create phantom hedges. A genuinely unpaired withdraw (or a second deposit in one tx) is logged.
- **Parser** (`try_from_inventory_trade`): resolves symbols/decimals, and now matches the ClearV3 sibling exactly — a non-positive price on a non-zero-equity fill now returns `Err(NonPositivePrice)` — recorded in `skipped_fills` rather than silently dropped as `Ok(None)` — and the block timestamp goes through `resolve_block_timestamp` (header fallback) so backfilled fills don't persist null timestamps.
- **Trade accountant:** `AccountantCtx` gains the `inventory` address; `reconstruct_log` selects the emitting contract per event type. Downstream Alpaca hedging is untouched.
- **Rebase note:** #969 turned `EvmCtx.inventory` into an `InventoryMode` enum, so the three read sites here now use `inventory_address()`.

## Testing

- `cargo nextest run -p st0x-hedge --all-features` (incl. e2e): all pass. `cargo clippy --workspace --all-targets --all-features` and `cargo fmt --check`: clean.
- New tests: parser happy-path + zero-equity + non-positive-price + USDC-on-deposit direction-flip; backfill pairing (real pair → one job, unpaired → skipped, bot-operator pair → skipped, interleaved ordering); accountant `InventoryTrade` perform arm; `reconstruct_log` address/round-trip.

## Anything else

- **Prod fill txs to sanity-check against once live:**
  - Bebop: `0xe13a11de734768f08a9c1ef66e8de3bcb9072f8cdabce9f1d819e1ae9909d4b9` (~0.0342 wtCOIN for 5 USDC via BebopRouter)
  - Univ4: `0x9ee8e401a6f12227df1a30a236b60ac83c72b2b1eb610d83cf292ae789eb0805` (0.01 wtCOIN → 2 USDC on the v4 pool)
- **On rollout, rewind the backfill checkpoint** to the inventory deploy block (or run a one-shot backfill) so adapter fills before this deploy get ingested — the persisted checkpoint predates inventory ingestion (noted at the checkpoint site).
- **Attribution vs signal:** the `operator` topic identifies which venue routed a settlement, but the hedge signal is `(token, vaultId, net Δ)` regardless of operator — per-venue metrics are easy to add later.
- Same ops dependency as [#969](https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/969): [RAI-1198](https://linear.app/makeitrain/issue/RAI-1198/redeploy-raindexinventory-owned-by-turnkey-migrate-vaults-regrant) supplies the final inventory address for config.

## Summary by CodeRabbit

* **New Features**
  * Added support for inventory-settled trades in managed inventory mode by pairing deposit/withdraw settlement legs and processing them through the existing trade pipeline.
  * Enabled inventory-aware on-chain recovery/backfill and added an ABI binding for inventory event decoding.

* **Bug Fixes**
  * Prevent inventory trade events from triggering vault registry discovery.
  * Quarantines/skips ambiguous, reorged, bot-origin, invalid, or non-introspectable inventory settlements; records skipped fills.

* **Documentation / Metrics / Tests**
  * Updated the event-processing specification and expanded metrics for inventory ambiguity/unpaired settlements.
  * Expanded integration and unit test coverage, including managed inventory e2e hedging.

### DIFF
--- a/src/onchain/backfill.rs
+++ b/src/onchain/backfill.rs
@@ -1,12 +1,15 @@
 //! Historical onchain event backfill with retry logic.
 //!
-//! Scans past blocks for `ClearV3` and `TakeOrderV3` events and pushes them
-//! into the apalis job queue for processing, ensuring no trades are missed
-//! after downtime. A persisted checkpoint records the last block that was
-//! fully enqueued.
+//! Scans past blocks for `ClearV3` and `TakeOrderV3` events on the OrderBook,
+//! plus `OperatorDeposit`/`OperatorWithdraw` events on the shared
+//! `RaindexInventory` (paired into `InventoryTrade`s), and pushes them into the
+//! apalis job queue for processing, ensuring no trades are missed after
+//! downtime. A persisted checkpoint records the last block that was fully
+//! enqueued.
 
+use alloy::primitives::TxHash;
 use alloy::providers::Provider;
-use alloy::rpc::types::Filter;
+use alloy::rpc::types::{Filter, Log};
 use alloy::sol_types::SolEvent;
 use backon::{BackoffBuilder, ExponentialBuilder, Retryable};
 use futures_util::future;
@@ -14,17 +17,20 @@ use itertools::Itertools;
 use metrics::counter;
 use serde::{Deserialize, Serialize};
 use sqlx::SqlitePool;
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
 use std::time::Duration;
-use tracing::{debug, info, trace, warn};
+use tracing::{debug, error, info, trace, warn};
 
 use st0x_config::EvmCtx;
 use st0x_evm::Evm;
 use st0x_execution::Executor;
 
 use super::OnChainError;
+use crate::bindings::IRaindexInventory::{OperatorDeposit, OperatorWithdraw};
 use crate::bindings::IRaindexV6::{ClearV3, TakeOrderV3};
 use crate::conductor::job::{Job, Label};
-use crate::onchain::trade::RaindexTradeEvent;
+use crate::onchain::trade::{BotOperator, InventoryTrade, RaindexTradeEvent};
 use crate::trading::onchain::inclusion::EmittedOnChain;
 use crate::trading::onchain::trade_accountant::{
     AccountForDexTrade, AccountantCtx, DexTradeAccountingJobQueue, TradeAccountingError,
@@ -57,6 +63,7 @@ pub(crate) fn get_backfill_retry_strat() -> ExponentialBuilder {
 pub(crate) async fn backfill_events<P: Provider + Clone, B: BackoffBuilder + Clone>(
     provider: &P,
     evm_ctx: &EvmCtx,
+    bot_operator: BotOperator,
     pool: &SqlitePool,
     end_block: u64,
     retry_strategy: B,
@@ -67,6 +74,7 @@ pub(crate) async fn backfill_events<P: Provider + Clone, B: BackoffBuilder + Clo
     backfill_range(
         provider,
         evm_ctx,
+        bot_operator,
         pool,
         start_block,
         end_block,
@@ -76,9 +84,12 @@ pub(crate) async fn backfill_events<P: Provider + Clone, B: BackoffBuilder + Clo
     .await
 }
 
-/// Fetches `ClearV3` / `TakeOrderV3` logs in `[from_block, to_block]`,
-/// pushes an `AccountForDexTrade` job for each, and advances the
-/// backfill checkpoint to `to_block` on success.
+/// Fetches `ClearV3` / `TakeOrderV3` logs from the OrderBook and
+/// `OperatorDeposit` / `OperatorWithdraw` logs from the shared
+/// `RaindexInventory` in `[from_block, to_block]`, pushes an
+/// `AccountForDexTrade` job for each (inventory legs first paired into a single
+/// `InventoryTrade` per settlement tx), and advances the backfill checkpoint to
+/// `to_block` on success.
 ///
 /// Skips RPC calls when `from_block > to_block` (already caught up),
 /// but still moves the checkpoint forward so a stale row does not
@@ -92,6 +103,7 @@ pub(crate) async fn backfill_events<P: Provider + Clone, B: BackoffBuilder + Clo
 pub(crate) async fn backfill_range<P: Provider + Clone, B: BackoffBuilder + Clone>(
     provider: &P,
     evm_ctx: &EvmCtx,
+    bot_operator: BotOperator,
     pool: &SqlitePool,
     from_block: u64,
     to_block: u64,
@@ -124,6 +136,7 @@ pub(crate) async fn backfill_range<P: Provider + Clone, B: BackoffBuilder + Clon
         let enqueued = enqueue_batch_events(
             provider,
             evm_ctx,
+            bot_operator,
             batch_start,
             batch_end,
             retry_strategy.clone(),
@@ -143,8 +156,9 @@ pub(crate) async fn backfill_range<P: Provider + Clone, B: BackoffBuilder + Clon
 /// Persistent job queue for backfill jobs.
 pub(crate) type BackfillJobQueue = crate::conductor::job::JobQueue<BackfillRange>;
 
-/// Apalis job that backfills missed `ClearV3` / `TakeOrderV3` events
-/// between `from_block` and `to_block` (inclusive).
+/// Apalis job that backfills missed `ClearV3` / `TakeOrderV3` orderbook fills
+/// and `OperatorDeposit` / `OperatorWithdraw` inventory settlements between
+/// `from_block` and `to_block` (inclusive).
 ///
 /// Enqueued by `OrderFillMonitor::run` on every (re)connect to cover
 /// the gap between the previously checkpointed block and the cutoff
@@ -187,6 +201,7 @@ where
         backfill_range(
             ctx.evm.provider(),
             &ctx.ctx.evm,
+            BotOperator(ctx.ctx.order_owner()),
             &ctx.pool,
             self.from_block,
             self.to_block,
@@ -202,6 +217,12 @@ where
 /// a cold start with no checkpoint begins at `deployment_block`. Pure so a
 /// caller that already holds the checkpoint (the fill monitor's poll loop) can
 /// reuse it without a second read.
+///
+/// RUNBOOK: the persisted checkpoint predates inventory (`OperatorDeposit` /
+/// `OperatorWithdraw`) ingestion, so resuming from it silently skips any
+/// adapter/inventory fills mined before this code's deploy block. On rollout,
+/// rewind the checkpoint to the inventory deploy block (or run a one-shot
+/// backfill over that range) so pre-deploy inventory fills are ingested.
 pub(crate) fn backfill_start_from_checkpoint(
     checkpoint: Option<u64>,
     deployment_block: u64,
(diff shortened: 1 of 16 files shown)


hardyjosh:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/970?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#970** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/970?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/970?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#969** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/969?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=970).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/onchain/trade.rs
```diff
@@ -326,6 +356,99 @@ impl OnchainTrade {
...
+        if trade_details.equity_amount().is_zero()? {
+            return Ok(None);
+        }
+
+        let price_per_share_usdc =
+            (trade_details.usdc_amount().value() / trade_details.equity_amount().inner())?;
+
+        if price_per_share_usdc.lt(Float::zero()?)? || price_per_share_usdc.is_zero()? {
+            return Ok(None);
+        }
```
0xgleb: should fix: the ClearV3/TakeOrderV3 path returns `Err(TradeValidationError::NonPositivePrice(...))` for this exact condition, and its comment explains why: silently dropping an unpriceable fill leaves the position unhedged. Returning `Ok(None)` here skips `persist_skipped_fill` entirely, so the fill vanishes with only a misleading "Event filtered out (no matching owner)" debug line downstream. Return the `NonPositivePrice` error instead, the `perform` error handler already has the arm that logs it and records it in `skipped_fills`. Longer term, extracting the shared tail (everything from `try_from_io` onward) into one helper both parsers call would stop the two copies drifting like this.

src/onchain/trade.rs
```diff
@@ -326,6 +356,99 @@ impl OnchainTrade {
...
+            symbol: equity_symbol,
+            equity_token,
+            amount: trade_details.equity_amount(),
+            direction: trade_details.direction(),
+            price,
+            block_number: log.block_number.or(receipt_block_number),
+            block_timestamp: log.block_timestamp.and_then(|timestamp_secs| {
+                let secs: i64 = timestamp_secs.try_into().ok()?;
+                DateTime::from_timestamp(secs, 0)
+            }),
```
0xgleb: should fix: the ClearV3 path calls `resolve_block_timestamp`, which falls back to fetching the block header when the log carries no timestamp. `blockTimestamp` on `eth_getLogs` results is optional in the RPC spec, and `process_queued_trade` hard-errors with `MissingBlockTimestamp` when this field is `None`, so on providers that omit it every inventory fill fails the job and exhausts retries while ClearV3 fills keep working. Call `resolve_block_timestamp(evm.provider(), log.block_timestamp, log.block_number.or(receipt_block_number))` here like the sibling path does. That also fixes this inline closure swallowing the `i64` conversion failure with `.ok()`, which the sibling's `timestamp_from_secs` propagates.

src/onchain/trade.rs
```diff
@@ -326,6 +356,99 @@ impl OnchainTrade {
...
 
+    /// Converts a paired [`InventoryTrade`] (an `OperatorDeposit` +
+    /// `OperatorWithdraw` for the same tx on the shared inventory) into an
+    /// [`OnchainTrade`] the hedge pipeline can consume.
+    ///
+    /// The pool received `deposit.amount` of `deposit.token` and sent
+    /// `withdraw.amount` of `withdraw.token` out. One side must be USDC and
+    /// the other a tokenized-equity ERC20; anything else fails validation the
+    /// same way a mis-shaped Raindex clear would.
+    pub(crate) async fn try_from_inventory_trade<EvmImpl: Evm>(
```
0xgleb, line 449: should fix: this function prices a hedge and picks its direction but ships with no tests at all. At minimum it needs a happy-path pair asserting symbol, direction, and price against literal values, the zero-equity and non-positive-price early returns, and a pair where USDC is on the deposit side so the direction flip is covered. The existing TakeOrderV3 parser tests already show how to mock the receipt and decimals calls.

src/onchain/trade.rs
```diff
@@ -326,6 +356,99 @@ impl OnchainTrade {
...
+
+        let input_amount = Float::from_fixed_decimal(trade.deposit.amount, input_decimals)?;
+        let output_amount = Float::from_fixed_decimal(trade.withdraw.amount, output_decimals)?;
+
+        let trade_details =
+            TradeDetails::try_from_io(&input_symbol, input_amount, &output_symbol, output_amount)?;
+
+        if trade_details.equity_amount().is_zero()? {
+            return Ok(None);
+        }
```
0xgleb: minor: this early return is silent, and downstream the `Ok(None)` shows up as "Event filtered out (no matching owner)", which isn't what happened. A `debug!` with the tx hash and the actual reason before returning would keep the drop observable, matching how the ClearV3 path reports its skips.

src/onchain/trade.rs
```diff
@@ -326,6 +356,99 @@ impl OnchainTrade {
...
+
+        if price_per_share_usdc.lt(Float::zero()?)? || price_per_share_usdc.is_zero()? {
+            return Ok(None);
+        }
+
+        let (equity_symbol_str, equity_token) = if input_symbol == "USDC" {
+            (output_symbol, trade.withdraw.token)
+        } else {
+            (input_symbol, trade.deposit.token)
+        };
```
0xgleb: minor: `TradeDetails::try_from_io` already worked out which side is the equity, so comparing `input_symbol == "USDC"` re-derives that classification with a raw string sentinel. Consider having `TradeDetails` carry the equity token (accept token addresses in `try_from_io`) so this block disappears.

src/onchain/trade.rs
```diff
@@ -215,6 +238,13 @@ async fn enrich_with_pyth_price<P: Provider>(
     }
 }
 
+async fn fetch_token_decimals<E: Evm>(evm: &E, token: Address) -> Result<u8, OnChainError> {
+    use st0x_evm::{IERC20, OpenChainErrorRegistry};
```
0xgleb: minor: function-body imports are limited to enum variants and cfg-gated code by the repo import rules, and neither applies here. Move `IERC20` and `OpenChainErrorRegistry` up to the workspace import group next to `use st0x_evm::Evm`.

src/onchain/backfill.rs
```diff
@@ -309,31 +326,64 @@ async fn enqueue_batch_events<P: Provider + Clone, B: BackoffBuilder + Clone>(
...
+    let mut withdraw_logs: Vec<alloy::rpc::types::Log> = Vec::new();
+    for log in inv_logs {
+        let Some(topic0) = log.topic0().copied() else {
+            continue;
+        };
+        if topic0 == OperatorDeposit::SIGNATURE_HASH {
+            if let (Ok(decoded), Some(tx_hash)) =
+                (log.log_decode::<OperatorDeposit>(), log.transaction_hash)
+            {
+                deposits_by_tx.insert(tx_hash, decoded.data().clone());
```
0xgleb: should fix: this map holds one deposit per tx, so if a single transaction ever routes two settlements (batch adapter, multicall), the later deposit overwrites the earlier one and every withdraw in that tx pairs against the surviving deposit, producing a wrong price and direction on a real hedge. If the contract guarantees one pair per tx, assert it: warn or error when inserting over an existing entry instead of silently overwriting. Also worth importing `HashMap` at the top of the module rather than fully qualifying `std::collections::HashMap` here.

src/onchain/backfill.rs
```diff
@@ -309,31 +326,64 @@ async fn enqueue_batch_events<P: Provider + Clone, B: BackoffBuilder + Clone>(
...
+            }
```
0xgleb: minor: a deposit that fails to decode or has no tx hash is dropped without a log line, and its paired withdraw then gets skipped with a "no paired deposit" warning that points at the wrong cause. Log the decode failure here so the operator sees the real reason.

src/onchain/backfill.rs
```diff
@@ -358,12 +408,29 @@ async fn enqueue_batch_events<P: Provider + Clone, B: BackoffBuilder + Clone>(
...
+                        ?tx_hash,
+                        "OperatorWithdraw without a paired OperatorDeposit in the same batch; skipping",
+                    );
+                    None
+                })?;
+                let inv = InventoryTrade {
+                    deposit,
+                    withdraw: withdraw_event.data().clone(),
+                };
+                Some((RaindexTradeEvent::InventoryTrade(Box::new(inv)), log))
```
0xgleb: should fix: none of the test updates in this file exercise the new pairing logic, they are all empty-array stubs keeping the asserters happy with the third eth_getLogs call. This is the only entry point for inventory fills, so it needs one test feeding a real deposit+withdraw pair through `enqueue_batch_events` asserting exactly one InventoryTrade job, and one with an unpaired withdraw asserting it gets skipped. A third case mixing an inventory pair with a take order in one batch would pin the interleaved (block, log_index) ordering.

src/onchain/backfill.rs
```diff
@@ -358,12 +408,29 @@ async fn enqueue_batch_events<P: Provider + Clone, B: BackoffBuilder + Clone>(
...
+                let withdraw_event = log.log_decode::<OperatorWithdraw>().ok()?;
+                let tx_hash = log.transaction_hash?;
+                let deposit = deposits_by_tx.get(&tx_hash).cloned().or_else(|| {
+                    warn!(
+                        target: "inventory",
+                        ?tx_hash,
+                        "OperatorWithdraw without a paired OperatorDeposit in the same batch; skipping",
+                    );
+                    None
+                })?;
```
0xgleb: minor: the `warn!` lives inside the `filter_map` via `or_else`, a side effect hidden in a transform. Materialize the miss as a value (or use a plain `match` on the lookup) and log it in its own step so the pipeline stays pure and the skip logic is visible at a glance.

src/onchain/backfill.rs
```diff
@@ -294,10 +296,25 @@ async fn enqueue_batch_events<P: Provider + Clone, B: BackoffBuilder + Clone>(
...
         .event_signature(TakeOrderV3::SIGNATURE_HASH);
 
+    // Inventory events: OperatorDeposit + OperatorWithdraw are emitted as a
+    // pair per settlement tx that routes through the shared inventory (Bebop
+    // adapter, univ4 hook, or any future venue). The pair -- one deposit +
+    // one withdraw, one side equity + one side USDC -- is the trade to hedge.
+    let inventory_filter = Filter::new()
+        .address(evm_ctx.inventory)
+        .from_block(batch_start)
+        .to_block(batch_end)
```
0xgleb, line 622: minor: the backfill starts from the persisted checkpoint, so adapter fills that happened before this code deploys will never be ingested and stay unhedged. Worth a runbook note to rewind the checkpoint to the inventory deployment block (or run a one-shot backfill over that range) when this rolls out.

src/onchain/backfill.rs
```diff
@@ -294,10 +296,25 @@ async fn enqueue_batch_events<P: Provider + Clone, B: BackoffBuilder + Clone>(
...
+        .event_signature(vec![
+            OperatorDeposit::SIGNATURE_HASH,
+            OperatorWithdraw::SIGNATURE_HASH,
+        ]);
```
0xgleb, line 626: should fix: this filter also matches the bot's own rebalancing traffic. Since #969 the rebalancer calls `deposit4`/`withdraw4` on the inventory, and those emit Operator events with the bot as operator (`find_recent_withdrawal` scans `OperatorWithdraw` to adopt the bot's own in-flight withdrawals). Every rebalancing withdrawal is therefore an expected unpaired `OperatorWithdraw`: the new warn fires on routine operations, genuine venue anomalies drown in that noise, and the description's framing of the unpaired case as a guard rather than an expected path doesn't hold. Consider skipping events whose operator is the bot's own address before pairing, which also removes the phantom-trade risk if the bot ever batches a deposit and a withdraw in one tx.

src/trading/onchain/trade_accountant.rs
```diff
@@ -119,6 +123,17 @@ where
...
+            InventoryTrade(inv) => {
+                OnchainTrade::try_from_inventory_trade(
+                    &ctx.cache,
+                    &ctx.evm,
+                    inv.as_ref(),
+                    reconstructed_log,
+                    &ctx.pyth_feed_ids,
+                )
+                .await
+            }
```
0xgleb, line 142: minor: this arm has no test. A variant of the existing `perform` tests with an InventoryTrade payload (seed the asserter with receipt, decimals, and symbol responses) would cover the wiring from event to accounting.

src/trading/onchain/trade_accountant.rs
```diff
@@ -250,12 +265,21 @@ async fn persist_skipped_fill(
...
+    use RaindexTradeEvent::{ClearV3, InventoryTrade, TakeOrderV3};
+
+    // Pick the emitting contract for each event type. InventoryTrade events
+    // are emitted by the inventory contract; the paired-event log data
+    // encodes the OperatorWithdraw (the equity-side change the hedge cares
+    // about), which is enough for the downstream parser.
+    let (address, log_data) = match &trade.event {
+        ClearV3(clear_event) => (orderbook, clear_event.as_ref().clone().into_log_data()),
+        TakeOrderV3(take_event) => (orderbook, take_event.as_ref().clone().into_log_data()),
+        InventoryTrade(inv) => (inventory, inv.withdraw.clone().into_log_data()),
```
0xgleb: minor: quick test worth adding: call `reconstruct_log` with an InventoryTrade and assert the log address is the inventory (not the orderbook) and that the data round-trips through `OperatorWithdraw::log_decode`. Cheap insurance for the address selection this PR introduces.

src/onchain/trade.rs
```diff
@@ -326,6 +356,99 @@ impl OnchainTrade {
...
+        let (input_symbol, output_symbol) = tokio::try_join!(
+            cache.resolve_symbol(evm, trade.deposit.token),
+            cache.resolve_symbol(evm, trade.withdraw.token),
+        )?;
+
+        let input_decimals = fetch_token_decimals(evm, trade.deposit.token).await?;
+        let output_decimals = fetch_token_decimals(evm, trade.withdraw.token).await?;
+
+        let input_amount = Float::from_fixed_decimal(trade.deposit.amount, input_decimals)?;
+        let output_amount = Float::from_fixed_decimal(trade.withdraw.amount, output_decimals)?;
```
0xgleb: nit: two same-typed `u8` locals feeding adjacent `from_fixed_decimal` calls, nothing stops a swap that would scale an amount by the wrong power of ten. Inlining `fetch_token_decimals(...).await?` into each call site removes the free-floating bindings.

src/onchain/backfill.rs
```diff
@@ -309,31 +326,64 @@ async fn enqueue_batch_events<P: Provider + Clone, B: BackoffBuilder + Clone>(
...
+                trace!(target: "inventory", "Retrying inv_logs for blocks between {batch_start}-{batch_end} after error: {err} (waiting {dur:?})");
+            }),
     )
     .await?;
 
     debug!(
         target: "orderbook",
         total_clear_logs = %clear_logs.len(),
         total_take_logs = %take_logs.len(),
+        total_inventory_logs = %inv_logs.len(),
```
0xgleb, line 674: nit: the surrounding docs still describe two event sources: the module docstring ("Scans past blocks for ClearV3 and TakeOrderV3 events"), the `backfill_range` and `BackfillRange` docs, and the `onchain_events_total` counter comment saying `kind()` yields "ClearV3"/"TakeOrderV3". All four now under-document the third InventoryTrade source, and the counter comment is the reference for the metrics label values.

0xgleb (changes requested):

src/onchain/trade.rs
```diff
@@ -215,6 +238,13 @@ async fn enrich_with_pyth_price<P: Provider>(
     }
 }
 
+async fn fetch_token_decimals<E: Evm>(evm: &E, token: Address) -> Result<u8, OnChainError> {
+    use st0x_evm::{IERC20, OpenChainErrorRegistry};
+    Ok(evm
+        .call::<OpenChainErrorRegistry, _>(token, IERC20::decimalsCall {})
+        .await?)
```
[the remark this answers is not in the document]
JuaniRios: Moved `IERC20`/`OpenChainErrorRegistry` to the module-top `use st0x_evm::{Evm, IERC20, OpenChainErrorRegistry}` and dropped the fn-body import. 74f24654.

src/onchain/trade.rs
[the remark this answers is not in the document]
JuaniRios: Added a `debug!` with tx hash + log_index before the zero-equity `Ok(None)` skip, matching the ClearV3 path. 74f24654.

src/onchain/trade.rs
JuaniRios: Fixed: the inventory path now returns `Err(TradeValidationError::NonPositivePrice(..))` (same as the ClearV3 sibling), so an unpriceable non-zero-equity fill is error-logged and recorded via `persist_skipped_fill(UnpriceableFill)` instead of vanishing as `Ok(None)`. 74f24654.

src/onchain/trade.rs
JuaniRios: Fixed: replaced the inline `.ok()` closure with `resolve_block_timestamp(evm.provider(), log.block_timestamp, log.block_number.or(receipt_block_number)).await?`, so it falls back to the block header when `eth_getLogs` omits `blockTimestamp` and propagates the i64 conversion error instead of swallowing it. 74f24654.

src/onchain/trade.rs
JuaniRios: Added parser tests: happy path (asserts symbol/direction/price), zero-equity -> `Ok(None)`, non-positive-price -> `Err(NonPositivePrice)`, and a USDC-on-deposit case exercising the direction flip. 74f24654.

src/onchain/trade.rs
JuaniRios: Added the `debug!` skip log before the `Ok(None)` return, so the drop is observable rather than showing up as the misleading "no matching owner" line downstream. 74f24654.

src/onchain/trade.rs
JuaniRios: Refactored: `TradeDetails` now carries the resolved `equity_symbol` + `equity_token` (from `try_from_io`, which takes the token addresses), so both parsers use `trade_details.equity_token()`/`equity_symbol()` and the raw `== "USDC"` sentinel block is gone. 74f24654.

src/onchain/trade.rs
JuaniRios: Moved `IERC20`/`OpenChainErrorRegistry` to the module-top workspace import group next to `st0x_evm::Evm`. 74f24654.

src/onchain/backfill.rs
JuaniRios: `deposits_by_tx.insert` now `warn!`s when it overwrites an existing deposit for the same tx (mispairing guard), and `HashMap` is imported at the top of the module. 74f24654.

src/onchain/backfill.rs
JuaniRios: The deposit bucketing `if let (Ok, Some)` guard is now a `let ... else` that `warn!`s "Skipping OperatorDeposit that failed to decode / had no tx hash" before dropping, so the paired withdraw's later warning isn't misleading. 74f24654.

src/onchain/backfill.rs
JuaniRios: Added real pairing tests through `enqueue_batch_events`: a deposit+withdraw pair -> exactly one InventoryTrade job; an unpaired withdraw -> skipped; a bot-operator pair -> skipped (T13); and an inventory pair interleaved with a take order asserting `(block, log_index)` order. 74f24654.

src/onchain/backfill.rs
JuaniRios: Materialized the unpaired-withdraw miss out of the `.or_else`: it's now an explicit `let Some(deposit) = deposits_by_tx.get(&tx_hash).cloned() else { warn!(..); return None; };` so the skip is visible and the pipeline stays pure. 74f24654.

src/onchain/backfill.rs
JuaniRios: Added a `RUNBOOK:` note at the checkpoint site: on rollout the persisted checkpoint predates inventory ingestion, so it must be rewound to the inventory deploy block (or a one-shot backfill run) to catch pre-deploy adapter fills. 74f24654.

src/onchain/backfill.rs
JuaniRios: Fixed: the pairing now skips any `Operator{Deposit,Withdraw}` whose decoded `operator == order_owner` (the bot's signing wallet, threaded from `BackfillRange::perform`) before bucketing. Since #969 the bot's own rebalancing emits these events, so this stops the warn-noise on every routine rebalance withdrawal and removes the phantom-trade risk if the bot ever batches a deposit+withdraw in one tx. 74f24654.

src/trading/onchain/trade_accountant.rs
JuaniRios: Added a `perform` test driving an `InventoryTrade` payload (seeded receipt/decimals/symbol) through the event->accounting wiring. 74f24654.

src/trading/onchain/trade_accountant.rs
JuaniRios: Added a `reconstruct_log` test for an InventoryTrade asserting the log address is the inventory (not the orderbook) and that the data round-trips through `OperatorWithdraw::log_decode`. 74f24654.

src/onchain/trade.rs
JuaniRios: Inlined `fetch_token_decimals(...).await?` into each `Float::from_fixed_decimal` call, removing the two free-floating `u8` bindings and the swap risk. 74f24654.

src/onchain/backfill.rs
JuaniRios: Updated the stale docs: the `BackfillRange` doc now names the OperatorDeposit/OperatorWithdraw source, and the `onchain_events_total` counter comment now lists "InventoryTrade" among the `kind()` values. (The module `//!` already references `RaindexTradeEvent` generically.) 74f24654.

src/onchain/trade.rs
JuaniRios: Addressed. The shared tail you suggested now exists as `finalize_onchain_trade`, called by both parsers, and it returns `Err(TradeValidationError::NonPositivePrice(..))` for this case rather than `Ok(None)` — so the fill reaches `persist_skipped_fill`. Pinned by `try_from_inventory_trade_non_positive_price_errors`.

src/onchain/trade.rs
JuaniRios: Addressed. The inventory path now goes through `finalize_onchain_trade`, which calls `resolve_block_timestamp(provider, log_timestamp, block_number)` with the same block-header fallback as ClearV3. The inline closure is gone, so the `i64` conversion failure propagates via `timestamp_from_secs` instead of being swallowed by `.ok()`.

src/onchain/trade.rs
JuaniRios: Addressed. `try_from_inventory_trade` now has a happy-path buy asserting symbol/direction/price/amount against literals, a USDC-on-deposit case covering the direction flip, the zero-equity `None` return, the non-positive-price error, a spoofed-token rejection, and two real prod fixtures (bebop and univ4).

src/onchain/trade.rs
JuaniRios: Addressed. The zero-equity return now logs `debug!` with the tx hash, log index and the actual reason before returning `Ok(None)`, so it no longer surfaces downstream as a misleading "no matching owner".

src/onchain/trade.rs
JuaniRios: Addressed. The `input_symbol == "USDC"` sentinel is gone; `validate_inventory_token_addresses` now uses the classification `TradeDetails` already carries via `usdc_token()`, `equity_symbol()` and `equity_token()`.

src/onchain/trade.rs
JuaniRios: Addressed. `IERC20` and `OpenChainErrorRegistry` moved to the workspace import group at the top of the module, alongside `Evm`.

src/onchain/trade.rs
JuaniRios: Leaving as-is, with reasoning. The two `u8` bindings are now outputs of a 4-way `tokio::try_join!` batch, so inlining `fetch_token_decimals(..).await?` into each `from_fixed_decimal` call site would serialize those RPCs. Both orientations (deposit 18 / withdraw 6 and the reverse) are directly test-covered, so a swap would fail the suite. Genuinely removing the residual risk needs distinct newtypes rather than the suggested inlining — happy to do that separately if you think it's worth it.

src/onchain/backfill.rs
JuaniRios: Addressed. A second same-tx `OperatorDeposit` no longer overwrites: the `Entry::Occupied` arm logs an error, increments `inventory_ambiguous_settlement_total`, and stores `DepositBucket::Ambiguous`, which `pair_inventory_settlements` never builds a trade from — fail closed, as you asked. `HashMap` is now imported at the top of the module.

src/onchain/backfill.rs
JuaniRios: Addressed. A deposit that fails to decode or has no tx hash now warns at the point of failure (and symmetrically for withdraws), so the operator sees the real cause rather than only the downstream "no paired deposit" warning.

src/onchain/backfill.rs
JuaniRios: Addressed. `backfill.rs` now feeds real decoded logs through `enqueue_batch_events`: a deposit+withdraw pair yielding exactly one `InventoryTrade` job, an unpaired withdraw being skipped, an inventory pair interleaved with a TakeOrderV3 asserting both enqueue in `(block, log_index)` order, plus ambiguous and mismatched-operator cases and real-fixture buy/sell assertions.

src/onchain/backfill.rs
JuaniRios: Addressed. `pair_inventory_settlements` is now a plain `for` loop that matches, logs and increments as statements, pushing only the paired case — no side effects hidden inside a transform. Behaviour is identical and the existing pairing tests still pass.

src/onchain/backfill.rs
JuaniRios: Addressed. `backfill_start_from_checkpoint` now carries a `RUNBOOK:` note to rewind the checkpoint to the inventory deploy block (or run a one-shot backfill over that range) on rollout, and the same note is in SPEC.md.

src/onchain/backfill.rs
JuaniRios: Good catch — addressed. Legs whose `operator` is the bot's own signing wallet are now dropped before bucketing (`bot_operator` threaded through as `BotOperator(ctx.ctx.order_owner())`), so routine rebalancing withdrawals no longer fire the unpaired warning and can't be batched into a phantom trade. Covered by `bot_operator_rebalancing_pair_is_skipped`.

src/trading/onchain/trade_accountant.rs
JuaniRios: Addressed. The `InventoryTrade` arm of `perform` is now covered end to end: a happy-path hedge, a real univ4 buy fixture, and the non-hedgeable-pair, unrecognized-token, unintrospectable-token and invalid-amount skip paths.

src/trading/onchain/trade_accountant.rs
JuaniRios: Addressed. `reconstruct_log_for_inventory_trade_uses_inventory_address_and_round_trips_withdraw` in `src/trading/onchain/trade_accountant.rs` asserts the reconstructed log is emitted by the inventory (not the orderbook) and that the data round-trips through `OperatorWithdraw::log_decode`.

src/onchain/backfill.rs
JuaniRios: Addressed. The module docstring and `backfill_range`'s doc now name the inventory source alongside ClearV3/TakeOrderV3, and the `onchain_events_total` `describe_counter!` text in `src/metrics.rs` now lists `InventoryTrade` as a label value. `BackfillRange`'s doc already covered it.

JuaniRios (approved):

0xgleb (approved):

PR #976, opened 2026-07-01
chore: wire staging Turnkey wallet config and secrets

JuaniRios:
## What

- Sets the correct staging Turnkey wallet address in `config/staging/st0x-hedge.toml`'s `[wallet]` section.
- Enables `extended_hours_counter_trading` for RKLB in staging.
- Refreshes `secret/st0x-hedge.toml.age` with the Turnkey API credentials for that wallet.
- Rotates `secret/tailscale-authkey-prod.age` and `secret/tailscale-authkey-staging.age`.

## Why

- The staging Turnkey wallet needs its address and signing credentials configured before the bot can use it.
- Per `AGENTS.md`, a secrets-schema change (adding the Turnkey API credential) must ship with a matching encrypted `.age` update or the `validate-config` deploy gate fails closed.

## How

- `config/staging/st0x-hedge.toml`'s `[wallet]` table now points at the correct Turnkey-controlled address (`organization_id` is unchanged); wallet construction is already generic over `kind = "turnkey"` (`crates/config/src/wallet.rs`), so no code changes were needed here.
- `secret/st0x-hedge.toml.age` re-encrypted via `ragenix` to add the `api_private_key` credential paired with that wallet.

## Testing

- No tests — config/secrets-only change, no code paths added or modified.

## Screenshots

N/a — config/secrets only, no UI change.

## Anything else

- Secrets-schema note: `secret/st0x-hedge.toml.age` now carries the Turnkey API credentials for the staging wallet — must be applied to the deployed secret before deploy or `validate-config` fails closed.
- `secret/tailscale-authkey-{prod,staging}.age` rotation is bundled into this commit but unrelated to the Turnkey wallet change.
- Base of this stack; #978 and #977 build on top.

PR #973, opened 2026-07-01
chore: point staging bot at the st0x.io tailnet

kaisbaccour:
## What

Moves the **staging** liquidity bot's MagicDNS name from the rainlang.xyz tailnet (\`taile5cf8a.ts.net\`) to the **st0x.io** tailnet (\`tail6094d7.ts.net\`), so its OTLP telemetry can reach \`st0x-management-observability\` (\`100.93.167.114\`), which only lives on the st0x.io tailnet.

Prod stays on rainlang.xyz — it migrates last, after staging is validated. \`tailscaleMagicDnsName\` is per-environment in \`flake.nix\`, so this leaves prod untouched.

## Why now

The observability work (#945) points telemetry at \`100.93.167.114\`, but the bots are on the rainlang.xyz tailnet, which cannot see that host. This is the first half of connecting staging to the self-hosted stack.

## Connectivity check

Audited \`config/staging\` — the only tailnet (\`100.x\`) address referenced is the obs box \`100.93.167.114\` (the migration target). Everything else is public (\`https://api.st0x.io\`) or local (\`sqlite\`). So moving staging off rainlang.xyz breaks no dependency. Note: SSH access to the staging droplet moves to the st0x.io tailnet after cutover.

## Operator steps (out of band — NOT done by this PR, in this order)

1. **Auth key** — generate a \`tag:st0x-infra\` auth key on the st0x.io tailnet and re-encrypt \`secret/tailscale-authkey-staging.age\` with it. (The key is used only on *first* enrollment.)
2. **Secret re-key** — the staging deploy also needs the HyperDX key dropped from \`secret/st0x-hedge.toml.age\`, or \`validate-config\` fails the deploy closed.
3. **Deploy staging** (workflow_dispatch; staging takes any ref — can deploy from this branch to validate before merge).
4. **\`tailscale up --force-reauth\`** on the droplet — an already-enrolled node re-auths from its stored node key, so a plain deploy will **not** move it to the new tailnet. This is mandatory.
5. **Scrape target** — add the Tailscale ACL (\`obs -> staging bot : metrics port\`) + scrape config on the st0x.devops side so VictoriaMetrics can pull \`/metrics\`. (Traces/logs push on their own once the node is on the tailnet; metrics are pull and need this.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

PR #978, opened 2026-07-01
feat: show per-asset inventory config flags

JuaniRios:
## What

- Adds per-asset inventory indicators for counter trading, rebalancing, and extended-hours counter trading.
- Renders the indicators as aligned green/red status lights in the equity inventory table.
- Adds glossary entries for `CT`, `Rebal`, and `Ext`.

## Why

- Operators need to confirm per-asset behavior flags while reviewing inventory state.
- Extended-hours and rebalancing can vary by asset, so a single global dashboard signal is not enough.

## How

- Extends `AssetSettings` with `extended_hours` and populates it in `settings_from_ctx`.
- Builds per-row `AssetFlags` from dashboard settings and keeps disabled counter-trading rows dimmed via the existing `trading` row behavior.
- Uses a shared status-grid class for the header labels and lights so `CT`, `Rebal`, and `Ext` stay aligned while the group remains compact.
- Adds `settings_from_ctx_includes_asset_operation_flags` to cover trading, rebalancing, and extended-hours DTO population.

## Testing

- `nix develop .#ci-backend -c cargo fmt -- --check`
- `nix develop .#ci-backend -c cargo nextest run -p st0x-dto`
- `nix develop .#ci-backend -c cargo nextest run -p st0x-hedge settings_from_ctx_includes_asset_operation_flags`
- `nix develop .#ci-dashboard -c bash -lc 'cd dashboard && bun run lint'`
- `nix develop .#ci-dashboard -c bash -lc 'cd dashboard && bun run check'`
- `nix run .#ci` passed before the final dashboard-only spacing/alignment tweaks; the final UI tweak was re-checked with dashboard lint/check.

## Screenshots

Not captured for the final state.

## Anything else

- Stacked on #976.
- Adds a dashboard settings DTO field; no database migration, secret update, or deployed config change is included.

### DIFF
--- a/crates/dto/src/settings.rs
+++ b/crates/dto/src/settings.rs
@@ -57,7 +57,82 @@ pub struct WalletSettings {
 pub struct AssetSettings {
     #[ts(type = "string")]
     pub symbol: Symbol,
-    pub trading: bool,
+    pub counter_trading: CounterTrading,
     pub rebalancing: bool,
     pub operational_limit: Option<String>,
+}
+
+/// Counter-trading (hedging) configuration for an asset.
+///
+/// Extended-hours counter-trading only makes sense when counter-trading is
+/// enabled: the hedge path gates on counter-trading before it ever consults the
+/// extended-hours session, so extended hours without counter-trading is a dead
+/// configuration that can never execute. Nesting `extended_hours` under
+/// `Enabled` makes that invalid combination unrepresentable on the wire, so the
+/// dashboard can never receive a contradictory pair.
+#[derive(Debug, Clone, Serialize, TS)]
+#[serde(rename_all = "camelCase", tag = "status")]
+pub enum CounterTrading {
+    Disabled,
+    #[serde(rename_all = "camelCase")]
+    Enabled {
+        extended_hours: bool,
+    },
+}
+
+#[cfg(test)]
+mod tests {
+    use serde_json::json;
+
+    use st0x_finance::Symbol;
+
+    use super::{AssetSettings, CounterTrading};
+
+    #[test]
+    fn counter_trading_disabled_serializes_with_status_tag_only() {
+        let value = serde_json::to_value(CounterTrading::Disabled).unwrap();
+        assert_eq!(value, json!({ "status": "disabled" }));
+    }
+
+    #[test]
+    fn counter_trading_enabled_serializes_extended_hours_flag() {
+        let value = serde_json::to_value(CounterTrading::Enabled {
+            extended_hours: true,
+        })
+        .unwrap();
+        assert_eq!(value, json!({ "status": "enabled", "extendedHours": true }));
+    }
+
+    #[test]
+    fn counter_trading_enabled_without_extended_hours_serializes_false_flag() {
+        let value = serde_json::to_value(CounterTrading::Enabled {
+            extended_hours: false,
+        })
+        .unwrap();
+        assert_eq!(
+            value,
+            json!({ "status": "enabled", "extendedHours": false })
+        );
+    }
+
+    #[test]
+    fn asset_settings_nests_counter_trading_under_status() {
+        let settings = AssetSettings {
+            symbol: Symbol::new("AAPL").unwrap(),
+            counter_trading: CounterTrading::Enabled {
+                extended_hours: true,
+            },
+            rebalancing: false,
+            operational_limit: None,
+        };
+
+        let value = serde_json::to_value(&settings).unwrap();
+        assert_eq!(value["symbol"], json!("AAPL"));
+        assert_eq!(
+            value["counterTrading"],
+            json!({ "status": "enabled", "extendedHours": true })
+        );
+        assert_eq!(value["rebalancing"], json!(false));
+        assert_eq!(value["operationalLimit"], json!(null));
+    }
 }
--- a/dashboard/src/lib/components/available-inventory.svelte
+++ b/dashboard/src/lib/components/available-inventory.svelte
@@ -98,6 +98,11 @@
   const formatRatio = (ratio: number): string => `${(ratio * 100).toFixed(1)}%`
 
   type PositionInfo = { net: string; priceUsdc: string | null }
+  type AssetFlags = {
+    counterTrading: boolean
+    rebalancing: boolean
+    extendedHours: boolean
+  }
 
   const positionMap = $derived(
     new Map(
@@ -108,6 +113,28 @@
           priceUsdc: position.last_price_usdc
         } satisfies PositionInfo
       ])
+    )
+  )
+
+  const assetFlagsMap = $derived(
+    new Map(
+      settings?.assets.map((asset) => {
+        const counterTrading = asset.counterTrading
+        const flags: AssetFlags =
+          counterTrading.status === 'enabled'
+            ? {
+                counterTrading: true,
+                rebalancing: asset.rebalancing,
+                extendedHours: counterTrading.extendedHours
+              }
+            : {
+                counterTrading: false,
+                rebalancing: asset.rebalancing,
+                extendedHours: false
+              }
+
+        return [asset.symbol, flags] as const
+      }) ?? []
     )
   )
 
@@ -123,12 +150,10 @@
     exposure: number
     netShares: string
     priceUsdc: string | null
-    trading: boolean
+    // null when the symbol has no settings entry (unconfigured): the status
+    // lights render as unknown rather than falsely showing all-disabled.
+    flags: AssetFlags | null
   }
-
-  const tradingSet = $derived(
-    new Set(settings?.assets.filter((asset) => asset.trading).map((asset) => asset.symbol) ?? [])
-  )
 
   const equityRows = $derived<EquityRow[]>(
     symbols.map((item) => {
@@ -139,6 +164,7 @@
       )
       const stripped = stripPrefix(item.symbol)
       const pos = positionMap.get(stripped)
+      const flags = assetFlagsMap.get(stripped) ?? null
 
       return {
         asset: stripped,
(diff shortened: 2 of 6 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/978?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#973** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/973?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#981** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/981?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#978** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/978?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/978?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#976** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/976?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=978).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

dashboard/src/lib/components/available-inventory.svelte
```diff
@@ -598,6 +655,31 @@
...
+            <div class={statusGroupClass}>
+              <span
+                class={statusLightClass(row.counterTrading, row.trading)}
+                role="img"
+                title={statusLabel('Counter-trading', row.counterTrading)}
+                aria-label={statusLabel('Counter-trading', row.counterTrading)}
+              ></span>
+
+              <span
+                class={statusLightClass(row.rebalancing, row.trading)}
```
0xgleb: should fix: all three lights pass row.trading into statusLightClass, so Rebal and Ext get opacity-50 whenever counter-trading is off, even if those modes are enabled. That makes a dim green Rebal light easy to read as off, which works against the per-asset visibility this PR adds. Consider only dimming the CT light from counter-trading, or drop the trading gate for Rebal/Ext.

dashboard/src/lib/components/available-inventory.svelte
```diff
@@ -139,6 +156,11 @@
       )
       const stripped = stripPrefix(item.symbol)
       const pos = positionMap.get(stripped)
+      const flags = assetFlagsMap.get(stripped) ?? {
```
0xgleb: minor: assetFlagsMap.get(stripped) ?? { all false } silently treats a missing settings row as explicitly disabled for CT/Rebal/Ext. If inventory symbols can appear without a settings entry, that is another invalid-state-as-representable problem — model presence explicitly (e.g. optional AssetFlags / known vs unconfigured) rather than defaulting bools to false.

dashboard/src/lib/components/available-inventory.svelte
```diff
@@ -236,6 +261,22 @@
 
   const dimClass = (base: string, trading = true): string => (trading ? base : `${base} opacity-40`)
 
+  const statusLightClass = (enabled: boolean, trading: boolean): string => {
```
0xgleb: should fix: statusLightClass(enabled, trading) is boolean blindness. Two unnamed bools at every call site hide that the second arg is dim-because-counter-trading-is-off, not rebalancing state. Use an options object or split helpers so call sites encode intent.

dashboard/src/lib/components/available-inventory.svelte
```diff
@@ -152,7 +174,10 @@
         exposure: pos?.priceUsdc ? parseFloat(pos.net) * parseFloat(pos.priceUsdc) : 0,
         netShares: pos?.net ?? '0',
         priceUsdc: pos?.priceUsdc ?? null,
-        trading: tradingSet.has(stripped)
+        counterTrading: flags.counterTrading,
+        rebalancing: flags.rebalancing,
+        extendedHours: flags.extendedHours,
+        trading: flags.counterTrading
```
0xgleb: should fix: EquityRow now carries both counterTrading and trading set to the same value. That duplicates one concept under two names and invites drift (some call sites use row.trading for dimming, others row.counterTrading for the CT light). Drop trading and use counterTrading everywhere, or rename once to match the domain.

dashboard/src/lib/components/available-inventory.svelte
```diff
@@ -111,6 +116,19 @@
     )
   )
 
+  const assetFlagsMap = $derived(
+    new Map(
+      settings?.assets.map((asset) => [
+        asset.symbol,
+        {
+          counterTrading: asset.trading,
```
0xgleb: minor: if the wire shape still allows trading=false with extendedHours=true, the UI will show green Ext beside red CT. That is a symptom of the DTO/config modeling gap — do not paper over it here with runtime && clamping; fix validation + AssetSettings so the dashboard never receives a contradictory pair.

crates/dto/src/settings.rs
```diff
@@ -59,5 +59,6 @@ pub struct AssetSettings {
     pub symbol: Symbol,
     pub trading: bool,
     pub rebalancing: bool,
+    pub extended_hours: bool,
```
0xgleb: should fix: trading and extended_hours as independent bools make extended_hours=true with trading=false representable on the wire, but that combo cannot execute (hedge path gates on trading before extended session). AGENTS type modeling: make invalid states unrepresentable — reject extended_hours_counter_trading=enabled when trading=disabled in config validation, and shape AssetSettings so the dependency is encoded in types (e.g. export OperationMode, or an ADT where extended hours exists only under enabled counter-trading). Do not clamp or derive in settings_from_ctx or the UI.

src/dashboard/mod.rs
```diff
@@ -193,6 +193,7 @@ pub(crate) fn settings_from_ctx(ctx: &st0x_config::Ctx) -> st0x_dto::Settings {
                 symbol: symbol.clone(),
                 trading: config.trading == OperationMode::Enabled,
                 rebalancing: config.rebalancing == OperationMode::Enabled,
+                extended_hours: config.extended_hours_counter_trading == OperationMode::Enabled,
```
0xgleb: should fix: this mapping faithfully emits extended_hours=true with trading=false for configs like extended_hours_counter_trading_parses — a state that should not exist. Fix in validate_files / EquityAssetConfig validation, then DTO types that cannot express the combo. settings_from_ctx should map valid config only, not silently coerce ext when trading is off.

0xgleb (changes requested):

dashboard/src/lib/components/available-inventory.svelte
JuaniRios: Fixed in 17a9fa2a. Dropped the counter-trading dim from the status lights entirely -- each of CT/Rebal/Ext now shows its own state at full opacity, so an enabled Rebal/Ext no longer reads as off. Row-level dimming (asset name + numeric cells) still conveys counter-trading-off.

dashboard/src/lib/components/available-inventory.svelte
JuaniRios: Fixed in 17a9fa2a. `statusLightClass` now takes a single `LightState` (`'enabled' | 'disabled' | 'unknown'`) instead of two unnamed bools. The dim-because-counter-trading-off argument is gone, so there's no hidden second meaning at the call site.

dashboard/src/lib/components/available-inventory.svelte
JuaniRios: Fixed in 17a9fa2a. Dropped both `counterTrading` and `trading` from `EquityRow` in favour of a single `flags: AssetFlags | null`. Dimming/sorting derive a local `ctActive` from `flags`, so there's one source of truth and no drift.

dashboard/src/lib/components/available-inventory.svelte
JuaniRios: Fixed in 17a9fa2a. `EquityRow` now carries `flags: AssetFlags | null`; a symbol with no settings entry stays `null` and renders a distinct grey "unknown" light (labelled "unconfigured") instead of defaulting the bools to false and showing all-disabled.

crates/dto/src/settings.rs
JuaniRios: Done in 17a9fa2a. `AssetSettings` now uses `counter_trading: CounterTrading` where `CounterTrading = Disabled | Enabled { extended_hours }` -- extended hours only exists under enabled counter-trading, so the invalid pair is unrepresentable on the wire. Config validation (`parse_and_validate`) also rejects `extended_hours_counter_trading=enabled` when `trading=disabled` via a dedicated `CtxError` variant. Confirmed the dead-combo claim: `check_execution_readiness` returns early on disabled trading before the extended-hours session is ever consulted.

src/dashboard/mod.rs
JuaniRios: Fixed in 17a9fa2a. With validation rejecting the combo up front, `settings_from_ctx` maps valid config only: `trading=Enabled` -> `CounterTrading::Enabled { extended_hours }`, `Disabled` -> `CounterTrading::Disabled`. No silent coercion -- the invalid pair can't reach this mapping.

dashboard/src/lib/components/available-inventory.svelte
JuaniRios: Fixed at the source in 17a9fa2a rather than clamped in the UI: config validation rejects the contradictory pair and the DTO models counter-trading as an ADT (see the settings.rs thread), so the dashboard can no longer receive `extendedHours=true` beside a disabled CT.

0xgleb (approved):

PR #981, opened 2026-07-01
fix: repair legacy USDC conversion events

JuaniRios:
## What

- Adds a migration repairing legacy `UsdcRebalanceEvent::ConversionConfirmed` payloads so historical aggregates replay under the current conversion schema ([RAI-1199](https://linear.app/makeitrain/issue/RAI-1199/repair-legacy-usdc-conversion-event-payloads))
- Adds `tests/migrations.rs` covering the repair migration

## Why

- Prod has legacy confirmation payloads with only `filled_amount`; current replay code expects `source_amount` and `received_amount`
- Those rows block `UsdcRebalance` aggregate hydration and startup health

## How

- Migration `20260701223808_repair_legacy_usdc_conversion_confirmed_events.sql` rewrites only legacy `UsdcRebalance` `ConversionConfirmed` rows (`filled_amount` present, `source_amount` missing) into `source_amount`/`received_amount` at par, bumping `event_version` to `2.0`
- Clears `UsdcRebalance` snapshots so aggregates rebuild from the repaired event stream
- Leaves already-migrated events and unrelated aggregates untouched

## Testing

- New `tests/migrations.rs::repairs_legacy_usdc_conversion_confirmed_events` covers: legacy rows rewritten to `source_amount`/`received_amount` at par and bumped to `2.0`, already-current rows left untouched, unrelated aggregate payloads left untouched, and `UsdcRebalance` snapshots cleared while other aggregate snapshots (e.g. `Position`) survive

## Screenshots

N/A

## Anything else

- This intentionally treats legacy `source_amount`/`received_amount` as equal to the old `filled_amount` — an approximation of historical accounting since the original split wasn't recorded

## Summary by CodeRabbit

* **Bug Fixes**
  * Repaired historical USDC conversion-confirmed records so affected balances now use the correct amount fields.
  * Ensured rebuilt aggregates are based on the corrected event history, helping keep past data consistent.

* **Tests**
  * Added coverage for the migration to verify legacy and current USDC records are updated correctly and unrelated data stays unchanged.

### DIFF
--- a/infra/default.nix
+++ b/infra/default.nix
@@ -222,6 +222,47 @@ let
         '';
       };
 
+      # Downloads a *consistent* copy of the live database: `scp`-ing
+      # /mnt/data/st0x-hedge.db directly (as ${env}-status does for its
+      # bundled DB) can race the bot's live writes and produce a torn,
+      # corrupt file -- SQLite's file format has no atomicity guarantee
+      # against a plain file copy of a database still being written to.
+      # `VACUUM INTO` runs a normal read transaction against the live DB and
+      # writes a guaranteed-consistent copy server-side first; only that
+      # copy gets shipped over the wire, then deleted from the remote host.
+      "${env}DbSnapshot" = pkgs.writeShellApplication {
+        name = "${env}-db-snapshot";
+        runtimeInputs = sshInputs ++ [ pkgs.coreutils ];
+        text = ''
+          ${resolveHost}
+          trap _cleanup_identity EXIT
+
+          ssh_remote() {
+            # shellcheck disable=SC2029
+            ssh ''${identity:+-i "$identity"} "root@$host_ip" "$@"
+          }
+
+          db_path="/mnt/data/st0x-hedge.db"
+          remote_snapshot="/tmp/st0x-hedge-snapshot-$(date -u +%Y%m%d%H%M%S).db"
+          out_dir="''${1:-./.tmp/$(date -u +%Y-%m-%d_%H-%M-%S)-${env}-db-snapshot}"
+          mkdir -p "$out_dir"
+          local_snapshot="$out_dir/st0x-hedge.db"
+
+          echo "Taking a consistent snapshot of ${env}'s live database..." >&2
+          ssh_remote "sqlite3 $db_path \"VACUUM INTO '$remote_snapshot'\""
+
+          cleanup_remote_snapshot() {
+            ssh_remote "rm -f $remote_snapshot" || true
+          }
+          trap 'cleanup_remote_snapshot; _cleanup_identity' EXIT
+
+          echo "Downloading snapshot to $local_snapshot..." >&2
+          scp ''${identity:+-i "$identity"} "root@$host_ip:$remote_snapshot" "$local_snapshot"
+
+          echo "$local_snapshot"
+        '';
+      };
+
       "${env}BotStart" = pkgs.writeShellApplication {
         name = "${env}-bot-start";
         runtimeInputs = sshInputs;
--- a/src/migration_verification.rs
+++ b/src/migration_verification.rs
@@ -1,0 +1,505 @@
+//! Verifies that migrations apply cleanly to a real (prod/staging) database
+//! and that every currently persisted event still replays under the
+//! CURRENT aggregate code.
+//!
+//! Catches cases where an event or aggregate shape change breaks legacy
+//! data that no migration has repaired yet. Never mutates the database it
+//! is pointed at: everything runs against a disposable `VACUUM INTO` copy
+//! in a scratch temp directory. Safe to point at a live database (once the
+//! writer is stopped) or a downloaded snapshot -- see
+//! `src/bin/verify-migrations.rs` for the CLI entry point used both as a
+//! pre-deploy gate and for manual testing while developing a migration.
+
+use std::fmt;
+use std::path::Path;
+use std::str::FromStr;
+
+use sqlx::SqlitePool;
+use sqlx::sqlite::SqliteConnectOptions;
+use thiserror::Error;
+
+use st0x_event_sorcery::{EventSourced, load_all_ids, load_entity};
+
+use crate::equity_redemption::EquityRedemption;
+use crate::inventory::snapshot::InventorySnapshot;
+use crate::offchain::order::OffchainOrder;
+use crate::onchain_trade::OnChainTrade;
+use crate::position::Position;
+use crate::tokenized_equity_mint::TokenizedEquityMint;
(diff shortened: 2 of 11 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/981?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#973** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/973?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#981** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/981?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/981?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#978** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/978?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#976** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/976?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=981).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

tests/migrations.rs
```diff
@@ -0,0 +1,243 @@
...
+    )
+    .await?;
+    insert_snapshot(&pool, "UsdcRebalance", "legacy-usdc-rebalance").await?;
+    insert_snapshot(&pool, "Position", "SPYM").await?;
+
+    sqlx::raw_sql(REPAIR_LEGACY_USDC_CONVERSION_CONFIRMED_EVENTS)
+        .execute(&pool)
+        .await?;
+
+    assert_conversion_field(
```
0xgleb, line 64: minor: this test checks json_extract after the repair SQL but never replays the aggregate. The prod failure is UsdcRebalance hydration, not SQLite field shape. After the assertions here, call load_entity on the legacy row (with enough preceding events for a valid stream) or add a migration_verification test that seeds filled_amount and runs verify_migrations. offchain/order.rs already has the legacy replay pattern.

src/migration_verification.rs
```diff
@@ -0,0 +1,424 @@
...
+        .map_err(VerificationError::Vacuum)?;
+    source_pool.close().await;
+
+    let scratch_options = SqliteConnectOptions::new().filename(&scratch_path);
+    let scratch_pool = SqlitePool::connect_with(scratch_options)
+        .await
+        .map_err(VerificationError::OpenScratch)?;
+
+    sqlx::migrate!()
+        .set_ignore_missing(true)
```
0xgleb, line 149: minor: verify_migrations is the deploy gate for exactly this bug class, but nothing in this module seeds a legacy ConversionConfirmed with filled_amount and then runs the full embedded migration chain. A regression test here would prove the repair migration plus replay path that prod depends on.

src/migration_verification.rs
```diff
@@ -0,0 +1,424 @@
...
+async fn clear_snapshots(pool: &SqlitePool) -> Result<(), VerificationError> {
+    sqlx::query("DELETE FROM snapshots")
+        .execute(pool)
+        .await
+        .map_err(VerificationError::ClearSnapshots)?;
+
+    Ok(())
+}
+
+async fn run_replay_checks(pool: &SqlitePool) -> Vec<AggregateReplayReport> {
```
0xgleb, line 179: minor: this aggregate list is maintained by hand. If someone adds a new EventSourced type and forgets to extend run_replay_checks, the deploy gate silently stops covering it. A registry or a unit test that asserts completeness would help.

infra/default.nix
```diff
@@ -222,6 +222,47 @@ let
...
+          trap _cleanup_identity EXIT
+
+          ssh_remote() {
+            # shellcheck disable=SC2029
+            ssh ''${identity:+-i "$identity"} "root@$host_ip" "$@"
+          }
+
+          db_path="/mnt/data/st0x-hedge.db"
+          remote_snapshot="/tmp/st0x-hedge-snapshot-$(date -u +%Y%m%d%H%M%S).db"
+          out_dir="''${1:-./claude-local-ctx/$(date -u +%Y-%m-%d_%H-%M-%S)-${env}-db-snapshot}"
```
0xgleb: nit: default snapshot output goes under claude-local-ctx, but AGENTS.md says generated tooling paths should live under .tmp/. Consider defaulting to .tmp/<timestamp>-${env}-db-snapshot unless the caller overrides.

0xgleb (approved):

tests/migrations.rs
JuaniRios: Good catch — the JSON-shape assertions here never proved the repaired event actually replays. Since these integration tests can't reach the `pub(crate)` `UsdcRebalance` / `load_entity`, I added the replay coverage where it belongs: a new `repair_migration_makes_legacy_conversion_confirmed_replay` test in `migration_verification.rs` (the deploy-gate module). It seeds a legacy `ConversionConfirmed` (`filled_amount`) stream, asserts `run_replay_checks` fails before the repair, then hydrates cleanly after. Fixed in 6452e924.

src/migration_verification.rs
JuaniRios: Done — added `repair_migration_makes_legacy_conversion_confirmed_replay` here: it seeds a legacy `ConversionConfirmed` with `filled_amount`, applies the repair migration, and runs `run_replay_checks` (the exact path the deploy gate uses), asserting replay fails pre-repair and succeeds post-repair. Fixed in 6452e924.

src/migration_verification.rs
JuaniRios: Agreed this is a real gap. A true completeness guard needs a deliberate registry (Rust has no reflection to enumerate `EventSourced` impls), which is broader than this fix's scope, so I filed it as tech debt: RAI-1214.

infra/default.nix
JuaniRios: Fixed — defaulted the snapshot output to `.tmp/<timestamp>-${env}-db-snapshot` per the AGENTS.md generated-paths convention. Fixed in 6452e924.

PR #1012, opened 2026-07-10
feat: sign Turnkey txs via SIGN_TRANSACTION so policies can gate them

JuaniRios:
## What

- Turnkey signer now signs the unsigned transaction via `SIGN_TRANSACTION` instead of the opaque hash via `SIGN_RAW_PAYLOAD` (`crates/evm/src/turnkey.rs`)
- Deleted the whole raw-signing path: `impl Signer`, `sign_hash`, `sign_raw_payload`, and the r/s/v parsing are gone
- Removed the `submit` operator CLI (`src/cli/submit.rs` + its wiring in `src/cli/mod.rs`) and the `Wallet::submit_raw` trait method it was the only caller of
- Dropped the dead `alloy-signer-turnkey` workspace dep (declared, used nowhere, not even in the lockfile)

## Why

- [RAI-1316](https://linear.app/makeitrain/issue/RAI-1316/migrate-turnkey-signer-to-sign-transaction-and-remove-the-raw-calldata): we want to turn on Turnkey policies, but a policy can only gate on `to`/`data`/`value`/selector when the activity is `SIGN_TRANSACTION` (the enclave parses the unsigned RLP server-side). Signing an opaque 32-byte hash means Turnkey sees nothing, so policies would gate nothing
- The `submit` CLI sent operator-supplied raw calldata to any address, the one untyped signing path in the binary, a bypass a policy can't inspect. If we're making everything policy-inspectable it has to go

## How

- `TxSigner::sign_transaction` hex-encodes `tx.encoded_for_signing()`, sends it as `SignTransactionIntentV2 { type: Ethereum }`, decodes the returned signed envelope with `TxEnvelope::decode_2718_exact`, and returns its `.signature()` so alloy's `WalletFiller` re-encodes as usual. Correctness holds cause Turnkey and alloy hash the same unsigned bytes
- Added a fail-closed check the issue didn't ask for but is the point of the migration: after decoding, recover the address from `keccak256(unsigned_rlp)` + the returned signature and require it equals our own. Without it we'd blindly trust whatever signature Turnkey hands back, which reopens the exact gap policies are meant to close. Binding to `keccak256(unsigned_rlp)` (the bytes we actually sent), not `tx.signature_hash()`, means a bad response fails loud (`SignerAddressMismatch`/`Rlp`) at runtime, never a bad on-chain signature
- Redacted the Turnkey response body from the trace log and the `Decode` error path, cause on a `SIGN_TRANSACTION` success it's a fully-signed broadcastable tx (a bearer instrument until we broadcast it), not the harmless r/s/v it used to be
- `submit_raw` was a trait-default method with a single caller and no impl overrides, so deleting it was clean. `send`/`send_pending`/`submit`/`submit_pending` stay, still used everywhere

## Testing

- New unit tests in `turnkey.rs`: happy path (asserts the request body shape + that the returned sig recovers to our address), wrong-key and wrong-content rejection (the content one proves the fail-closed check actually fires), legacy + EIP-1559 variants, both chain-id guards (set-before-encode, and mismatch errors with zero HTTP calls), `SignatureRecovery` on an unrecoverable sig, and that the `Decode` error redacts the body
- `cargo nextest run -p st0x-evm --all-features` green (81 tests), workspace `check`, `clippy --all-targets --all-features`, and `fmt` all clean

## Anything else

- One thing not covered in CI: the real Turnkey `SIGN_TRANSACTION` wire format (0x-prefix, EIP-155 preimage shape) is only pinned by the `#[ignore]`d `turnkey_integration` test, cause every CI test builds its Turnkey fixture with a local signer, so it can't catch a mismatch with the real API. The fail-closed recover-address check means a wrong assumption errors loud at staging, it can't produce a bad signature, but before this ships run `turnkey_integration` once against a real Turnkey org. Follow-up to automate this in CI (golden fixture or scheduled job) tracked separately
- Policy definitions themselves live in Turnkey org config, not this repo. This PR just makes the bot's transactions policy-inspectable
- Heads up for whoever writes the policy: the startup `approve(U256::MAX)` (`src/onchain/approvals.rs`) and the CCTP standing allowance mean a policy can pin selector + spender but not amount. Pre-existing, no change here, just worth knowing


JuaniRios:
* **#1012** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1012?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1012?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1012).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/evm/src/turnkey.rs
```diff
@@ -784,80 +820,284 @@ mod tests {
...
+        });
+
+        let signer = TracingTurnkeySigner::new(
+            mock_client(&server),
+            TurnkeyOrganizationId::new("org-test".to_string()),
+            address,
+            Some(chain_id),
+        );
+
+        TxSigner::sign_transaction(&signer, &mut tx).await.unwrap();
```
0xgleb, line 1035: minor: this discards the returned Signature, and the encoded_for_signing() assert below is close to a tautology because chain_id is already Some, so signing mutates nothing the assert covers. The happy path never checks the returned value; it leans on parse_signature's internal recovery check to panic the unwrap. Capturing the signature and asserting it equals the one locally_signed_envelope_hex produced, or that it recovers to `address` over the unsigned RLP, would pin the actual return value.

0xgleb (approved):

PR #1013, opened 2026-07-10
fix: clear USDC rebalance guard even when inflight release underflows

JuaniRios:
## What

- Terminal USDC transfer events now clear the trigger guard, tracking entry, and active-rebalance id even when the in-memory inflight release underflows (`InsufficientInflight`), so a transfer that already settled on-chain never latches USDC rebalancing off.
- New `UsdcSettlementOutcome` lets `on_usdc_rebalance` skip the immediate imbalance check when a terminal event left the in-memory ledger unreconciled, so it can't fire a duplicate transfer against stale inventory.

## Why

- Fixes [RAI-1322](https://linear.app/makeitrain/issue/RAI-1322/usdc-terminal-success-cleanup-skips-guardtracking-clear-when-inflight).
- In prod a transfer that had settled on-chain hit `InsufficientInflight` in `settle_transfer` (the inflight was never reserved after a mid-transfer restart). The reactor bailed on the `?` before cleanup ran, so `usdc_in_progress` stayed latched, the sweep re-selected the stale tracking entry every tick and paged a false "timed out after CCTP burn" alert, and USDC rebalancing was stuck until a bot restart.

## How

- `complete_usdc_rebalance` and `cancel_tracked_usdc_rebalance` catch `InsufficientInflight`, `warn!` with full context, zero any residual source inflight via `clear_usdc_inflight` (a leftover positive inflight keeps `has_inflight()` true, which blocks every future snapshot from healing the ledger), and return `DeferredToSnapshot`. Every other error still propagates, so `SettledUsdcExceedsInitiatedAmount` keeps preserving state.
- `apply_usdc_rebalance_event` is now an expression-match where every arm evaluates to a `UsdcSettlementOutcome`, so a new terminal arm can't silently fall through to the old implicit default. `OperatorReconciled` returns `DeferredToSnapshot`, cause it zeroes source inflight without crediting the destination.
- `on_usdc_rebalance` enqueues a fresh imbalance check only on `Reconciled`; on `DeferredToSnapshot` it defers to the next snapshot poll.

## Testing

- Regression tests: unreserved inflight (zero and partial) on the success and `WithdrawalFailed` cancel paths clears the guard/tracking/active id and zeroes residual source inflight; the generic-error arm still hard-fails and keeps the guard latched; every `DeferredToSnapshot` path (missing tracking, underflow, `OperatorReconciled`) leaves 0 pending imbalance-check jobs while a reserved settlement enqueues exactly 1.
- `cargo check`, `clippy`, and `nextest -p st0x-hedge` all green.

## Anything else

- The root cause (a resumed AlpacaToBase transfer never rebuilds its source inflight reservation) and two residual stale-inventory-window risks (per-venue USDC snapshots heal one venue at a time, and a guard-clear vs `cancel_pending` TOCTOU) are deferred to [RAI-1324](https://linear.app/makeitrain/issue/RAI-1324/alpacatobase-resume-from-statenone-does-not-reconstruct-inflight). This PR keeps a confirmed transfer from wedging rebalancing; RAI-1324 removes the desync at the root.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USDC rebalancing settlement handling to avoid duplicate imbalance checks.
  * Deferred fresh imbalance evaluation until the next snapshot when inventory reconciliation can’t complete immediately.
  * Enhanced terminal cleanup for deposits, conversions, and failed withdrawals, including better handling of unreserved/partially reserved inflight.
  * Improved recovery by clearing stale in-progress state and restoring healing in subsequent snapshots.
  * Preserved active rebalance state for unrecoverable destination overflow errors.
* **Tests**
  * Added/extended regression coverage for terminal USDC settlement paths, including Alpaca-to-base and deferral behavior.

### DIFF
--- a/src/rebalancing/trigger/usdc.rs
+++ b/src/rebalancing/trigger/usdc.rs
@@ -13,7 +13,8 @@ use st0x_finance::{Usd, Usdc};
 use super::{RebalancingService, RebalancingServiceError};
 use crate::conductor::job::{Job, JobQueue, Label, QueuePushError};
 use crate::inventory::{
-    BroadcastingInventory, Imbalance, ImbalanceThreshold, Inventory, TransferOp, Venue,
+    BroadcastingInventory, Imbalance, ImbalanceThreshold, Inventory, InventoryError,
+    InventoryViewError, TransferOp, Venue,
 };
 use crate::usdc_rebalance::{RebalanceDirection, UsdcRebalanceEvent, UsdcRebalanceId};
 
@@ -140,6 +141,28 @@ pub(super) enum UsdcTerminalAction {
     NotTerminal,
     Clear,
     PreservePostBurn,
+}
+
+/// Whether a terminal USDC settlement event reconciled the in-memory
+/// inventory ledger. `DeferredToSnapshot` means the settlement was accepted
+/// as authoritative (the durable event proves funds arrived) but the
+/// in-memory available bookkeeping was skipped -- either because a terminal
+/// *success* arrived with no tracking context to reconcile against (a
+/// post-restart resume where the transfer already settled), or because
+/// reconciliation underflowed (the resume desync that fails to reconstruct
+/// inflight). The no-tracking clause is terminal-success-only: a pre-burn
+/// terminal *failure* with no tracking context reconciles as `Reconciled`
+/// instead, because nothing moved, so an immediate imbalance check reads data
+/// as fresh as a post-snapshot one (see `cancel_tracked_usdc_rebalance`). In
+/// the underflow case, the source venue's residual inflight is zeroed so
+/// `has_inflight()` can clear and future snapshots resume healing; the
+/// available balances on both sides are left for the snapshot poll to
+/// correct. Callers must not act on `self.inventory` as if it reflects this
+/// settlement until the next snapshot poll heals it.
+#[derive(Debug, Clone, Copy)]
+pub(super) enum UsdcSettlementOutcome {
+    Reconciled,
+    DeferredToSnapshot,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -572,7 +595,8 @@ impl RebalancingService {
         }
 
         let terminal_action = self.usdc_terminal_action(&id, &event).await;
-        self.apply_usdc_rebalance_event(&id, &event, terminal_action)
+        let settlement_outcome = self
+            .apply_usdc_rebalance_event(&id, &event, terminal_action)
             .await?;
 
         let is_clearable_terminal = terminal_action == UsdcTerminalAction::Clear;
@@ -606,7 +630,25 @@ impl RebalancingService {
         if is_clearable_terminal {
             self.equity_scheduler.cancel_pending().await;
             self.usdc_scheduler.cancel_pending().await;
-            self.usdc_scheduler.enqueue_check().await;
+            match settlement_outcome {
+                UsdcSettlementOutcome::Reconciled => {
+                    self.usdc_scheduler.enqueue_check().await;
+                }
+                UsdcSettlementOutcome::DeferredToSnapshot => {
+                    // The in-memory ledger was not reconciled by this terminal
+                    // event (see UsdcSettlementOutcome::DeferredToSnapshot), so
+                    // checking imbalance now would read stale inventory and could
+                    // dispatch a duplicate transfer for funds that already
+                    // arrived. Defer the fresh check to the next periodic
+                    // snapshot poll, which heals the ledger.
+                    debug!(
+                        target: "rebalance",
+                        id = %id,
+                        "Deferring fresh USDC imbalance check to next snapshot poll: \
+                         in-memory inventory was not reconciled by this terminal event"
+                    );
+                }
+            }
         }
 
         Ok(())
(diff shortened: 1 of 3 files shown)


JuaniRios:
* **#1013** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1013?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1013?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1013).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/rebalancing/trigger/usdc.rs
```diff
@@ -1040,25 +1085,59 @@ impl RebalancingService {
...
+                    actual_inflight = ?inflight,
+                    "USDC source inflight underflow on terminal cancel; treating \
+                     the durable failure event as authoritative, zeroing residual \
+                     source inflight so snapshots can heal, and skipping in-memory \
+                     available reconciliation (heals on next snapshot poll)."
+                );
+                *inventory = inventory.clone().clear_usdc_inflight(source_venue, now)?;
+                UsdcSettlementOutcome::DeferredToSnapshot
+            }
+            Err(error) => return Err(error.into()),
```
0xgleb, line 1126: minor: this generic-error arm has no test, while its settle-path twin got `deposit_confirmed_with_destination_overflow_hard_fails_and_leaves_guard_latched`. `cancel_inflight` releases inflight and then does `available + amount`, which can overflow with a generic `Arithmetic` error when `available` is near Float's max, so the arm is reachable the same way. A companion test driving `WithdrawalFailed` with source `available` seeded at `max_positive_value` and inflight reserved would lock down the hard-fail-and-stay-latched invariant on this path too.

src/rebalancing/trigger/usdc.rs
```diff
@@ -142,6 +143,23 @@ pub(super) enum UsdcTerminalAction {
...
+/// inventory ledger. `DeferredToSnapshot` means the settlement was accepted
+/// as authoritative (the durable event proves funds arrived) but the
+/// in-memory available bookkeeping was skipped -- either because no tracking
+/// context existed to reconcile against, or because reconciliation
+/// underflowed (the resume desync that fails to reconstruct inflight). In the
+/// underflow case, the source venue's residual inflight is zeroed so
+/// `has_inflight()` can clear and future
+/// snapshots resume healing; the available balances on both sides are left
+/// for the snapshot poll to correct. Callers must not act on `self.inventory`
+/// as if it reflects this settlement until the next snapshot poll heals it.
```
0xgleb: minor: this says `DeferredToSnapshot` covers the case where "no tracking context existed to reconcile against", but `cancel_tracked_usdc_rebalance` returns `Reconciled` from its missing-tracking and `!source_transfer_started()` branches. That asymmetry looks right (a terminal success proves reality changed while the ledger provably lags, whereas a pre-burn failure means nothing moved, so an immediate check reads data as fresh as a post-snapshot one and lets the still-present imbalance retry), but the doc as written doesn't scope the no-tracking clause to terminal successes. Worth tightening so a future terminal arm following this doc literally doesn't pick the wrong variant.

src/rebalancing/trigger/usdc.rs
```diff
@@ -142,6 +143,23 @@ pub(super) enum UsdcTerminalAction {
...
+#[derive(Debug, Clone, Copy)]
+pub(super) enum UsdcSettlementOutcome {
```
0xgleb, line 163: minor: the doc scopes this type to terminal settlement events, but the exhaustive match in `apply_usdc_rebalance_event` forces every non-terminal arm to return `Reconciled` as a placeholder. That's only safe because `on_usdc_rebalance` consults the value behind `is_clearable_terminal`, a separate classifier with no structural link to this type. Something like an outcome enum with a `NonTerminal` case (or `Option<UsdcSettlementOutcome>`) would make terminal-only-ness compiler-checked instead of convention.

0xgleb (approved):

src/rebalancing/trigger/usdc.rs
JuaniRios: Good catch -- added `withdrawal_failed_with_source_credit_overflow_hard_fails_and_leaves_guard_latched` as the cancel-path twin. It seeds Hedging with both available and inflight at Float max so `cancel_inflight` succeeds (max - max = 0) but the credit-back (max + max) overflows into a generic `Arithmetic` error, and asserts the guard/tracking/active-id all stay latched. Needed a small `with_usdc_inflight` seeding helper since no single event produces the inflight-present-but-available-not-debited resume desync. Fixed in cec1d721.

src/rebalancing/trigger/usdc.rs
JuaniRios: Agreed, tightened the doc: the "no tracking context" -> `DeferredToSnapshot` clause is now explicitly scoped to terminal *successes*, and it notes that a pre-burn terminal *failure* with no tracking reconciles as `Reconciled` instead (with a pointer to `cancel_tracked_usdc_rebalance`). Fixed in cec1d721.

src/rebalancing/trigger/usdc.rs
JuaniRios: Fair point -- the terminal-only-ness is convention, not compiler-checked. I looked at `Option<UsdcSettlementOutcome>` / a `NonTerminal` variant, but since `is_clearable_terminal` is derived independently of the outcome, that just relocates the convention (the caller's terminal branch keeps a "can't happen" arm). The clean fix is unifying the terminal classifier with the outcome so one classification yields both -- bigger than this bugfix's scope, so I filed RAI-1331 to track it: https://linear.app/makeitrain/issue/RAI-1331

PR #1023, opened 2026-07-14
feat: hold pre-send redemption resumes while the asset is frozen

0xgleb:
## What

[RAI-1139](https://linear.app/makeitrain/issue/RAI-1139)

Add the dividend freeze gate to automated redemption recovery and startup resume paths. Uncommitted redemptions hold while frozen or while status is unknown; `SendPending` and later states continue to completion.

## Why

The automatic rebalance trigger already respected the dividend freeze, but recovery jobs could bypass it and send a frozen asset to issuance’s redemption wallet. Conversely, stopping after durable send intent could strand a committed transfer. The gate therefore has to match the workflow’s commitment boundary rather than physical transfer completion.

## How

- Classify states before durable send intent as safe to hold.
- Treat `SendPending` as committed and allow it and all later states to continue.
- Fail closed when issuance freeze status cannot be confirmed.
- Reuse one `FreezeStatusReader` for trigger, recovery, and startup wiring.
- Install the reader before CQRS catch-up so replay cannot emit an ungated trigger.
- Leave held aggregates recoverable so later inventory polling can re-check status.

## Testing

Added coverage for:

- frozen and unknown-status holds before intent;
- disabled-guard behavior;
- `SendPending` continuation; and
- continuation through later send/completion states.

## Screenshots

Not applicable.

## Anything else

This is liquidity-side defense in depth. Issuance remains the authoritative supply lock; liquidity avoids initiating an unsafe transfer while ensuring a durably committed send is not abandoned.

### DIFF
--- a/src/conductor.rs
+++ b/src/conductor.rs
@@ -121,7 +121,7 @@ use crate::wrapped_equity_recovery::{
 };
 
 pub(crate) use builder::CqrsFrameworks;
-use manifest::QueryManifest;
+use manifest::{BuiltFrameworks, QueryManifest};
 use trading_queues::{EquityRecoveryInputs, TradingJobQueues, setup_trading_job_queues};
 
 /// CQRS/event-store and apalis worker pools over the same SQLite database.
@@ -1273,17 +1273,21 @@ impl PositionAndRebalancing {
     }
 }
 
-/// Wires the dividend freeze guard onto the rebalancing service when enabled.
+/// Wires the dividend freeze guard onto the rebalancing service before its
+/// CQRS frameworks are built, so catch-up cannot observe an unwired fail-open
+/// trigger. Returns the shared reader for the redemption-resume gate, which is
+/// wired after the transfer service's stores exist.
 ///
-/// When disabled, the trigger's `freeze_status` stays `None` and the equity gate
-/// fails open -- an operator escape hatch for an issuance outage that would
-/// otherwise fail closed every equity cycle. Issuance secrets remain mandatory
-/// either way, so re-enabling is a pure plaintext-config change.
-async fn wire_freeze_guard(
+/// When disabled, both `freeze_status` readers stay `None` and the gates fail
+/// open -- an operator escape hatch for an issuance outage that would
+/// otherwise fail closed every equity cycle and hold every redemption resume.
+/// Issuance secrets remain mandatory either way, so re-enabling is a pure
+/// plaintext-config change.
+async fn wire_rebalancing_freeze_guard(
     rebalancing_service: &RebalancingService,
     freeze_check: OperationMode,
     issuance: &IssuanceStatusCtx,
-) -> anyhow::Result<()> {
+) -> anyhow::Result<Option<Arc<IssuanceClient>>> {
     match freeze_check {
         OperationMode::Enabled => {
             let reader = Arc::new(IssuanceClient::new(
@@ -1291,18 +1295,20 @@ async fn wire_freeze_guard(
                 issuance.api_key.header_value(),
             )?);
 
-            rebalancing_service.set_freeze_status_reader(reader).await;
+            rebalancing_service
+                .set_freeze_status_reader(reader.clone())
+                .await;
+            Ok(Some(reader))
         }
         OperationMode::Disabled => {
             warn!(
                 target: "rebalance",
-                "Dividend freeze guard is disabled by config; equity rebalancing will \
-                 proceed without consulting issuance"
+                "Dividend freeze guard is disabled by config; equity rebalancing and \
+                 redemption resumes will proceed without consulting issuance"
             );
+            Ok(None)
         }
     }
-
-    Ok(())
 }
 
 /// Builds the rebalancing [`RaindexService`]: writes settle through the shared
@@ -1469,32 +1475,20 @@ fn spawn_rebalancing_infrastructure<Chain: Wallet + Clone>(
             usdc_notifier.clone(),
         ));
 
-        wire_freeze_guard(
+        let freeze_status_reader = wire_rebalancing_freeze_guard(
             &rebalancing_service,
             rebalancing_ctx.freeze_check,
             &deps.ctx.issuance,
         )
         .await?;
 
-        let broadcaster = Arc::new(Broadcaster::new(deps.event_sender, deps.pool.clone()));
-        let hedge_latency = HedgeLatencyProjection::new(deps.pool.clone());
-        let rebalance_timing = RebalanceTimingProjection::new(deps.pool.clone());
-        let equity_timing = EquityTimingProjection::new(deps.pool.clone());
-        let lifecycle_failure = LifecycleFailureProjection::new(deps.pool.clone());
-        catch_up_stage_timing_projections(&rebalance_timing, &equity_timing).await?;
-
-        let manifest = QueryManifest::new(
+        let built = build_rebalancing_frameworks(
+            &deps.pool,
+            deps.event_sender,
             rebalancing_service.clone(),
-            broadcaster,
-            hedge_latency,
-            rebalance_timing,
-            equity_timing,
-            lifecycle_failure,
-        );
-
-        let built = manifest
-            .build(deps.pool.clone(), equity_transfer_services)
-            .await?;
+            equity_transfer_services,
+        )
+        .await?;
 
         rebalancing_service
             .recover_pending_offchain_order_symbols(&built.position_projection)
@@ -1517,6 +1511,10 @@ fn spawn_rebalancing_infrastructure<Chain: Wallet + Clone>(
             built.mint.clone(),
             built.redemption.clone(),
         ));
+
+        if let Some(reader) = freeze_status_reader {
+            recovery_transfer.set_freeze_status_reader(reader).await;
+        }
 
         // Built outside `QueryManifest`: services depend on mint/redemption stores
         // that the manifest produces.
@@ -1638,6 +1636,34 @@ fn spawn_rebalancing_infrastructure<Chain: Wallet + Clone>(
             resume_tokenization_queue,
         })
     })
+}
+
+/// Wires the rebalancing read models and builds the CQRS frameworks: catches
+/// the stage-timing projections up to the event log, then hands every query
+/// processor to [`QueryManifest`] so a new one cannot be forgotten.
+async fn build_rebalancing_frameworks(
+    pool: &SqlitePool,
+    event_sender: broadcast::Sender<Statement>,
+    rebalancing_service: Arc<RebalancingService>,
+    equity_transfer_services: EquityTransferServices,
+) -> anyhow::Result<BuiltFrameworks> {
+    let broadcaster = Arc::new(Broadcaster::new(event_sender, pool.clone()));
+    let hedge_latency = HedgeLatencyProjection::new(pool.clone());
+    let rebalance_timing = RebalanceTimingProjection::new(pool.clone());
+    let equity_timing = EquityTimingProjection::new(pool.clone());
+    let lifecycle_failure = LifecycleFailureProjection::new(pool.clone());
+    catch_up_stage_timing_projections(&rebalance_timing, &equity_timing).await?;
+
+    let manifest = QueryManifest::new(
+        rebalancing_service,
+        broadcaster,
+        hedge_latency,
+        rebalance_timing,
+        equity_timing,
+        lifecycle_failure,
+    );
+
+    manifest.build(pool.clone(), equity_transfer_services).await
 }
 
 /// Catches the lifecycle-failure read model up to the event log before its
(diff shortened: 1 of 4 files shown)


0xgleb:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1023?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1023** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1023?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1023?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1030** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1030?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1024](https://github.com/ST0x-Technology/st0x.liquidity/pull/1024) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1024?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1023).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/equity_redemption.rs
```diff
@@ -1061,6 +1061,49 @@ impl EquityRedemption {
...
+    /// Exhaustive `match` so a new variant forces re-classification.
+    pub(crate) fn is_pre_send(&self) -> bool {
+        match self {
+            Self::VaultWithdrawPending { .. }
+            | Self::VaultWithdrawSubmitted { .. }
+            | Self::WithdrawnFromRaindex { .. }
+            | Self::UnwrapPending { .. }
+            | Self::UnwrapSubmitted { .. }
+            | Self::TokensUnwrapped { .. } => true,
+            Self::SendPending { .. }
```
JuaniRios, line 1089: critical: `SendPending` is on the wrong side of this boundary, so the gate is skipped at exactly the send it exists to prevent. `SendPending` is the state BEFORE the external transfer. `PrepareSend` is a pure state transition with no side effect, and `transition_send_tokens` is what calls `Tokenizer::send_for_redemption(...)`. It runs only when the aggregate is in `SendPending` and the `SendTokens` command is dispatched. `TokensSent` is emitted after that call returns a transaction hash.

`resume_redemption` calls `held_by_dividend_freeze` only when `is_pre_send()` is true. So an aggregate that is durably in `SendPending` (a crash between the two dispatches in `unwrap_and_send`, a startup resume, or an operator `stox transfer recheck`) dispatches `SendTokens` with no freeze check, and sends tokens to the issuer's redemption wallet for an asset that may have frozen since the last check at `TokensUnwrapped`.

The new test `resume_redemption_completes_send_intent_despite_freeze` locks this in: it seeds `SendPending`, sets the reader to `Frozen`, resumes, and asserts `Completed`. The doc argues the check cannot be atomic with the send, which is true, but that unavoidable window is narrow and separate from skipping a check that a resume could perform. The two sibling states with the same shape, `VaultWithdrawPending` and `UnwrapPending`, are both classified as pre-send. Please move `SendPending` into the pre-send branch and make `TokensSent` the commitment boundary, then update that test to assert the redemption holds at `SendPending` while frozen, and add one asserting it completes when resumed from `TokensSent`.

src/rebalancing/equity/mod.rs
```diff
@@ -1146,13 +1168,74 @@ impl CrossVenueEquityTransfer {
...
+            let entity = self.load_redemption_entity(aggregate_id).await?;
+
+            // States before the durable `SendPending` intent hold during a
+            // dividend freeze. The freeze check cannot be atomic with the
+            // external token send, so the persisted intent is the stable
+            // commitment boundary: from `SendPending` onward the redemption
+            // MUST complete even if a freeze begins after that intent. Holding
+            // returns Ok; inventory-poll recovery re-drives the aggregate and
+            // re-checks the freeze on each attempt.
+            if entity.is_pre_send()
```
JuaniRios, line 1566: should fix: this gate covers only the resume branch. `resume_equity_to_hedging` is the `TransferEquityToHedging` job entry point and the path re-entered after a restart or an apalis retry. Its `None` branch runs `start_redemption` and then `unwrap_and_send`, which dispatches `PrepareSend` and `SendTokens` with no freeze check at all. The trigger's own gate is evaluated at enqueue time, not execution time, and the apalis row is durable. So a job enqueued while the asset was enabled and executed after the freeze began (crash before the first `Redeem` event, then orphan-requeued at startup) sends tokens for a frozen asset. The same hole exists for a fresh redemption whose asset freezes while the job waits on the vault withdraw or the unwrap. This is the case where holding costs nothing, because nothing is committed yet. Please move the check above the branch: call `held_by_dividend_freeze(aggregate_id, symbol)` in `resume_equity_to_hedging` before it decides between start and resume, and add a test that a fresh redemption holds while frozen.

src/rebalancing/equity/mod.rs
```diff
@@ -1146,13 +1168,74 @@ impl CrossVenueEquityTransfer {
...
+                && self
+                    .held_by_dividend_freeze(aggregate_id, entity.symbol())
+                    .await
+            {
+                return Ok(());
```
JuaniRios, line 1571: should fix: holding returns `Ok(())` without persisting a held state, refreshing tracking, or keeping a retry driver alive. `TransferEquityToHedging` and `ResumeTokenizationAggregate` treat `Ok(())` as job success, so the apalis row can go terminal while the aggregate stays in `TokensUnwrapped` or earlier. Startup recovery then reconstructs it as `ActiveTransfer` with `last_progress_at` taken from the old event, so the timeout sweep can emit `FailTransfer` for a redemption that this gate deliberately held. That contradicts the PR's claim that held aggregates stay recoverable. If the terminal row is a `TransferEquityToHedging` row, restart recovery also skips the generic resume, because `load_transfer_jobs` treats terminal transfer rows as ownership. Please return a distinct held outcome instead of ordinary success, have the jobs re-enqueue with delay without spending the failure budget, and exclude a held aggregate from the transfer timeout. A test that holds a `TokensUnwrapped` redemption across `transfer_timeout` and then unfreezes would pin it.

src/rebalancing/equity/mod.rs
```diff
@@ -1146,13 +1168,74 @@ impl CrossVenueEquityTransfer {
...
+                    target: "rebalance",
+                    %aggregate_id,
+                    %symbol,
+                    "Holding pre-send redemption resume: asset is frozen for a \
+                     dividend; it will resume after unfreeze"
+                );
+                true
+            }
+            Err(error) => {
+                // Fail closed: sending tokens for a frozen asset strands them
```
JuaniRios, line 1534: should fix: this rationale is not backed by the issuance contract this repo pins. The comment says sending tokens for a frozen asset strands them in issuance's redemption wallet until unfreeze, and `is_pre_send`'s doc repeats it. The SPEC for the pinned issuance revision (`st0x-issuance-client` / `st0x-issuance-dto` at tag 0.2.0) says the opposite under "Freeze invariant, frozen is not de-listed": freezing gates new mints only, a frozen asset stays in `list_enabled_assets()`, and in-flight redemption detection continues. The gate fires on already-started redemptions, and in `WithdrawnFromRaindex`, `UnwrapPending`, `UnwrapSubmitted`, and `TokensUnwrapped` the tokens have already left the vault. If issuance keeps settling redemptions while frozen, holding here strands value in our own wallet rather than preventing a strand. Please cite the specific issuance SPEC section or redemption handler in the `held_by_dividend_freeze` doc, or reconcile the two documents before shipping.

src/equity_redemption.rs
```diff
@@ -1061,6 +1061,49 @@ impl EquityRedemption {
...
+            | Self::Pending { .. }
+            | Self::Completed { .. }
+            | Self::Failed { .. }
+            | Self::Reconciled { .. } => false,
+        }
+    }
+
+    /// The symbol this redemption operates on; every lifecycle state carries
+    /// it.
+    pub(crate) fn symbol(&self) -> &Symbol {
```
JuaniRios, line 1100: should fix: `is_pre_send()` and `symbol()` are new exhaustive matches over all 12 variants, and no test exercises them directly. The only coverage is indirect, through the freeze-guard integration tests, which drive a redemption through `TokensUnwrapped`, `SendPending`, and `TokensSent` only. `VaultWithdrawPending`, `VaultWithdrawSubmitted`, `WithdrawnFromRaindex`, `UnwrapPending`, `UnwrapSubmitted`, `Pending`, `Completed`, `Failed`, and `Reconciled` are never checked. `is_pre_send()` is the sole gate deciding whether the freeze fail-safe applies, so misclassifying any of them either strands a safe redemption or sends a frozen one. Please add a table-driven unit test in this file's test module that builds one instance of each variant and asserts both classifiers against the documented boundary. It is cheap and needs no async setup.

src/rebalancing/equity/mod.rs
```diff
@@ -1146,13 +1168,74 @@ impl CrossVenueEquityTransfer {
...
+            // MUST complete even if a freeze begins after that intent. Holding
+            // returns Ok; inventory-poll recovery re-drives the aggregate and
+            // re-checks the freeze on each attempt.
+            if entity.is_pre_send()
+                && self
+                    .held_by_dividend_freeze(aggregate_id, entity.symbol())
+                    .await
+            {
+                return Ok(());
+            }
```
JuaniRios, line 1572: should fix: `recover_redemption`'s catch-all arm calls `resume_redemption(...)?` and then returns `RecheckOutcome::Resumed` unconditionally. With this gate, a frozen pre-send redemption returns `Ok(())` without moving, so `stox transfer recheck --kind redemption` reports `Resumed` for an aggregate that did nothing. The HTTP recovery endpoint has the same problem: it counts a failure only on `Err`, so a batch of frozen redemptions reports zero failures. These are the commands the runbooks point an operator at when a transfer looks stuck, so this sends them chasing a phantom success during an incident. Only a log line records the real reason. Once `resume_redemption` returns a distinguishable hold outcome, please map it to a new `RecheckOutcome` variant and surface it as its own counter in the resume response.

src/conductor.rs
```diff
@@ -3683,33 +3709,74 @@ mod tests {
...
+            wrapper,
+            Address::ZERO,
+            mint_store,
+            redemption_store,
+        )
+    }
+
     #[tokio::test]
-    async fn wire_freeze_guard_enabled_wires_the_reader() {
+    async fn wire_rebalancing_freeze_guard_enabled_returns_reader_for_resume_gate() {
```
JuaniRios, line 4800: should fix: this test calls `wire_rebalancing_freeze_guard()` directly and then calls `set_freeze_status_reader(reader)` itself. That replays the wiring rather than exercising it. The `if let` inside `spawn_rebalancing_infrastructure` that this PR added is never executed by any test, and nothing confirms the same reader instance reaches both the trigger service and `CrossVenueEquityTransfer`. That plumbing is the actual risk surface of the conductor change, because the reader is created early and threaded into a component built later. Please drive `spawn_rebalancing_infrastructure` end to end, or extract the wiring block into a small function that both production and the test call, and assert the built transfer has the reader.

src/rebalancing/equity/mod.rs
```diff
@@ -1146,13 +1168,74 @@ impl CrossVenueEquityTransfer {
...
     #[allow(clippy::cognitive_complexity)]
     pub(crate) async fn resume_redemption(
         &self,
         aggregate_id: &RedemptionAggregateId,
     ) -> Result<(), RedemptionError> {
         loop {
-            match self.load_redemption_entity(aggregate_id).await? {
+            let entity = self.load_redemption_entity(aggregate_id).await?;
+
+            // States before the durable `SendPending` intent hold during a
```
JuaniRios, line 1559: minor: these seven lines restate the rationale already documented on `EquityRedemption::is_pre_send`, added in this same PR, in near-identical wording. The boundary rationale now lives in two places that must be kept in sync by hand, and call-site prose like this predictably drifts. `is_pre_send`'s name and doc already carry the why at its definition. Please keep only what the doc does not cover: that holding returns `Ok` and that inventory-poll recovery re-drives and re-checks later.

JuaniRios (commented):
SendPending is classified as post-commit, but it is the state before send_for_redemption runs, so a redemption resumed from it sends tokens for a frozen asset with no check. The new test locks that in as intended. Separately the fresh-start branch of the job entry point is not gated at all, holds are reported as ordinary job success and can later be timed out, and the rationale rests on an issuance behavior the pinned issuance SPEC contradicts.

JuaniRios (changes requested):
Requesting changes. There is a correctness defect at the heart of the gate:

- `is_pre_send()` classifies `SendPending` as post-commit, but `SendPending` is the state before the external send. `PrepareSend` has no side effect; `transition_send_tokens` is what calls `send_for_redemption`. So a redemption resumed from `SendPending` (crash between the two dispatches, startup resume, or `stox transfer recheck`) dispatches `SendTokens` with no freeze check and sends tokens for a frozen asset. `resume_redemption_completes_send_intent_despite_freeze` locks that in as intended behavior. The commitment boundary should be `TokensSent`.

Also blocking: the fresh-start branch of `resume_equity_to_hedging` is not gated at all; a hold is returned as ordinary job success, so the row can go terminal and the timeout sweep can later fail a deliberately held redemption; and the rationale rests on an issuance behavior that the pinned issuance SPEC contradicts under "frozen is not de-listed".

PR #1024, opened 2026-07-14
feat: fail-closed dividend freeze gate on manual redemption commands

0xgleb:
## What

[RAI-1139](https://linear.app/makeitrain/issue/RAI-1139)

Add a fail-closed dividend freeze gate to manual redemption-side CLI commands, with an explicit operator `--force` override.

## Why

`transfer-equity --direction to-alpaca` and `alpaca-redeem` were the remaining paths that could send a frozen asset to issuance’s redemption wallet. An unavailable status service must not silently permit that transfer, while incident response still needs a deliberate audited bypass.

## How

- Query issuance freeze status before either command initiates redemption.
- Refuse both frozen assets and unconfirmable status by default.
- Represent enforcement versus override explicitly through `FreezeGateMode` after argument parsing.
- Make `--force` bypass the gate without performing an issuance API call.
- Log and display the override so operators can identify forced execution.
- Leave mint-side commands unchanged because issuance rejects frozen-asset mints before stranded funds are possible.

## Testing

Added coverage for:

- frozen and unfrozen assets;
- issuance status errors failing closed;
- forced execution succeeding; and
- `--force` making no freeze-status API request.

## Screenshots

Not applicable.

## Anything else

The override is intended only for incident response. Normal command execution remains fail closed.

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -27,6 +27,7 @@ use st0x_execution::{AlpacaAccountId, Direction, FractionalShares, Positive, Sym
 use st0x_finance::Usdc;
 use st0x_registry::SymbolCache;
 
+use self::rebalancing::FreezeGateMode;
 use crate::offchain::order::{OffchainOrder, OffchainOrderId, OrderPlacer};
 use crate::performance::equity_timing::EquityTimingProjection;
 use crate::performance::rebalance::RebalanceTimingProjection;
@@ -302,6 +303,11 @@ pub enum Commands {
         /// Alpaca redemption wallet (overrides [tokenization] config)
         #[arg(long = "redemption-wallet")]
         redemption_wallet: Option<Address>,
+        /// Deliberately bypass the dividend freeze gate on to-alpaca (logged).
+        /// Without this, a frozen or unconfirmable asset refuses the
+        /// redemption-side send.
+        #[arg(long = "force")]
+        force: bool,
     },
 
     /// Wrap tokenized equity into wrapped ERC-4626 vault shares
@@ -604,6 +610,10 @@ pub enum Commands {
         /// Alpaca redemption wallet (overrides [tokenization] config)
         #[arg(long = "redemption-wallet")]
         redemption_wallet: Option<Address>,
+        /// Deliberately bypass the dividend freeze gate (logged). Without
+        /// this, a frozen or unconfirmable asset refuses the redemption send.
+        #[arg(long = "force")]
+        force: bool,
     },
 
     /// Convert USDC to/from USD on Alpaca
@@ -886,6 +896,7 @@ enum SimpleCommand {
         quantity: FractionalShares,
         issuer_request_id: Option<Uuid>,
         redemption_wallet: Option<Address>,
+        freeze_gate: FreezeGateMode,
     },
     WrapEquity {
         symbol: Symbol,
@@ -1133,6 +1144,7 @@ enum ProviderCommand {
         symbol: Symbol,
         quantity: FractionalShares,
         redemption_wallet: Option<Address>,
+        freeze_gate: FreezeGateMode,
     },
     DividendBump {
         symbol: Symbol,
@@ -1233,12 +1245,14 @@ fn classify_command(command: Commands) -> Result<SimpleCommand, ProviderCommand>
             quantity,
             issuer_request_id,
             redemption_wallet,
+            force,
         } => Ok(SimpleCommand::TransferEquity {
             direction,
             symbol,
             quantity,
             issuer_request_id,
             redemption_wallet,
+            freeze_gate: FreezeGateMode::from_force_flag(force),
         }),
         Commands::WrapEquity { symbol, quantity } => {
             Ok(SimpleCommand::WrapEquity { symbol, quantity })
@@ -1313,10 +1327,12 @@ fn classify_command(command: Commands) -> Result<SimpleCommand, ProviderCommand>
             symbol,
             quantity,
             redemption_wallet,
+            force,
         } => Err(ProviderCommand::AlpacaRedeem {
             symbol,
             quantity,
             redemption_wallet,
+            freeze_gate: FreezeGateMode::from_force_flag(force),
         }),
         Commands::DividendBump { symbol, quantity } => {
             Err(ProviderCommand::DividendBump { symbol, quantity })
@@ -1451,14 +1467,18 @@ async fn run_simple_command<W: Write>(
             quantity,
             issuer_request_id,
             redemption_wallet,
+            freeze_gate,
         } => {
             rebalancing::transfer_equity_command(
                 stdout,
-                direction,
-                &symbol,
-                quantity,
-                issuer_request_id,
-                redemption_wallet,
+                rebalancing::TransferEquityArgs {
+                    direction,
+                    symbol,
+                    quantity,
+                    issuer_request_id,
+                    redemption_wallet,
+                    freeze_gate,
+                },
                 ctx,
                 pool,
             )
@@ -1811,9 +1831,17 @@ async fn run_provider_command<W: Write>(
             symbol,
             quantity,
             redemption_wallet,
+            freeze_gate,
         } => {
-            rebalancing::alpaca_redeem_command(stdout, symbol, quantity, redemption_wallet, ctx)
-                .await
+            rebalancing::alpaca_redeem_command(
+                stdout,
+                symbol,
+                quantity,
+                redemption_wallet,
+                freeze_gate,
+                ctx,
+            )
+            .await
         }
         ProviderCommand::DividendBump { symbol, quantity } => {
             dividend::dividend_bump_command(stdout, symbol, quantity, ctx, provider).await
--- a/src/cli/rebalancing.rs
+++ b/src/cli/rebalancing.rs
@@ -19,6 +19,7 @@ use st0x_execution::{
     FractionalShares, Symbol, TimeInForce,
 };
 use st0x_finance::Usdc;
+use st0x_issuance_client::IssuanceClient;
 use st0x_raindex::{RaindexService, RaindexVaultId};
 use st0x_tokenization::{
     AlpacaTokenizationService, IssuerRequestId, TokenizationRequest, TokenizationRequestStatus,
@@ -33,6 +34,7 @@ use crate::equity_redemption::{
 };
 use crate::rebalancing::equity::{CrossVenueEquityTransfer, EquityTransferServices};
 use crate::rebalancing::to_wrapped_equities;
+use crate::rebalancing::trigger::freeze::FreezeStatusReader;
 use crate::rebalancing::usdc::{CrossVenueCashTransfer, UsdcSettlementParams, UsdcTransferError};
 use crate::telemetry::TelemetrySender;
 use crate::telemetry::broker::InstrumentedAlpacaBroker;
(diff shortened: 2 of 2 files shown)


0xgleb:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1024?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1024** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1024?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1024?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1030** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1030?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1023](https://github.com/ST0x-Technology/st0x.liquidity/pull/1023) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1023?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1024).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/cli/rebalancing.rs
```diff
@@ -182,6 +259,8 @@ pub(super) async fn transfer_equity_command<Writer: Write>(
         }
 
         TransferDirection::ToAlpaca => {
```
JuaniRios, line 384: should fix: the `to-alpaca` arm has no test coverage. This PR names `transfer-equity --direction to-alpaca` as one of the two paths it exists to close, but every new test drives `alpaca_redeem_command` only. No test calls `transfer_equity_command` with `TransferDirection::ToAlpaca`; the two existing tests the diff touches still pass `ToRaindex` and only gained the new field. There is also no `classify_command` test asserting that `transfer-equity --direction to-alpaca --force` maps to `FreezeGateMode::Bypass`.

The gaps are wider than one path. The PR's Testing section claims coverage for unfrozen assets and for issuance status errors failing closed, but neither exists: the only status the mock returns is `frozen`. So the `Err` branch, which is the actual fail-closed behavior this PR is about, is untested on both commands. AGENTS.md makes coverage for new logic non-negotiable. Please add tests driving `transfer_equity_command` with `ToAlpaca` for frozen, status-error, and enabled, plus a `Bypass` case asserting no issuance call, and add an issuance 5xx or transport case for the shared helper.

src/cli/rebalancing.rs
```diff
@@ -182,6 +259,8 @@ pub(super) async fn transfer_equity_command<Writer: Write>(
...
+            ensure_not_frozen_for_redemption(stdout, symbol, freeze_gate, ctx).await?;
```
JuaniRios: should fix: the gate runs once here, and then `resume_equity_to_hedging` is called with a freshly generated aggregate id, so it always takes `start_redemption`. That performs `withdraw_from_raindex` (submit and wait for confirmations) and then `unwrap_and_send` (unwrap, confirmations, then `send_for_redemption`). That is a multi-minute window between the check and the transfer to the redemption wallet. The in-process path deliberately checks at the latest safe point, consulting `held_by_dividend_freeze` on every pass while the aggregate is pre-send, but the CLI-built transfer never gets a reader wired, so it cannot. These commands are incident-response tooling run around dividend windows, which is exactly when a freeze lands mid-flow. Please construct the `IssuanceClient` once in `build_equity_transfer_services` when the mode is `Enforce`, call `set_freeze_status_reader` on the transfer, and skip it for `Bypass` so `--force` keeps its no-API-call behavior.

src/cli/rebalancing.rs
```diff
@@ -131,16 +133,91 @@ async fn build_equity_transfer_services(
...
+        )?;
+        return Ok(());
+    }
+
+    let reader = IssuanceClient::new(
+        ctx.issuance.base_url.clone(),
+        ctx.issuance.api_key.header_value(),
+    )?;
+
+    match reader.is_frozen(symbol).await {
```
JuaniRios: should fix: this treats one read of `GET /tokenized-assets/<underlying>/status` as authoritative permission to send tokens on-chain. The issuance SPEC says that field is the projected view state and is only as fresh as the projection: the view updates asynchronously after a freeze commits, so there is a window in which a just-committed freeze still reports enabled. Issuance also has no reject-redemption point, because it reacts to on-chain transfers. So an operator running this inside that window gets a green light and strands the tokens. Please pin the freshness assumption: either cite the SPEC caveat in the doc comment and state why a single read is accepted (for example that the dividend scheduler confirms propagation before the record date), or re-read after a short delay before committing to the send.

src/cli/rebalancing.rs
```diff
@@ -131,16 +133,91 @@ async fn build_equity_transfer_services(
     Ok(EquityTransferCliServices { transfer, wallet })
```
JuaniRios, line 176: should fix: this behavior is not in SPEC.md. The spec still documents only the automated rebalancing freeze guard and its `rebalancing.freeze_check` toggle. AGENTS.md puts SPEC.md first in the planning hierarchy. As it stands, an operator or maintainer cannot tell from the spec whether the manual redemption commands honor `freeze_check`, ignore it, or treat `--force` as a separate override. Please add the manual redemption CLI behavior to the Dividend Freeze section: fail-closed by default, `--force` bypass, and how it relates to `rebalancing.freeze_check`.

src/cli/mod.rs
```diff
@@ -302,6 +303,11 @@ pub enum Commands {
         /// Alpaca redemption wallet (overrides [tokenization] config)
         #[arg(long = "redemption-wallet")]
         redemption_wallet: Option<Address>,
+        /// Deliberately bypass the dividend freeze gate on to-alpaca (logged).
```
JuaniRios, line 374: minor: `--force` is declared for both directions and always converted through `FreezeGateMode::from_force_flag`, but only the `ToAlpaca` arm consults it. `transfer-equity --direction to-raindex --force` is accepted, prints nothing, logs nothing, and does nothing. In incident-response tooling an accepted flag that does nothing invites the operator to believe a guard was overridden when none applied, and it leaves the `--force` surface ambiguous for whoever adds a gate to the mint direction. Please reject the combination at parse time, or bail in `classify_command` with a message that `--force` applies only to `--direction to-alpaca`.

src/cli/rebalancing.rs
```diff
@@ -3639,6 +3727,106 @@ mod tests {
...
+            &mut stdout,
+            Symbol::new("AAPL").unwrap(),
+            FractionalShares::new(float!(10)),
+            None,
+            FreezeGateMode::Bypass,
+            &ctx,
+        )
+        .await;
+
+        // Past the gate the fixture ctx lacks wallet config, so the command
```
JuaniRios: minor: this test's freeze assertion is a negative (`!message.contains("frozen for a dividend")`), which passes for any error, including one raised before the gate is reached. Its green depends on the shared fixture having `redemption_wallet: None` and `wallet: None` so the command fails a few lines past the gate. That fixture is used by many tests and filling those fields in is a natural change; this test would then run past `AlpacaTokenizationService::new`. Only `status_mock.assert_calls(0)` and the stdout check carry real signal today. Please assert the exact downstream error the fixture produces, so the test pins where execution actually stopped.

src/cli/rebalancing.rs
```diff
@@ -131,16 +133,91 @@ async fn build_equity_transfer_services(
...
     stdout: &mut Writer,
-    direction: TransferDirection,
     symbol: &Symbol,
-    quantity: FractionalShares,
-    issuer_request_id: Option<Uuid>,
-    redemption_wallet_flag: Option<Address>,
+    gate: FreezeGateMode,
+    ctx: &Ctx,
+) -> anyhow::Result<()> {
+    if gate == FreezeGateMode::Bypass {
```
JuaniRios: nit: this dispatches with `if gate == FreezeGateMode::Bypass` and lets everything else fall through to the enforcing path. AGENTS.md requires an exhaustive `match` in production code so the compiler flags a new variant, and an equality test has the same blind spot as `matches!`. A third mode added later would silently inherit the enforce path. Please use `match gate { FreezeGateMode::Bypass => { ... return Ok(()); } FreezeGateMode::Enforce => {} }`.

src/cli/rebalancing.rs
```diff
@@ -131,16 +133,91 @@ async fn build_equity_transfer_services(
...
+            "Operator --force bypassing the dividend freeze gate without consulting issuance"
+        );
+        writeln!(
+            stdout,
+            "⚠️  --force: bypassing dividend freeze for {symbol}"
+        )?;
+        return Ok(());
+    }
+
+    let reader = IssuanceClient::new(
```
JuaniRios, line 269: nit: `IssuanceClient::new(...)?` sends the raw `ClientError` to `main`, which prints it to stderr. src/rebalancing/trigger/freeze.rs documents that this must not happen: `ClientError` embeds the request URL and socket address in both `Debug` and `Display`, which is why `FreezeClientFailure` exists to reduce it at the boundary. The `is_frozen` call below respects that and formats the endpoint-free error, but this constructor does not, and `ClientError::NotABase` and `Build` carry the configured base URL. CLI output gets pasted into chat during incidents, so a misconfigured `[issuance] base_url` becomes a message disclosing the internal endpoint. Please map the constructor error through the same boundary and bail with the endpoint-free display.

src/cli/rebalancing.rs
```diff
@@ -3639,6 +3727,106 @@ mod tests {
...
+                extended_hours_counter_trading: OperationMode::Disabled,
+                operational_limit: None,
+            },
+        );
+        ctx.issuance =
+            st0x_config::test_issuance_status_ctx(Url::parse(&server.base_url()).unwrap());
+        (ctx, status_mock)
+    }
+
+    // The dividend freeze gate on the manual redemption-side command: a
```
JuaniRios: nit: this section-marker comment restates what the two test names and their assertion messages already say. AGENTS.md lists section markers and restating what the code does under DON'T comment. Please delete it.

JuaniRios (commented):
The gate is in the right places and --force is modeled cleanly. Coverage does not match the description, though: nothing tests the to-alpaca path this PR names as its target, and neither the unfrozen case nor the status-error case (the actual fail-closed branch) is covered on either command. The check also runs once at command start, minutes and two on-chain transactions before the send, because the CLI-built transfer never gets a freeze reader wired.

JuaniRios (changes requested):
Requesting changes. The gate is in the right places and `--force` is modeled cleanly, but:

- Coverage does not match the description. No test drives `transfer_equity_command` with `ToAlpaca`, which is the path this PR names as its target, and neither the unfrozen case nor the issuance-error case is covered on either command. The error branch is the actual fail-closed behavior this PR exists to add.
- The check runs once at command start, then `start_redemption` performs a vault withdraw and an unwrap with confirmations before `send_for_redemption`. That is a multi-minute window, and the CLI-built transfer never gets a freeze reader wired, so it cannot re-check the way the in-process path does. These commands are run around dividend windows, which is exactly when a freeze lands mid-flow.

Please also document the manual freeze-gate behavior in SPEC.md, and route the `IssuanceClient` construction error through the same scrubbing boundary so the issuance endpoint does not reach operator output.

PR #1028, opened 2026-07-14
fix: disarm session guard after coordinated shutdown

0xgleb:
## What

Related to [RAI-1042](https://linear.app/makeitrain/issue/RAI-1042/liquidity-issuance-v1). Prevent the session cancellation guard from requesting a second server-supervisor shutdown after the coordinated shutdown path has already completed.

## Why

The full package suite exposed an error-level channel-closed log during an otherwise successful bot shutdown. The guard exists to stop detached tasks when the session future is cancelled; on the normal path, the shutdown coordinator has already stopped those tasks.

## How

The guard owns an optional bundle of the conductor abort handle and server-supervisor handle. It stays armed while the session future can be cancelled, then is explicitly disarmed after await_shutdown returns and before its result is propagated.

## Testing

- Added a traced regression test that waits for the supervisor channel to close, drops the disarmed guard, and asserts that no duplicate-shutdown error is logged
- Passed nix run .#ci across backend and dashboard

## Screenshots

Not applicable.

## Anything else

This issue was discovered while validating the dividend NAV-bump notification and is isolated here for focused review.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -256,11 +256,10 @@ async fn run_bot_session_inner(
     // trading against the database and the bound server port would leak. This
     // guard aborts the conductor and shuts the supervisor down on drop, so a
     // cancelled session actually stops. The graceful path already stops both
-    // before this guard drops, making it a no-op there.
-    let _session_guard = SessionTaskGuard {
-        conductor: bot_task.abort_handle(),
-        server_supervisor: server_supervisor.clone(),
-    };
+    // before this guard drops, so the guard is disarmed after the coordinator
+    // returns.
+    let mut session_guard =
+        SessionTaskGuard::new(bot_task.abort_handle(), server_supervisor.clone());
 
     let mut sigterm = tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())
         .context("failed to register SIGTERM handler")?;
@@ -271,14 +270,16 @@ async fn run_bot_session_inner(
         }
     };
 
-    await_shutdown(
+    let shutdown_result = await_shutdown(
         server_supervisor,
         bot_task,
         shutdown_token,
         shutdown_signal,
         GRACEFUL_SHUTDOWN_TIMEOUT,
     )
-    .await?;
+    .await;
+    session_guard.disarm();
+    shutdown_result?;
 
     info!(target: "startup", "Shutdown complete");
     Ok(())
@@ -298,14 +299,35 @@ async fn run_bot_session_inner(
 /// synchronously, so callers must not assume the conductor has stopped the
 /// instant the session future returns.
 struct SessionTaskGuard {
+    tasks: Option<SessionTasks>,
+}
+
+struct SessionTasks {
     conductor: AbortHandle,
     server_supervisor: SupervisorHandle,
 }
 
+impl SessionTaskGuard {
+    fn new(conductor: AbortHandle, server_supervisor: SupervisorHandle) -> Self {
+        Self {
+            tasks: Some(SessionTasks {
+                conductor,
+                server_supervisor,
+            }),
+        }
+    }
+
+    fn disarm(&mut self) {
+        self.tasks = None;
+    }
+}
+
 impl Drop for SessionTaskGuard {
     fn drop(&mut self) {
-        self.conductor.abort();
-        shutdown_supervisor(&self.server_supervisor);
+        if let Some(tasks) = &self.tasks {
+            tasks.conductor.abort();
+            shutdown_supervisor(&tasks.server_supervisor);
+        }
     }
 }
 
@@ -657,6 +679,34 @@ mod tests {
             post_wait.is_ok(),
             "supervisor.wait() should return after await_shutdown shut it down"
         );
+    }
+
+    #[tracing_test::traced_test]
+    #[tokio::test]
+    async fn disarmed_session_guard_does_not_repeat_supervisor_shutdown() {
+        let shutdown_token = CancellationToken::new();
+        let bot_task = tokio::spawn(async { Ok(()) });
+        let supervisor = SupervisorBuilder::default().build().run();
+        let mut session_guard = SessionTaskGuard::new(bot_task.abort_handle(), supervisor.clone());
+        let supervisor_for_wait = supervisor.clone();
+
+        await_shutdown(
+            supervisor,
+            bot_task,
+            shutdown_token,
+            std::future::pending::<()>(),
+            Duration::from_secs(5),
+        )
+        .await
+        .unwrap();
+        tokio::time::timeout(Duration::from_secs(1), supervisor_for_wait.wait())
+            .await
+            .unwrap()
+            .unwrap();
+        session_guard.disarm();
+        drop(session_guard);
+
+        assert!(!logs_contain("Failed to shutdown server supervisor"));
     }
 
     #[tokio::test]


0xgleb:
* **#1028** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1028?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1028?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1028).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/lib.rs
```diff
@@ -659,6 +681,34 @@ mod tests {
         );
     }
 
+    #[tracing_test::traced_test]
```
JuaniRios: should fix: the defect this PR fixes lives at the call site in `run_bot_session_inner`, where `disarm()` sits between `await_shutdown(...).await` and `shutdown_result?`. This test builds a `SessionTaskGuard` by hand and calls `disarm()` itself, so it supplies the very call the production code is supposed to make. Move `disarm()` after `shutdown_result?`, so it is skipped on the error path, or delete it in a later refactor, and this test stays green. What it actually asserts is that `Drop` is a no-op when `tasks` is `None`, which is a property of the `Option`. I checked the other entry points and the e2e suite: nothing else asserts on the duplicate-shutdown log either, so the regression is uncovered. Please extract the three lines (`await_shutdown`, `disarm`, propagate) into one small function, call it from `run_bot_session_inner`, and have the test drive that function. Please also add a positive control that drops an armed guard after the same coordinated shutdown and asserts the error log IS present, so the negative assertion stays meaningful if `shutdown()` ever becomes infallible upstream.

src/lib.rs
```diff
@@ -271,14 +270,16 @@ async fn run_bot_session_inner(
...
         shutdown_signal,
         GRACEFUL_SHUTDOWN_TIMEOUT,
     )
-    .await?;
+    .await;
+    session_guard.disarm();
+    shutdown_result?;
 
     info!(target: "startup", "Shutdown complete");
     Ok(())
```
JuaniRios, line 413: minor: the type-level doc was not updated with the behavior. It still says the guard fires when the session is cancelled "before `await_shutdown` runs", and that on the normal path the second `shutdown` "is ignored". Both are now wrong. The guard is deliberately kept armed through `await_shutdown`, which is exactly why `disarm()` is called after the await, so it covers cancellation during the coordinated shutdown too. And the second `shutdown` is not ignored: `SupervisorHandle::shutdown` returns `SendError` on a closed channel and `shutdown_supervisor` logs it at error level, which is the defect this PR fixes. The inline comment at the construction site was updated, so the two comments now disagree. A reader could conclude the `disarm()` call is redundant and remove it. Please restate the contract: the guard stays armed while the session future can be cancelled, `await_shutdown` already stops both on every path it returns from, so callers must disarm right after it returns and before propagating its result.

JuaniRios (commented):
The fix is correct and minimal. The regression test does not protect it, though: it constructs the guard by hand and calls disarm() itself, so moving or deleting the production call leaves it green. The type-level doc also still describes the old behavior and now contradicts the inline comment at the construction site.

JuaniRios (changes requested):
Requesting changes. The fix itself is correct and minimal; the problem is that nothing protects it:

- The regression test builds a `SessionTaskGuard` by hand and calls `disarm()` itself, so it supplies the very call the production code is supposed to make. Move `disarm()` after `shutdown_result?` or delete it, and the test stays green. Nothing else in the suite or the e2e tests asserts on the duplicate-shutdown log.
- The type-level doc still describes the old behavior and now contradicts the inline comment at the construction site, which could lead someone to remove the `disarm()` call as redundant.

ueco-jb (approved):
Reviewed the shutdown path: keeping the guard armed through await_shutdown and disarming in finish_session_shutdown only after it returns is correct, since both supervisors are stopped on every return before the one ? early exit. The disarm and armed-drop cases are both pinned with a positive control. No findings.  
LGTM :+1:

ueco-jb:
@coderabbitai full review

src/lib.rs
```diff
@@ -349,15 +349,16 @@ async fn run_bot_session_inner(
...
+    // be interrupted (the spawned session task being aborted -- how the e2e
     // chaos tests simulate a process crash), the conductor `JoinHandle` and the
     // runtime supervisor handles would simply be dropped. Dropping a `JoinHandle`
     // detaches its task rather than aborting it, so the conductor would keep
     // trading against the database and the bound server port would leak. This
-    // guard aborts the conductor and shuts both supervisors down on drop, so a
-    // cancelled session actually stops. The graceful path already stops both
-    // before this guard drops, making it a no-op there.
+    // guard therefore stays armed through coordinated shutdown. Every returning
+    // path has stopped the conductor and both supervisors, so
```
agryaznov: [nit] "Every returning path has stopped the conductor and both supervisors" reads as a claim about the whole function, but the two `?` returns just below it, on the SIGTERM registration and on `report_when_started`, return with the conductor spawned and both ports bound, and the armed drop is the only thing that stops them. A reader who takes the sentence at face value can move the guard below `report_when_started` and silently leak the conductor and the bound ports on a startup failure. Scope it to the shutdown future and say the earlier returns leave the guard armed on purpose.

agryaznov (commented):
Cross-review findings -- 1 comment.

**The description no longer matches the diff.** At this PR's merge base (`87d0350`) `session_guard.disarm()` already sits between the shutdown match and `shutdown_result?`; that ordering, the `armed` flag and the original guard test landed on master on 2026-08-18 in #1035. What is left here is the extraction into `finish_session_shutdown`, the doc rewrite and the two tests, all worth having. But the title says `fix:`, the What claims it prevents a second server-supervisor shutdown, and the Why cites an error-level channel-closed log, while since #1035 that case has been `debug!("Supervisor was already stopped")` -- the exact string the new tests assert on, at debug. Retitle to `refactor:`/`test:` and rewrite What and Why, or the merge commit claims a fix master already has.

src/lib.rs
```diff
@@ -1059,16 +1080,57 @@ mod tests {
...
+        let conductor_task = tokio::spawn(std::future::pending::<()>());
+        let supervisor = pending_test_supervisor();
+        let guard = SessionTaskGuard {
+            conductor: conductor_task.abort_handle(),
+            server_supervisor: supervisor.clone(),
+            equity_price_supervisor: None,
+            armed: true,
+        };
+
         shutdown_supervisor(&supervisor);
```
JuaniRios: low: These tests do not pin that the guard stays armed while the coordinated-shutdown future is pending. Moving `disarm()` before the await leaves both tests green, but cancellation in that window would detach the conductor and let it keep trading. Drive `finish_session_shutdown` with a pending future, cancel that in-progress helper future, and assert that the armed guard still cancels the conductor and repeats supervisor shutdown.

JuaniRios (changes requested):
Review finding: 1 blocking comment. The production fix is already on master through #1035, so I recommend closing this PR as superseded. If the added test coverage is kept, address the inline gap and reframe the PR as a test/refactor follow-up.

src/lib.rs
findolor: Added a regression that drives finish_session_shutdown into a deliberately pending coordinated-shutdown window, cancels the in-progress helper task, and asserts that the still-armed guard aborts the conductor and repeats supervisor cleanup. The existing completed and armed-control tests remain green.

findolor:
Reframed this PR as a test/refactor follow-up because #1035 already contains the production ordering fix. The remaining value is the extracted shutdown handoff, corrected contract documentation, and direct cancellation-window coverage. Full local CI passes.

JuaniRios (approved):

rouzwelt (approved):

PR #1030, opened 2026-07-14
feat: notify completed dividend NAV bumps

0xgleb:
## What

Advances [RAI-1042](https://linear.app/makeitrain/issue/RAI-1042/liquidity-issuance-v1) by notifying operators when a dividend NAV bump has completed onchain.

## Why

The wrapper donation command confirmed the transaction only in local CLI output. Dividend operations need a positive lifecycle signal that reaches the configured operations channel without exposing the donated share quantity.

## How

- Build the configured notifier through one shared alerts boundary used by both the conductor and CLI
- Send the completion notification only after the wrapper donation is confirmed
- Include the equity symbol and transaction hash, but no share quantity
- Keep delivery best-effort so a Telegram failure cannot turn a confirmed onchain donation into a CLI failure
- Parse Telegram Bot API envelopes so HTTP 200 with ok false is a delivery failure
- Strip token-bearing URLs and provider response bodies from notification errors

## Testing

- Added success-path coverage for the exact notification and absence of the share quantity
- Added coverage proving notification failure preserves CLI success after confirmation
- Added Telegram coverage for redacted request URLs, redacted API bodies, and HTTP-success envelopes that report failure
- Passed nix run .#ci across backend and dashboard

## Screenshots

Not applicable.

## Anything else

The notification transport hardening applies to existing operational alerts as well as this dividend lifecycle signal.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -4157,6 +4157,13 @@ other, startup fails fast rather than running with a half-configured alert
 channel. A notification delivery failure is logged but never crashes the
 monitor.
 
+After a wrapper donation is confirmed, the bot sends a dividend NAV-bump
+completion notification containing the equity symbol and transaction hash, but
+no share quantity. Delivery is best-effort: a Telegram failure is logged without
+turning the confirmed on-chain operation into a CLI failure. Telegram errors
+never expose the bot-token URL or response body, and an HTTP-success response
+whose Bot API envelope reports `ok: false` is treated as failed delivery.
+
 ### BaseToAlpaca deposit send
 
 The CCTP mint on the BaseToAlpaca leg sets `mintRecipient` to the bot's own
--- a/src/alerts/mod.rs
+++ b/src/alerts/mod.rs
@@ -1,5 +1,5 @@
-//! Operational alerting: out-of-band notifications for conditions an operator
-//! must react to (currently, a low native-gas balance on the bot wallet).
+//! Operational alerting: out-of-band notifications for conditions and completed
+//! lifecycle operations an operator needs to see.
 //!
 //! The [`Notifier`] trait abstracts the delivery channel; [`TelegramNotifier`]
 //! is the only implementation today. Monitors that raise alerts (see
@@ -17,6 +17,10 @@ pub(crate) use telegram::TelegramNotifier;
 
 use async_trait::async_trait;
 use reqwest::StatusCode;
+use std::sync::Arc;
+use tracing::{debug, info};
+
+use st0x_config::AlertsCtx;
 
 /// Sends an operational alert over some channel.
 ///
@@ -27,14 +31,32 @@ pub(crate) trait Notifier: Send + Sync {
     async fn notify(&self, message: &str) -> Result<(), NotifierError>;
 }
 
+/// Builds the configured operational notification channel.
+///
+/// An absent `[alerts]` section is represented explicitly by [`NoopNotifier`].
+/// If the section is present, construction failures propagate so callers never
+/// silently discard notifications an operator configured.
+pub(crate) fn build_notifier(
+    alerts: Option<&AlertsCtx>,
+) -> Result<Arc<dyn Notifier>, NotifierError> {
+    let Some(alerts) = alerts else {
+        debug!("Operational alerting is not configured; using NoopNotifier");
+        return Ok(Arc::new(NoopNotifier));
+    };
+    let notifier =
+        TelegramNotifier::new(&alerts.bot_token, alerts.chat_id, alerts.message_thread_id)?;
+    info!("Telegram operational notifier configured");
+    Ok(Arc::new(notifier))
+}
+
 #[derive(Debug, thiserror::Error)]
 pub(crate) enum NotifierError {
     #[error("failed to build Telegram HTTP client")]
     ClientBuild(#[source] reqwest::Error),
     #[error("Telegram sendMessage request failed")]
     Request(#[source] reqwest::Error),
-    #[error("Telegram API returned error status {status}: {body}")]
-    ApiError { status: StatusCode, body: String },
+    #[error("Telegram API reported failed delivery with HTTP status {status}")]
+    ApiError { status: StatusCode },
 }
 
 /// A [`Notifier`] that discards every message without error.
--- a/src/alerts/telegram.rs
+++ b/src/alerts/telegram.rs
@@ -5,6 +5,7 @@
 //! monitor can be exercised against a capturing mock in tests.
 
 use async_trait::async_trait;
+use serde::Deserialize;
 use serde_json::json;
 
 use super::{Notifier, NotifierError};
@@ -87,16 +88,28 @@ impl Notifier for TelegramNotifier {
             .json(&body)
             .send()
             .await
-            .map_err(NotifierError::Request)?;
+            .map_err(|error| NotifierError::Request(error.without_url()))?;
 
         let status = response.status();
-        if status.is_success() {
-            return Ok(());
+        if !status.is_success() {
+            return Err(NotifierError::ApiError { status });
         }
+        let envelope = response
+            .json::<TelegramResponse>()
+            .await
+            .map_err(|error| NotifierError::Request(error.without_url()))?;
 
-        let body = response.text().await.map_err(NotifierError::Request)?;
-        Err(NotifierError::ApiError { status, body })
+        if envelope.ok {
+            Ok(())
+        } else {
+            Err(NotifierError::ApiError { status })
+        }
     }
+}
+
+#[derive(Deserialize)]
+struct TelegramResponse {
+    ok: bool,
 }
 
 #[cfg(test)]
@@ -104,7 +117,7 @@ mod tests {
     use httpmock::Method::POST;
     use httpmock::MockServer;
     use reqwest::StatusCode;
-    use serde_json::{Value, json};
+    use serde_json::json;
 
     use super::*;
 
@@ -183,7 +196,7 @@ mod tests {
                     .json_body(json!({
                         "ok": false,
                         "error_code": 400,
-                        "description": "Bad Request: chat not found",
+                        "description": "SENSITIVE_PROVIDER_BODY",
                     }));
             })
             .await;
@@ -192,15 +205,53 @@ mod tests {
             TelegramNotifier::with_base_url(&server.base_url(), "123:abc", 42, None).unwrap();
 
         let error = notifier.notify("hello").await.unwrap_err();
+        let rendered_error = format!("{error:?}");
 
-        let NotifierError::ApiError { status, body } = error else {
+        let NotifierError::ApiError { status, .. } = error else {
             panic!("expected ApiError, got: {error}");
         };
 
         assert_eq!(status, StatusCode::BAD_REQUEST);
-        let parsed: Value = serde_json::from_str(&body).unwrap();
-        assert_eq!(parsed["description"], json!("Bad Request: chat not found"));
+        assert!(!rendered_error.contains("SENSITIVE_PROVIDER_BODY"));
 
         mock.assert_async().await;
+    }
+
+    #[tokio::test]
+    async fn notify_redacts_token_bearing_url_from_request_errors() {
+        let notifier =
+            TelegramNotifier::with_base_url("http://127.0.0.1:1", "SENSITIVE_BOT_TOKEN", 42, None)
+                .unwrap();
+
+        let error = notifier.notify("hello").await.unwrap_err();
+
+        assert!(!format!("{error:?}").contains("SENSITIVE_BOT_TOKEN"));
+    }
+
+    #[tokio::test]
+    async fn notify_rejects_unsuccessful_bot_api_envelope() {
+        let server = MockServer::start_async().await;
+
+        server
+            .mock_async(|when, then| {
+                when.method(POST).path("/bot123:abc/sendMessage");
+                then.status(200)
+                    .header("content-type", "application/json")
+                    .json_body(json!({
+                        "ok": false,
+                        "error_code": 400,
+                        "description": "chat not found",
+                    }));
+            })
+            .await;
+        let notifier =
+            TelegramNotifier::with_base_url(&server.base_url(), "123:abc", 42, None).unwrap();
+
+        let error = notifier.notify("hello").await.unwrap_err();
+
+        let NotifierError::ApiError { status, .. } = error else {
+            panic!("expected ApiError, got: {error}");
+        };
+        assert_eq!(status, StatusCode::OK);
     }
 }
--- a/src/cli/wrapper.rs
+++ b/src/cli/wrapper.rs
@@ -7,6 +7,7 @@ use st0x_config::Ctx;
 use st0x_execution::{FractionalShares, Positive, Symbol};
 use st0x_wrapper::{Wrapper, WrapperService};
 
+use crate::alerts::{Notifier, build_notifier};
 use crate::rebalancing::to_wrapped_equities;
 
 pub(super) async fn wrap_equity_command<Writer: Write>(
@@ -144,13 +145,15 @@ pub(super) async fn donate_equity_command<Writer: Write>(
         base_wallet,
         to_wrapped_equities(&ctx.assets.equities.symbols),
     );
+    let notifier = build_notifier(ctx.alerts.as_ref())?;
 
-    donate_equity_with_wrapper(stdout, &wrapper, owner, symbol, quantity).await
+    donate_equity_with_wrapper(stdout, &wrapper, notifier.as_ref(), owner, symbol, quantity).await
 }
 
 async fn donate_equity_with_wrapper<Writer: Write, WrapperImpl: Wrapper + ?Sized>(
     stdout: &mut Writer,
     wrapper: &WrapperImpl,
+    notifier: &dyn Notifier,
     owner: Address,
     symbol: Symbol,
     quantity: Positive<FractionalShares>,
@@ -181,6 +184,17 @@ async fn donate_equity_with_wrapper<Writer: Write, WrapperImpl: Wrapper + ?Sized
 
     let donate_tx_hash = wrapper.donate(wrapped_token, underlying_amount).await?;
 
+    let notification =
+        format!("Dividend NAV bump completed: {symbol}; transaction {donate_tx_hash}");
+    if let Err(error) = notifier.notify(&notification).await {
+        tracing::error!(
+            target: "dividend",
+            ?error,
+            %symbol,
+            %donate_tx_hash,
+            "Dividend NAV bump notification delivery failed"
+        );
+    }
     writeln!(stdout, "   Transaction hash: {donate_tx_hash}")?;
     writeln!(stdout, "Donation completed successfully!")?;
 
@@ -190,6 +204,8 @@ async fn donate_equity_with_wrapper<Writer: Write, WrapperImpl: Wrapper + ?Sized
 #[cfg(test)]
 mod tests {
     use alloy::primitives::Address;
+    use async_trait::async_trait;
+    use reqwest::StatusCode;
     use url::Url;
 
     use st0x_config::ExecutionThreshold;
@@ -203,6 +219,7 @@ mod tests {
         donate_equity_command, donate_equity_with_wrapper, unwrap_equity_command,
         unwrap_equity_with_wrapper, wrap_equity_command, wrap_equity_with_wrapper,
     };
+    use crate::alerts::{CapturingNotifier, NoopNotifier, Notifier, NotifierError};
     use crate::test_utils::positive_shares;
 
     fn create_ctx_without_rebalancing() -> Ctx {
@@ -245,6 +262,41 @@ mod tests {
             rest_api: None,
             issuance: create_test_issuance_ctx(),
             redemption_wallet: None,
+        }
+    }
+
+    struct FailingNotifier;
+
+    #[async_trait]
+    impl Notifier for FailingNotifier {
+        async fn notify(&self, _message: &str) -> Result<(), NotifierError> {
+            Err(NotifierError::ApiError {
+                status: StatusCode::INTERNAL_SERVER_ERROR,
+            })
+        }
+    }
+
+    #[derive(Default)]
+    struct RejectTransactionHashWriter {
+        output: Vec<u8>,
+    }
+
+    impl std::io::Write for RejectTransactionHashWriter {
+        fn write(&mut self, buffer: &[u8]) -> std::io::Result<usize> {
+            let mut candidate = self.output.clone();
+            candidate.extend_from_slice(buffer);
+            if String::from_utf8_lossy(&candidate).contains("Transaction hash:") {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::BrokenPipe,
+                    "transaction hash output rejected",
+                ));
+            }
+            self.output.extend_from_slice(buffer);
+            Ok(buffer.len())
+        }
+
+        fn flush(&mut self) -> std::io::Result<()> {
+            Ok(())
         }
     }
 
@@ -459,11 +511,13 @@ mod tests {
         let wrapper = MockWrapper::new()
             .with_wrapped_token(wrapped_token)
             .with_tokenized_shares(underlying_token);
+        let notifier = CapturingNotifier::default();
         let mut stdout = Vec::new();
 
         donate_equity_with_wrapper(
             &mut stdout,
             &wrapper,
+            &notifier,
             Address::repeat_byte(0xaa),
             Symbol::new("AAPL").unwrap(),
             positive_shares("10.5"),
@@ -480,6 +534,62 @@ mod tests {
         assert!(output.contains("no shares minted"));
         assert!(output.contains("Transaction hash:"));
         assert!(output.contains("Donation completed successfully"));
+        let transaction_hash = output
+            .lines()
+            .find_map(|line| line.trim().strip_prefix("Transaction hash: "))
+            .unwrap();
+        let messages = notifier.messages();
+        assert_eq!(
+            messages,
+            vec![format!(
+                "Dividend NAV bump completed: AAPL; transaction {transaction_hash}"
+            )]
+        );
+        assert!(!messages[0].contains("10.5"));
+    }
+
+    #[tokio::test]
+    async fn donate_equity_succeeds_when_notification_delivery_fails() {
+        let wrapper = MockWrapper::new()
+            .with_wrapped_token(Address::repeat_byte(0x22))
+            .with_tokenized_shares(Address::repeat_byte(0x11));
+        let mut stdout = Vec::new();
+
+        donate_equity_with_wrapper(
+            &mut stdout,
+            &wrapper,
+            &FailingNotifier,
+            Address::repeat_byte(0xaa),
+            Symbol::new("AAPL").unwrap(),
+            positive_shares("10.5"),
+        )
+        .await
+        .unwrap();
+
+        let output = String::from_utf8(stdout).unwrap();
+        assert!(output.contains("Donation completed successfully"));
+    }
+
+    #[tokio::test]
+    async fn donate_equity_notifies_after_confirmation_even_if_stdout_closes() {
+        let wrapper = MockWrapper::new()
+            .with_wrapped_token(Address::repeat_byte(0x22))
+            .with_tokenized_shares(Address::repeat_byte(0x11));
+        let notifier = CapturingNotifier::default();
+        let mut stdout = RejectTransactionHashWriter::default();
+
+        donate_equity_with_wrapper(
+            &mut stdout,
+            &wrapper,
+            &notifier,
+            Address::repeat_byte(0xaa),
+            Symbol::new("AAPL").unwrap(),
+            positive_shares("10.5"),
+        )
+        .await
+        .unwrap_err();
+
+        assert_eq!(notifier.messages().len(), 1);
     }
 
     #[tokio::test]
@@ -490,6 +600,7 @@ mod tests {
         let error = donate_equity_with_wrapper(
             &mut stdout,
             &wrapper,
+            &NoopNotifier,
             Address::repeat_byte(0xaa),
             Symbol::new("AAPL").unwrap(),
             positive_shares("10.5"),
@@ -511,6 +622,7 @@ mod tests {
         let error = donate_equity_with_wrapper(
             &mut stdout,
             &wrapper,
+            &NoopNotifier,
             Address::repeat_byte(0xaa),
             Symbol::new("AAPL").unwrap(),
             positive_shares("10.5"),
--- a/src/conductor.rs
+++ b/src/conductor.rs
@@ -52,7 +52,7 @@ use st0x_tokenization::AlpacaTokenizationService;
 use st0x_tokenization::Tokenizer;
 use st0x_wrapper::WrapperService;
 
-use crate::alerts::{NoopNotifier, NotifierError, TelegramNotifier};
+use crate::alerts::build_notifier;
 use crate::conductor::exit::{ConductorExit, MonitorTaskError};
 use crate::conductor::monitor::order_fills::{CutoffProbe, probe_cutoff_block_support};
 use crate::dashboard::Broadcaster;
@@ -584,7 +584,7 @@ impl Conductor {
             Err(CtxError::NotRebalancing) => None,
             Err(error) => return Err(error.into()),
         };
-        let notifier = build_operational_notifier(ctx.alerts.as_ref())?;
+        let notifier = build_notifier(ctx.alerts.as_ref())?;
 
         let PositionAndRebalancing {
             position,
@@ -1271,29 +1271,6 @@ impl PositionAndRebalancing {
             })
         }
     }
-}
-
-/// Builds the shared operational alerting notifier.
-///
-/// Returns `Ok(Arc<NoopNotifier>)` when `[alerts]` is absent — silence is
-/// the correct behaviour for an unconfigured optional channel.
-///
-/// Returns `Err` when `[alerts]` IS present but `TelegramNotifier` fails to
-/// initialise. The caller must propagate this so the server fails to start:
-/// an operator who configured `[alerts]` believes alerts are active; silently
-/// falling back to Noop would suppress all redrive-limit and terminal-error
-/// pages with no runtime indication.
-fn build_operational_notifier(
-    alerts: Option<&st0x_config::AlertsCtx>,
-) -> Result<Arc<dyn crate::alerts::Notifier>, NotifierError> {
-    let Some(alerts) = alerts else {
-        debug!("Operational alerting: [alerts] section absent, using NoopNotifier");
-        return Ok(Arc::new(NoopNotifier));
-    };
-    let notifier =
-        TelegramNotifier::new(&alerts.bot_token, alerts.chat_id, alerts.message_thread_id)?;
-    info!("Operational alerting: Telegram notifier configured");
-    Ok(Arc::new(notifier))
 }
 
 /// Wires the dividend freeze guard onto the rebalancing service when enabled.
@@ -3702,7 +3679,7 @@ mod tests {
             inventory,
             Arc::new(MockWrapper::new()),
             RebalancingSchedulers::new(&apalis_pool),
-            Arc::new(NoopNotifier),
+            Arc::new(crate::alerts::NoopNotifier),
         ))
     }
 
@@ -8411,7 +8388,7 @@ mod tests {
                 vault_registry_projection,
                 schedulers: RebalancingSchedulers::new(&apalis_pool),
                 telemetry: TelemetrySender::disabled(),
-                notifier: Arc::new(NoopNotifier),
+                notifier: Arc::new(crate::alerts::NoopNotifier),
             },
         )
         .await
@@ -10683,18 +10660,18 @@ mod tests {
         );
     }
 
-    /// When `[alerts]` is absent, `build_operational_notifier` returns a `NoopNotifier`
+    /// When `[alerts]` is absent, `build_notifier` returns a `NoopNotifier`
     /// that silently discards notifications without error.
     #[tokio::test]
-    async fn build_operational_notifier_returns_ok_noop_when_alerts_absent() {
-        let notifier = build_operational_notifier(None).unwrap();
+    async fn build_notifier_returns_ok_noop_when_alerts_absent() {
+        let notifier = build_notifier(None).unwrap();
         notifier
             .notify("test message")
             .await
             .expect("NoopNotifier must not error on notify");
     }
 
-    /// When `[alerts]` IS present, `build_operational_notifier` constructs a real
+    /// When `[alerts]` IS present, `build_notifier` constructs a real
     /// Telegram notifier and returns `Ok` -- it must NOT fail startup for a
     /// well-formed config, and it must NOT silently fall back to `NoopNotifier`
     /// (which would suppress every redrive-limit and terminal-error page).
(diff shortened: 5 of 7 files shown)


0xgleb:
* **#1030** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1030?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1023](https://github.com/ST0x-Technology/st0x.liquidity/pull/1023) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1023?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1024](https://github.com/ST0x-Technology/st0x.liquidity/pull/1024) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1024?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1030?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1030).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/alerts/mod.rs
```diff
@@ -27,14 +31,32 @@ pub(crate) trait Notifier: Send + Sync {
...
+        TelegramNotifier::new(&alerts.bot_token, alerts.chat_id, alerts.message_thread_id)?;
+    info!("Telegram operational notifier configured");
+    Ok(Arc::new(notifier))
+}
+
 #[derive(Debug, thiserror::Error)]
 pub(crate) enum NotifierError {
     #[error("failed to build Telegram HTTP client")]
     ClientBuild(#[source] reqwest::Error),
     #[error("Telegram sendMessage request failed")]
```
JuaniRios, line 75: should fix: `ApiError` loses every Telegram diagnostic. The body is no longer read, and `TelegramResponse` deserializes only `ok`, so `error_code` and `description` are discarded. Those fields are the only way to tell "chat not found" from "bot was blocked" from "message thread not found". None of them carries the bot token; the token lives in the URL path, which `without_url()` already handles. The `ok: false` branch also reuses the HTTP status, so an envelope rejection renders as "Telegram API reported failed delivery with HTTP status 200 OK", which reads as a contradiction. Delivery is best-effort and failures are only logged, so a bad `chat_id` or a bot removed from the group leaves the alert channel silently dead, and the post-burn stall loop re-logs that same contentless line on every sweep. This affects the gas monitor and the rebalancing alerts too, not just the new dividend signal. Please split the variant: keep `ApiError { status }` for transport-level non-2xx, and add an envelope variant carrying a typed `error_code`, so no opaque `String` variant is introduced.

src/cli/wrapper.rs
```diff
@@ -144,13 +145,15 @@ pub(super) async fn donate_equity_command<Writer: Write>(
         base_wallet,
         to_wrapped_equities(&ctx.assets.equities.symbols),
     );
+    let notifier = build_notifier(ctx.alerts.as_ref())?;
```
JuaniRios: should fix: `build_notifier` returns a `NoopNotifier` with only a `debug!` line when `[alerts]` is absent. The help text for `donate-equity` and `dividend-bump` tells the operator to point `--config` and `--secrets` at a different, issuer-side config pair than the one the server runs with, and those hand-maintained files are unlikely to carry an `[alerts]` section. On that path the notification is discarded silently, while stdout still prints "Donation completed successfully!". The same line prints when `notify()` fails, because that error only reaches `tracing::error!`. So the operator cannot tell "channel notified" from "nothing sent". Please print the notification outcome on stdout as one of sent, skipped because `[alerts]` is not configured, or failed with the error, and raise the unconfigured branch in `build_notifier` from `debug!` to `warn!`.

Separately, this call is `build_notifier(...)?`, so a notifier construction failure aborts the command. For standalone `donate-equity` that is harmless, because nothing has happened yet. But `dividend_bump_command` calls this as step 3 of 3, after the market buy has filled and the tokenize has minted onchain. `reqwest::Client::builder().build()` is fallible in practice, most plausibly from a malformed `HTTPS_PROXY` or `ALL_PROXY` value that reqwest parses at build time. The operator would then be left with equity bought and tokenized but no NAV bump, caused purely by a notification setup problem. Please build the notifier once at the top of `dividend_bump_command`, before the buy, and pass it down.

src/alerts/telegram.rs
```diff
@@ -87,24 +88,36 @@ impl Notifier for TelegramNotifier {
...
+            .map_err(|error| NotifierError::Request(error.without_url()))?;
 
         let status = response.status();
-        if status.is_success() {
-            return Ok(());
+        if !status.is_success() {
+            return Err(NotifierError::ApiError { status });
         }
+        let envelope = response
+            .json::<TelegramResponse>()
```
JuaniRios, line 102: minor: this PR adds a new failure path (a 200 response whose body does not deserialize into `TelegramResponse`), and no test covers it. All three mock-server tests return well-formed JSON. Previously a 200 short-circuited to `Ok(())` with no body parsing at all, so this is new logic. Please add an httpmock case that returns status 200 with a non-JSON body and assert the resulting error variant.

src/alerts/telegram.rs
```diff
@@ -87,24 +88,36 @@ impl Notifier for TelegramNotifier {
...
+            .await
+            .map_err(|error| NotifierError::Request(error.without_url()))?;
```
JuaniRios: minor: a body that fails to deserialize is mapped to `NotifierError::Request`, the same variant as the `.send()` network failure above, whose message is "Telegram sendMessage request failed". That message is accurate only before a response exists. During an incident an on-call engineer reads it and reasonably concludes the POST never reached Telegram, when in fact the request succeeded and only the body shape was wrong. Please add a distinct variant for a response body that cannot be parsed, and keep `Request` for genuine pre-response transport failures.

src/alerts/telegram.rs
```diff
@@ -87,24 +88,36 @@ impl Notifier for TelegramNotifier {
...
+        if envelope.ok {
+            Ok(())
+        } else {
+            Err(NotifierError::ApiError { status })
+        }
     }
 }
 
+#[derive(Deserialize)]
+struct TelegramResponse {
```
JuaniRios, line 126: minor: `ok` is declared non-optional, so a 2xx body that is missing the field, types it differently, or is not JSON becomes a hard deserialization failure, and a message Telegram actually delivered is then reported as an error. Nothing in the diff pins that assumption. Please add a doc comment on this struct citing the Bot API "Making requests" section and the guarantee it relies on (the response is a JSON object that always has a Boolean field `ok`), or add a test that encodes a real captured response body.

src/alerts/telegram.rs
```diff
@@ -192,15 +205,53 @@ mod tests {
...
 
         assert_eq!(status, StatusCode::BAD_REQUEST);
-        let parsed: Value = serde_json::from_str(&body).unwrap();
-        assert_eq!(parsed["description"], json!("Bad Request: chat not found"));
+        assert!(!rendered_error.contains("SENSITIVE_PROVIDER_BODY"));
 
         mock.assert_async().await;
     }
+
+    #[tokio::test]
```
JuaniRios, line 236: minor: this test unwraps the error and only checks that the debug string does not contain the token. It never asserts the variant. AGENTS.md says to unwrap and assert the exact variant with `matches!`, and the neighboring tests in this file already destructure `NotifierError::ApiError { status }`. As written it would still pass if a refactor changed which variant a connection failure produces. Please add a `let NotifierError::Request(_) = error else { panic!(...) };` before the substring check.

SPEC.md
```diff
@@ -4157,6 +4157,13 @@ other, startup fails fast rather than running with a half-configured alert
 channel. A notification delivery failure is logged but never crashes the
 monitor.
 
+After a wrapper donation is confirmed, the bot sends a dividend NAV-bump
```
JuaniRios, line 5372: minor: this paragraph records omitting the share quantity as a confidentiality property, but the same message carries the donation transaction hash, and the donation is a plain public ERC-20 transfer into the wrapper. Anyone with the hash can read the exact amount from the chain in one call. So the property does not hold; only the literal string is absent from the chat message. Someone later tightening confidentiality around dividend sizing will read this and build on a guarantee that is one lookup away from being false. Please reword it to state the actual intent, or, if withholding the size is a real requirement, reconsider including the hash.

src/cli/wrapper.rs
```diff
@@ -181,6 +184,17 @@ async fn donate_equity_with_wrapper<Writer: Write, WrapperImpl: Wrapper + ?Sized
 
     let donate_tx_hash = wrapper.donate(wrapped_token, underlying_amount).await?;
 
+    let notification =
+        format!("Dividend NAV bump completed: {symbol}; transaction {donate_tx_hash}");
+    if let Err(error) = notifier.notify(&notification).await {
+        tracing::error!(
```
JuaniRios: nit: AGENTS.md says to qualify imports only to prevent ambiguity, and gives `info!` rather than `tracing::info!` as the example. This file has no other `tracing` usage and no `use tracing::...` import. Please add `use tracing::error;` to the external-imports group and call `error!(...)`. The same applies to `crate::alerts::NoopNotifier` in the conductor test changes.

JuaniRios (commented):
The feature works and the redaction hardening is welcome, but two things regress operability. ApiError now discards error_code and description, so an ok:false envelope renders as "HTTP status 200 OK" and a dead alert channel is undiagnosable, and that applies to the gas and rebalancing alerts too. On the issuer config path the notification can be a silent no-op while the CLI prints unqualified success.

JuaniRios (changes requested):
Requesting changes. The feature works and the redaction hardening is welcome, but two things regress operability:

- `ApiError` now discards `error_code` and `description`, and the `ok: false` branch reuses the HTTP status, so an envelope rejection renders as "HTTP status 200 OK". Neither field carries the bot token. A dead alert channel becomes undiagnosable, and this applies to the gas monitor and the rebalancing alerts, not just the new dividend signal.
- On the issuer config path the notification can be a silent no-op. `build_notifier` returns a `NoopNotifier` with a `debug!` line when `[alerts]` is absent, while stdout still prints unqualified success. The same line prints when delivery fails.

Please also build the notifier before the buy in `dividend_bump_command`, so a notifier construction failure cannot abort the flow after the tokenize has already minted.

src/alerts/mod.rs
0xgleb: Fixed. Non-2xx responses remain ApiError with the HTTP status. A 2xx envelope with ok=false now becomes EnvelopeRejected with a typed numeric error code, while a missing code becomes MalformedEnvelope. Telegram descriptions and response bodies remain excluded.

src/cli/wrapper.rs
0xgleb: Fixed. The notifier is constructed before any buy or tokenization step. The CLI reports sent, skipped, or failed on stdout, and an absent alerts section now emits a warning instead of a debug line.

src/alerts/telegram.rs
0xgleb: Added a 200 response with a non-JSON body and an exact ResponseDecode assertion.

src/alerts/telegram.rs
0xgleb: Fixed. Response decoding now has its own ResponseDecode variant; Request is reserved for failures before a response exists.

src/alerts/telegram.rs
0xgleb: Added the Bot API Making requests citation and documented the guaranteed Boolean ok field and error_code contract.

src/alerts/telegram.rs
0xgleb: Fixed. The redaction test first destructures NotifierError::Request, then verifies that the token-bearing URL is absent.

SPEC.md
0xgleb: Fixed. The SPEC now says only that the literal quantity is omitted for concision and explicitly notes that the public transaction reveals the amount.

src/cli/wrapper.rs
0xgleb: Fixed. The wrapper imports and calls error! directly, and the conductor test uses the already imported NoopNotifier rather than a qualified path.

crates/cli/src/cli/wrapper.rs
```diff
@@ -190,6 +204,21 @@ async fn donate_equity_with_wrapper<Writer: Write, WrapperImpl: Wrapper + ?Sized
 
     let donate_tx_hash = wrapper.donate(wrapped_token, underlying_amount).await?;
 
+    let notification =
+        format!("Dividend NAV bump completed: {symbol}; transaction {donate_tx_hash}");
+    match notifier.notify(&notification).await {
+        Ok(()) => writeln!(stdout, "   Notification: sent")?,
+        Err(notification_error) => {
+            error!(
+                target: "dividend",
```
agryaznov: [low] `dividend` is a new tracing target. docs/observability.md requires a new target to be added both to its "Existing targets" table and to `DOMAIN_TARGETS` in `crates/config/src/telemetry.rs` (whose `crate_filter_enables_every_domain_target_in_use` test says its list mirrors every `target: "..."` in the workspace); neither is updated here. Beyond the bookkeeping, what this line reports is the failure of an operational alert, so it wants `operational_alert` -- the target the pipeline actually pages on -- rather than a bespoke one nothing downstream matches.

SPEC.md
```diff
@@ -6079,6 +6079,14 @@ a leftover `[alerts]` table (`bot_token`) in the secrets file. Deployed config
 versions still carry the former, deployed secret versions the latter, and the
 previous build required them.)
 
+After a wrapper donation is confirmed, the bot emits a structured
+`operational_alert` ERROR log containing the equity symbol and transaction hash.
+The message omits the literal share quantity to stay concise; this is not a
+confidentiality guarantee because the public transaction reveals the amount. The
+production notifier cannot fail in-process: human delivery is owned by the Cloud
```
agryaznov: [high] This makes Cloud Logging and Grafana the owner of delivery, but only `st0x-cli` emits this alert and neither way operators run it reaches that stream. On the VMs the bump runs as `docker exec`, and `gcplogs` exports the container's main process only, so exec output returns to the ssh client. On the droplet the CLI starts no systemd unit. The ERROR event lands in the terminal of whoever ran the command, where the confirmation already was, so the gap stays open. Emit it from a process inside the pipeline, or scope this paragraph to the CLI running as the logged container process.

src/alerts/mod.rs
```diff
@@ -1,6 +1,9 @@
-//! Operational alerting: out-of-band notifications for conditions an operator
-//! must react to (low native-gas balance, stuck rebalancing transfers,
-//! dead-lettered hedges, supervised-worker terminal failures).
+//! Operational alerting.
+//!
+//! These out-of-band notifications cover conditions and completed lifecycle
```
agryaznov: [medium] docs/observability.md calls `operational_alert` the ERROR events the pipeline pages on, and every current sender is a fault. A routine completion now rides the same channel at the same level with nothing in the payload to tell them apart, so the rule that pages for a dead-lettered hedge pages for a donation that went fine. Needs a discriminating field or a second target the rule can split on.

crates/cli/src/cli/wrapper.rs
```diff
@@ -190,6 +204,21 @@ async fn donate_equity_with_wrapper<Writer: Write, WrapperImpl: Wrapper + ?Sized
 
     let donate_tx_hash = wrapper.donate(wrapped_token, underlying_amount).await?;
 
+    let notification =
+        format!("Dividend NAV bump completed: {symbol}; transaction {donate_tx_hash}");
+    match notifier.notify(&notification).await {
+        Ok(()) => writeln!(stdout, "   Notification: sent")?,
```
agryaznov: [medium] The donation is irreversible by this point, yet this line and the `Notification: failed` line below propagate write errors with `?`. Pipe into `head` and a broken pipe turns a landed bump into a non-zero exit with no success line; the natural response is to re-run and bump twice. The new closed-stdout test asserts exactly that shape. Make the output after the receipt best-effort, like the notification already is.

crates/cli/src/cli/wrapper.rs
agryaznov: [low] `dividend` is a new target, and docs/observability.md requires adding one to its table and to `DOMAIN_TARGETS`, whose test asserts that list mirrors the workspace; neither is updated. This line reports a failed operational alert, so it wants `operational_alert` rather than a target nothing matches.

crates/cli/src/cli/wrapper.rs
```diff
@@ -190,6 +204,21 @@ async fn donate_equity_with_wrapper<Writer: Write, WrapperImpl: Wrapper + ?Sized
 
     let donate_tx_hash = wrapper.donate(wrapped_token, underlying_amount).await?;
 
+    let notification =
+        format!("Dividend NAV bump completed: {symbol}; transaction {donate_tx_hash}");
```
agryaznov: [low] The message names symbol and hash but not the chain, while `--network` selects one and the NAV is per wrapper. A hash does not identify a chain, so once a symbol is wrapped on two chains the operator cannot tell which ratio moved, which is what the alert exists to say. `network` is already in scope at the call site.

crates/cli/src/cli/wrapper.rs
```diff
@@ -813,7 +909,66 @@ mod tests {
...
+            &wrapper,
+            &notifier,
+            Address::repeat_byte(0xaa),
+            Symbol::new("AAPL").unwrap(),
+            positive_shares("10.5"),
+        )
+        .await
+        .unwrap_err();
+
+        assert_eq!(notifier.messages().len(), 1);
```
agryaznov: [low] The `unwrap_err()` discards the error, so this passes on any failure rather than the broken pipe it provokes; AGENTS.md asks for the exact variant. And keying the rejecting writer on `"Transaction hash:"` pins only that the notification precedes that line, so reordering output fails the test though its named behaviour is unchanged. Key on the final success line and assert `BrokenPipe`.

crates/cli/src/cli/wrapper.rs
```diff
@@ -274,6 +307,25 @@ mod tests {
         wrap_equity_with_wrapper,
     };
 
+    #[derive(Default)]
+    struct CapturingNotifier {
```
agryaznov: [nit] `src/alerts/mod.rs` already has a `CapturingNotifier` of this shape, and this PR already moved `NotifierError::Simulated` onto `test-support`. Gate `test_support` the same way and export it; the cli dev-deps already enable that feature.

agryaznov (changes requested):
Cross-review findings -- 7 comments.

**tl;dr** The CLI fires one alert after the donation confirms. Its relationship to the bump holds: only on a confirmed bump, once, no blocking or rollback on delivery failure, no masked amounts. The delivery story does not: the only emitter is the operator CLI, run so its output never reaches the stream the alert rule reads, on the channel the pipeline pages on, with nothing separating a success from a fault.

Rebase note: master renamed `TradingChainContext` to `HedgedChainContext` in the file this PR threads the notifier through.

ueco-jb (commented):
I didn't find any new findings over those that are already up needs to be addressed before any approval can land.

crates/cli/src/cli/wrapper.rs
```diff
@@ -144,6 +149,51 @@ pub(super) async fn donate_equity_command<Writer: Write>(
...
+impl Notifier for BotProcessNotifier {
+    async fn notify(&self, message: &str) -> Result<(), NotifierError> {
+        self.client
+            .post(&self.url)
+            .json(&DividendNavBumpNotice {
+                message: message.to_owned(),
+            })
+            .send()
+            .await?
+            .error_for_status()?;
```
agryaznov: [medium] Nothing exercises this notifier against a server. The path, the POST verb, the `message` field name and the non-2xx mapping live half here and half in the route registration, and no test crosses the two: the api.rs test builds `DividendNavBumpNotice` in-process, and every CLI test passes a double. A typo in either half compiles, passes CI, and first shows up as `Notification: failed` after an irreversible donation. `rebalancing.rs` already drives the sibling loopback endpoints against `httpmock` with `ctx.server_port = server.port()`; one such test here would pin the contract.

crates/cli/src/cli/wrapper.rs
```diff
@@ -190,8 +243,28 @@ async fn donate_equity_with_wrapper<Writer: Write, WrapperImpl: Wrapper + ?Sized
...
+    let notification =
+        format!("Dividend NAV bump completed: {symbol} on {chain}; transaction {donate_tx_hash}");
+    let _ = writeln!(stdout, "   Transaction hash: {donate_tx_hash}");
+    match notifier.notify(&notification).await {
+        Ok(()) => {
+            let _ = writeln!(stdout, "   Notification: sent");
+        }
+        Err(notification_error) => {
+            error!(
+                target: "operational_alert",
```
agryaznov: [medium] This fault alert is emitted by the CLI process, whose output is precisely what the new endpoint exists to bypass -- `gcplogs` exports the container's main process, not the `docker exec` session. So the one event that records a lost completion notice is itself lost, while `alert = true, alert_kind = "fault"` asserts it reaches the rule that pages. Log it on a CLI-local target without the alert markers, and have the `Notification: failed` line state that the bump did not reach the pipeline and needs recording by hand.

agryaznov (commented):
Re-review at e4f8292 -- 2 comments.

The delivery model now holds: the CLI no longer tries to emit the alert itself, it POSTs the confirmed bump to the running bot's loopback-only `/alerts/dividend-nav-bump`, and the bot emits it into the stream the pipeline actually reads. `alert_kind` splits completions from pages and the docs say so, the post-receipt writes are best-effort so a broken pipe can no longer turn a landed donation into a non-zero exit, the message names the chain, the closed-stdout test asserts `BrokenPipe`, and `CapturingNotifier` is shared through `test-support` instead of copied. All seven earlier findings are addressed. What is left is the new hop itself: the wire contract between the CLI and the route has no test, and the branch that runs when that hop fails alerts into the same unread stream the endpoint was built to escape.

agryaznov (commented):
a couple more issues found

crates/cli/src/cli/wrapper.rs
findolor: Fixed. Delivery failure now logs on the normal CLI module target without `operational_alert`, `alert`, or `alert_kind` markers. Stdout explicitly says the completion did not reach the alert pipeline and must be recorded manually; the donation still exits successfully.

crates/cli/src/cli/wrapper.rs
findolor: Fixed. Added real `httpmock` coverage around `dividend_completion_notifier` that pins the POST path, exact `message` JSON field, successful 204 handling, and non-2xx mapping to `NotifierError::Http`.

src/api.rs
```diff
@@ -1378,6 +1378,26 @@ pub struct ResumeResponse {
...
+/// NAV bump to the running bot's exported log pipeline.
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DividendNavBumpNotice {
+    pub message: String,
+}
+
+async fn report_dividend_nav_bump(Json(notice): Json<DividendNavBumpNotice>) -> StatusCode {
+    error!(
+        target: "operational_alert",
```
JuaniRios: [high] This completion still feeds the production fault pager. Current `T0Trade/t0.devops` matches every `operational_alert` line and does not inspect `alert_kind`, and I found no companion routing PR, so a successful NAV bump becomes a false on-call page. Land/deploy a backward-compatible completion exclusion and separate completion route first (or use a distinct non-paging target until it exists), then make the docs match the real rollout.

crates/cli/src/cli/dividend.rs
```diff
@@ -144,7 +157,14 @@ async fn dividend_bump_with_operations<Writer: Write + Send, Operations: Dividen
...
+            stdout,
+            symbol,
+            filled_quantity,
+            network,
+            notifier.as_ref(),
+            ctx,
+        )
         .await?;
 
     writeln!(stdout, "✅ Dividend NAV bump completed")?;
```
JuaniRios: [medium] The inner donation helper makes post-confirmation output best-effort, but this final `writeln!(...)?` can still return a non-zero exit after the buy, tokenization, donation, and notification have completed. That can prompt an operator or automation to repeat an irreversible bump. Make this write best-effort too and add a whole-command test with a persistently closed writer.

src/alerts/mod.rs
```diff
@@ -31,10 +35,14 @@ pub(crate) trait Notifier: Send + Sync {
 /// `Result` stays in the trait so alert-failure handling at the call sites
 /// (retry/backoff paths, bounded-timeout sends) remains exercisable in tests.
 #[derive(Debug, thiserror::Error)]
-pub(crate) enum NotifierError {
+#[non_exhaustive]
+pub enum NotifierError {
+    /// Failed to deliver an alert to the running bot process.
```
JuaniRios: [low] The surrounding docs say this error is uninhabited outside tests and notification is infallible in production, but `Http` is unconditional and the production CLI notifier returns it on connection, timeout, and HTTP-status failures. Update the module/type docs to describe both the infallible log notifier and the fallible CLI-to-bot notifier.

JuaniRios (changes requested):
Review findings: 3 changes required before merge.

crates/cli/src/cli/wrapper.rs
```diff
@@ -162,7 +167,51 @@ pub(super) async fn donate_equity_command<Writer: Write>(
...
+        .await
+}
+
+pub(super) fn dividend_completion_notifier(ctx: &Ctx) -> anyhow::Result<Arc<dyn Notifier>> {
+    let client = reqwest::Client::builder()
+        .timeout(Duration::from_secs(5))
+        .build()?;
+    let url = format!(
+        "http://127.0.0.1:{}/alerts/dividend-nav-bump",
+        ctx.server_port
```
agryaznov: [medium] On the invocation both commands document, this posts to the wrong port and the completion never reaches the pipeline.

`ctx.server_port` is whatever config the operator passed, and the help on `donate-equity` and `dividend-bump` says to point `--config`/`--secrets` at the issuer wallet. In-repo that is `config/s01-issuer.toml` with `server_port = 8011`; the bot runs on `8001` in both `config/prod` and `config/staging`. So every bump prints `Notification: failed` after an irreversible donation, and the feature is inert in production.

Resolve the port from the bot's own config rather than the issuer's, and probe the endpoint inside `dividend_completion_notifier` so the pre-buy ordering catches this before the money moves rather than after.

src/api.rs
```diff
@@ -2780,9 +2799,10 @@ async fn require_loopback(
 
 pub(crate) fn routes(ops_api: Option<&OpsApiConfig>) -> Router<AppState> {
     // The operator socket: mutation endpoints for the in-container CLI only.
-    // The IAP-gated `/liquidity-write` mount is the network route to the same
-    // handlers.
+    // Transfer handlers also have an IAP-gated `/liquidity-write` mount for
+    // remote operators; completion notices deliberately remain loopback-only.
     let loopback_only = Router::new()
+        .route("/alerts/dividend-nav-bump", post(report_dividend_nav_bump))
```
agryaznov: [low] Nothing stops a later change from mounting this under `/liquidity-write`, which would let any IAP-authenticated remote caller write straight into the pager stream.

The comment above states loopback-only as a deliberate property, but the only test drives the bare path with a non-loopback peer. Every sibling route here is deliberately dual-mounted, so adding this one to the write list is a plausible one-line mistake with no test to catch it.

Add a case asserting `POST /liquidity-write/alerts/dividend-nav-bump` is 404, next to the existing loopback test.

src/lib.rs
```diff
@@ -37,7 +37,7 @@ use crate::trading::onchain::trade_accountant::DexTradeAccountingJobQueue;
 /// and apalis drain.
 const GRACEFUL_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(90);
 
-mod alerts;
+pub mod alerts;
```
agryaznov: [nit] This exports more of the alerting surface than the CLI needs.

The CLI implements `Notifier` and matches on `NotifierError`, and its tests use `CapturingNotifier`. `LogNotifier` is bot-internal: nothing outside this crate constructs it, and it is the one type whose `alert_kind = "fault"` behaviour should not be callable from another binary.

Keep the module `pub` but leave `LogNotifier` as `pub(crate)`, per the visibility rule in AGENTS.md.

agryaznov (commented):
left inline comments

crates/cli/src/cli/wrapper.rs
findolor: Fixed. Dividend commands now require the bot config (or `ST0X_BOT_CONFIG`), load only its `server_port`, and issue a HEAD probe to the loopback completion endpoint while preparing the notifier before any buy or donation. The deployed `s01` wrapper explicitly exports `/run/st0x/st0x-hedge.config`, overridable with `S01_BOT_CONFIG`. Tests cover port loading, probe success, and unavailable endpoints.

src/api.rs
findolor: Fixed. Successful dividend completions now emit on the distinct `operational_notice` target with `notice = true` and `notice_kind = "dividend_nav_bump"`; they no longer contain `operational_alert`, so the existing substring-based fault pager cannot match them. SPEC.md now documents the deployed split.

src/api.rs
findolor: Fixed. Added an exact regression test with the IAP write router enabled asserting `POST /liquidity-write/alerts/dividend-nav-bump` returns 404.

src/lib.rs
findolor: Fixed. `LogNotifier` is now `pub(crate)` while the externally needed `Notifier`, `NotifierError`, and test-support notifier remain available.

crates/cli/src/cli/dividend.rs
findolor: Fixed. The final whole-command success write is best-effort and logs a local error instead of changing the exit status. Added a whole-command test whose writer closes permanently on the final completion write and verifies buy, tokenization, and donation all complete exactly once.

src/alerts/mod.rs
findolor: Fixed. The module and `NotifierError` docs now distinguish the infallible bot-internal `LogNotifier` from the fallible CLI HTTP notifier, including connection, timeout, and HTTP-status failures.

src/api.rs
```diff
@@ -2780,9 +2798,13 @@ async fn require_loopback(
...
 pub(crate) fn routes(ops_api: Option<&OpsApiConfig>) -> Router<AppState> {
     // The operator socket: mutation endpoints for the in-container CLI only.
-    // The IAP-gated `/liquidity-write` mount is the network route to the same
-    // handlers.
+    // Transfer handlers also have an IAP-gated `/liquidity-write` mount for
+    // remote operators; completion notices deliberately remain loopback-only.
     let loopback_only = Router::new()
+        .route(
+            "/alerts/dividend-nav-bump",
+            post(report_dividend_nav_bump).head(|| async { StatusCode::NO_CONTENT }),
```
agryaznov: [low] Nothing on this side pins that HEAD answers, so losing this handler breaks every dividend bump with no CI signal.

The probe is now a hard gate: if HEAD stops returning 2xx the command aborts before the buy. The CLI test mocks HEAD against `httpmock`, so it passes whatever this route does, and the two tests here only drive POST and the `/liquidity-write` 404. A refactor that drops the `.head(...)` arm, or a middleware that rejects the method, compiles and ships green.

Add a case asserting `HEAD /alerts/dividend-nav-bump` from a loopback peer returns 204, beside the POST test.

crates/cli/src/cli/wrapper.rs
```diff
@@ -159,10 +164,56 @@ pub(super) async fn donate_equity_command<Writer: Write>(
...
+
+pub(super) async fn dividend_completion_notifier(
+    bot_config: &Path,
+) -> anyhow::Result<Arc<dyn Notifier>> {
+    let server_port = load_operator_api_port(bot_config)?;
+    let client = reqwest::Client::builder()
+        .timeout(Duration::from_secs(5))
+        .build()?;
+    let url = format!("http://127.0.0.1:{server_port}/alerts/dividend-nav-bump");
+    client.head(&url).send().await?.error_for_status()?;
```
agryaznov: [low] A bot that is down or restarting now blocks the NAV bump entirely, and that is when an operator is most likely to be running one by hand.

The bump is an issuer operation, funded and signed by the issuer wallet: buy, tokenize, donate. None of it needs the bot. This probe makes bot liveness a precondition for starting it, with no way past: the command exits before the buy and there is no flag to proceed without the notice.

Failing before money moves is the right default, so keep it, but add an opt-out (`--no-notify`, skipping the probe and printing the same record-it-manually line the delivery-failure path prints) so an outage cannot strand a dividend.

agryaznov (commented):
two more low-severity issues found, otherwise LGTM

src/api.rs
```diff
@@ -1378,6 +1378,25 @@ pub struct ResumeResponse {
...
+/// Wire contract used by the in-container CLI to hand a confirmed dividend
+/// NAV bump to the running bot's exported log pipeline.
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DividendNavBumpNotice {
+    pub message: String,
+}
+
+async fn report_dividend_nav_bump(Json(notice): Json<DividendNavBumpNotice>) -> StatusCode {
+    error!(
```
JuaniRios: should fix: A successful dividend completion is still emitted at ERROR severity. Moving it off operational_alert avoids the current substring pager, but generic error-rate monitoring will still classify success as failure. Register operational_notice in DOMAIN_TARGETS, emit this at INFO, and test the intended filter and level contract.

docs/observability.md
```diff
@@ -40,15 +40,15 @@ info!(%symbol, %shares, "Hedging trade");
...
-| `operational_alert` | Operator alerts (ERROR events the log pipeline pages on) |
-| `orderbook`         | Onchain orderbook interactions                           |
-| `rebalancing`       | Portfolio rebalancing                                    |
-| `startup`           | Application initialization                               |
-| `tokenization`      | Tokenized equity minting                                 |
-| `wallet`            | Alpaca wallet / onchain wallet                           |
+| Target              | Subsystem                                                    |
+| ------------------- | ------------------------------------------------------------ |
+| `hedge`             | Hedging / position management                                |
+| `operational_alert` | Operator faults and lifecycle notices routed by `alert_kind` |
```
JuaniRios: should fix: This still documents lifecycle completions on operational_alert with alert_kind=completion, but the implementation emits operational_notice with notice_kind=dividend_nav_bump and never emits that completion alert_kind. Document the two distinct targets and fields, and add operational_notice to both this table and DOMAIN_TARGETS so operators configure the real signal.

src/api.rs
```diff
@@ -1378,6 +1378,25 @@ pub struct ResumeResponse {
     pub redemptions_failed: usize,
 }
 
+/// Wire contract used by the in-container CLI to hand a confirmed dividend
+/// NAV bump to the running bot's exported log pipeline.
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DividendNavBumpNotice {
```
JuaniRios: should fix: This audit boundary accepts only an opaque free-form message, so the bot cannot validate symbol, chain, or transaction hash before acknowledging delivery. Send typed fields and construct the canonical log message in the bot; reject blank or malformed notices in endpoint tests.

JuaniRios (changes requested):
The three earlier blockers are fixed, but the replacement completion channel is not internally consistent yet. Routine success is still emitted at ERROR severity, and the operator documentation still describes the retired target and field contract.

src/api.rs
JuaniRios: Good catch. I removed the HEAD probe and the HEAD route, so nothing gates the bump on HEAD anymore. See the other thread for why.

The route contract is now pinned on both sides with the same JSON literal:
- bot side: `dividend_nav_bump_notice_is_logged_as_info_operational_notice` POSTs `{"symbol":"AAPL","chain":"base","txHash":"0x…12ab"}` from a loopback peer to the real router and expects 204 plus the INFO notice
- CLI side: `bot_notice_client_posts_the_wire_contract` asserts the client sends exactly that body (`json_body`) to `DIVIDEND_NAV_BUMP_NOTICE_PATH`, and the path constant is shared by the route and the client

So if either side changes shape, one of the two tests fails.

Fixed in f7b3980a.

crates/cli/src/cli/wrapper.rs
JuaniRios: Agree, bot liveness should not gate an issuer operation. I went a bit further than `--no-notify`: there's no pre-flight contact with the bot at all now.

1. Before the buy, the CLI only reads `server_port` from the bot config and builds the HTTP client. Those are local setup steps, so a broken config still fails before money moves.
2. The CLI talks to the bot only after the donation receipt.
3. If that POST fails (bot down, restarting, non-204), the command still exits 0 and prints `WARNING: the donation succeeded, but the bot did not record the completion (...). Do not repeat the donation. Record it manually: <symbol> on <chain>, transaction <hash>`.

This goes to the operator's stdout, not only to a CLI log line that nobody reads. So a bot outage can't strand a dividend, and there's no flag to forget. Tests: `bot_notice_client_setup_does_not_contact_the_bot` and `donate_equity_succeeds_and_warns_when_notice_delivery_fails`.

Fixed in f7b3980a.

src/api.rs
JuaniRios: Fixed in f7b3980a. The notice is now `info!` on target `operational_notice`, and `operational_notice` is in `DOMAIN_TARGETS`.

Tests:
- `crate_filter_exports_info_operational_notices` (config crate): at the production `info` filter, the INFO notice gets exported and a DEBUG event on the same target gets dropped
- `crate_filter_enables_every_domain_target_in_use` now lists `operational_alert` and `operational_notice`
- the endpoint test asserts the line is ` INFO ` on ` operational_notice: ` and that nothing lands on `operational_alert`

docs/observability.md
JuaniRios: Fixed in f7b3980a. `alert_kind` is gone completely. With a separate target it was redundant, and `hedge.rs` never set it anyway. `src/alerts/mod.rs` is back to master.

docs/observability.md now has one row per target:
- `operational_alert`: ERROR, `alert = true`, pages
- `operational_notice`: INFO, `notice = true`, `notice_kind` (for example `dividend_nav_bump`), never pages

I also added a note that a `RUST_LOG` override with a bare level above `info` needs an explicit `operational_notice=info` segment. SPEC's "Structured log channel" section matches this now.

src/api.rs
JuaniRios: Fixed in f7b3980a. `DividendNavBumpNotice` is now `{ symbol: Symbol, chain: Chain, tx_hash: TxHash }` with `deny_unknown_fields`. The bot builds the canonical message and also logs the three values as structured fields (`symbol`, `chain`, `tx_hash`).

`dividend_nav_bump_notice_rejects_malformed_bodies` covers these cases, and none of them emits a notice:
- blank symbol: 422
- unknown chain: 422
- short tx hash: 422
- missing tx hash: 422
- extra field: 422
- the old free-form `{"message": ...}`: 422
- non-JSON: 400

On the CLI side, the client now requires exactly 204 and doesn't follow redirects or system proxies. So "recorded" means the handler actually ran.

JuaniRios:
@coderabbitai full review

JuaniRios:
@coderabbitai full review

JuaniRios:
@coderabbitai full review

JuaniRios:
@coderabbitai full review

JuaniRios:
@coderabbitai full review

JuaniRios:
@coderabbitai full review

JuaniRios:
@coderabbitai full review

JuaniRios:
@coderabbitai full review

JuaniRios:
@coderabbitai full review

rouzwelt (approved):

JuaniRios (approved):
LGTM, my review threads are addressed.

JuaniRios:
Closing this. The bump already works and the CLI prints the tx hash, so this adds a new bot endpoint and a required CLI flag just to get one log line into Grafana. If we want these in Grafana later, the bot can detect the donation onchain from the vaults it already watches, with no CLI or endpoint changes. Thanks for the work and the review rounds on this.

PR #1029, opened 2026-07-14
fix: eliminate Nix CI evaluation warnings

0xgleb:
## What

Eliminate repository-owned Nix evaluation warnings from the CI command by using the current nixfmt package name and pinning the fixed-output hash for the event-sorcery 0.1.1 Cargo dependency.

## Why

The full CI gate warned that nixfmt-rfc-style is deprecated and that the Cargo.lock event-sorcery source had no output hash. The existing outputHashes entry still referenced an obsolete rev rather than the current tag and commit.

## How

- Select pkgs.nixfmt as the flake formatter
- Replace the stale event-sorcery rev key with the exact tag and locked commit
- Pin the hash reported by Nix for that fixed-output derivation

## Testing

- Rebuilt nix run .#st0x-dto -- dashboard/src/lib/api without either repository evaluation warning
- Passed nix run .#ci across backend and dashboard

## Screenshots

Not applicable.

## Anything else

This warning cleanup is kept separate from the dividend behavior change for focused review.

PR #1034, opened 2026-07-15
fix: validate Turnkey configuration at boundary

0xgleb:
## What

Validate Turnkey organization IDs and API authentication keys at the configuration boundary instead of deferring malformed values to the signing SDK.

## Why

Whitespace-only organization IDs and malformed P-256 private keys could previously deserialize into apparently valid configuration and fail only when constructing or using the Turnkey client. That delayed startup failure, weakened error clarity, and allowed invalid states to exist in memory.

## How

- Parse `TurnkeyOrganizationId` through a fallible smart constructor.
- Trim surrounding organization-ID whitespace and reject an empty result.
- Parse `TurnkeyApiPrivateKey` through `TurnkeyP256ApiKey` and reject invalid P-256 key material.
- Keep API private-key `Debug` output redacted.
- Surface failures as configuration-parse errors before wallet construction.
- Gate signer-specific test imports so feature-minimal test builds remain valid.

## Testing

Added coverage for:

- trimming valid organization IDs;
- rejecting whitespace-only organization IDs;
- rejecting malformed P-256 API keys; and
- rejecting both invalid values through the full `WalletKind::Turnkey` configuration path.

## Screenshots

Not applicable.

## Anything else

The deprecated infallible constructors remain temporarily for compatibility, while all configuration deserialization uses the validated path.

### DIFF
--- a/crates/evm/src/lib.rs
+++ b/crates/evm/src/lib.rs
@@ -1143,12 +1143,16 @@ where
 
 #[cfg(test)]
 mod tests {
+    #[cfg(any(feature = "turnkey", feature = "local-signer"))]
     use alloy::consensus::{Receipt, ReceiptEnvelope, ReceiptWithBloom};
     use alloy::network::EthereumWallet;
     use alloy::node_bindings::{Anvil, AnvilInstance};
-    use alloy::primitives::{Address, Bloom, U256};
+    #[cfg(any(feature = "turnkey", feature = "local-signer"))]
+    use alloy::primitives::Bloom;
+    use alloy::primitives::{Address, U256};
     use alloy::providers::ProviderBuilder;
     use alloy::providers::mock::Asserter;
+    #[cfg(any(feature = "turnkey", feature = "local-signer"))]
     use alloy::rpc::types::TransactionReceipt;
     use alloy::signers::local::PrivateKeySigner;
     use alloy::sol;
@@ -1539,6 +1543,69 @@ mod tests {
         assert!(
             matches!(error, EvmError::WalletConfigParse(_)),
             "expected WalletConfigParse error from credentials parse, got: {error:?}"
+        );
+    }
+
+    #[cfg(feature = "turnkey")]
+    #[tokio::test]
+    async fn wallet_kind_turnkey_rejects_blank_organization_id_during_parse() {
+        let anvil = Anvil::new().spawn();
+        let provider = ProviderBuilder::new()
+            .disable_recommended_fillers()
+            .connect_http(anvil.endpoint_url());
+
+        let ctx = WalletCtx {
+            settings: MaybeParser(Some(serde_json::json!({
+                "address": format!("{}", Address::random()),
+                "organization_id": "   "
+            }))),
+            credentials: MaybeParser(Some(serde_json::json!({
+                "api_private_key":
+                    "0000000000000000000000000000000000000000000000000000000000000001"
+            }))),
+            provider,
+            required_confirmations: 1,
+        };
+
+        let result = WalletKind::Turnkey.try_into_wallet(ctx).await;
+        let Err(error) = result else {
+            panic!("expected blank organization ID to fail during parsing");
+        };
+
+        assert!(
+            matches!(error, EvmError::WalletConfigParse(_)),
+            "expected WalletConfigParse error, got: {error:?}"
+        );
+    }
+
+    #[cfg(feature = "turnkey")]
+    #[tokio::test]
+    async fn wallet_kind_turnkey_rejects_invalid_api_key_during_parse() {
+        let anvil = Anvil::new().spawn();
+        let provider = ProviderBuilder::new()
+            .disable_recommended_fillers()
+            .connect_http(anvil.endpoint_url());
+
+        let ctx = WalletCtx {
+            settings: MaybeParser(Some(serde_json::json!({
+                "address": format!("{}", Address::random()),
+                "organization_id": "org-test-123"
+            }))),
+            credentials: MaybeParser(Some(serde_json::json!({
+                "api_private_key": "not-a-p256-private-key"
+            }))),
+            provider,
+            required_confirmations: 1,
+        };
+
+        let result = WalletKind::Turnkey.try_into_wallet(ctx).await;
+        let Err(error) = result else {
+            panic!("expected malformed API private key to fail during parsing");
+        };
+
+        assert!(
+            matches!(error, EvmError::WalletConfigParse(_)),
+            "expected WalletConfigParse error, got: {error:?}"
         );
     }
 
--- a/crates/evm/src/turnkey.rs
+++ b/crates/evm/src/turnkey.rs
@@ -4,8 +4,11 @@
 //! enclaves for low-latency signing (50-100ms). Like
 //! [`RawPrivateKeyWallet`](super::local::RawPrivateKeyWallet), it wraps
 //! the base provider with a [`WalletFiller`] -- the only difference is
-//! the signer: `TurnkeySigner` (remote signing via Turnkey API) instead
+//! the signer: `TracingTurnkeySigner` (remote signing via Turnkey API) instead
 //! of `PrivateKeySigner` (local key).
+
+use std::sync::Arc;
+use std::time::{Duration, SystemTime, SystemTimeError, UNIX_EPOCH};
 
 use alloy::consensus::{SignableTransaction, TxEnvelope};
 use alloy::eips::eip2718::{Decodable2718, Eip2718Error};
@@ -24,8 +27,6 @@ use futures::lock::Mutex;
 use reqwest::header::CONTENT_TYPE;
 use serde::Deserialize;
 use serde::Serialize;
-use std::sync::Arc;
-use std::time::{Duration, SystemTime, SystemTimeError, UNIX_EPOCH};
 use tracing::{info, trace};
 use turnkey_api_key_stamper::{Stamp, StampHeader, TurnkeyP256ApiKey};
 use turnkey_client::generated::{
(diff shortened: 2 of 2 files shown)


0xgleb:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1034?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1034** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1034?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1034?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1034).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/evm/src/turnkey.rs
```diff
@@ -42,33 +43,89 @@ use crate::{Evm, EvmError, TryIntoWallet, Wallet, WalletCtx};
...
+        value.try_into()
+    }
+
+    #[deprecated(note = "use TurnkeyApiPrivateKey::try_new")]
     pub fn new(value: String) -> Self {
         Self(value)
     }
 }
 
+impl TryFrom<String> for TurnkeyApiPrivateKey {
```
JuaniRios, line 106: critical: this validator panics instead of returning an error for the most likely bad input. `TurnkeyP256ApiKey::from_strings` hex-decodes the string and then calls `SecretKey::from_bytes(bytes.as_ref().into())`. That `.into()` is generic-array's `From<&[T]> for &GenericArray<T, N>`, which asserts the slice length is exactly 32 and panics otherwise. Only non-hex input takes the `Err` path. Reproduced against the locked versions: `from_strings("not-a-p256-private-key")` returns `Err`, but `from_strings("deadbeef")`, `"00"`, and `""` all panic with `assertion left == right failed`. The rustdoc on `try_new` promises an `InvalidApiPrivateKey` error. Because this now runs inside a serde `Deserialize` impl, a truncated or padded secret aborts the process at boot with a generic-array assertion that never mentions Turnkey, the wallet, or the config field. `validate-config` also cannot catch it, because the deploy gate keeps the wallet section as an untyped `toml::Value` and never deserializes these credentials, so the deploy passes and the host crash-loops. Please hex-decode yourself first, reject a length other than 32 with a dedicated error variant that carries the observed length and never the key material, and only then hand the bytes to the SDK. Apply the same guard at the `TracingTurnkeyClient::from_api_key` call site.

crates/evm/src/turnkey.rs
```diff
@@ -730,6 +787,32 @@ mod tests {
...
+    }
+
+    #[test]
+    fn organization_id_rejects_only_whitespace() {
+        let error = TurnkeyOrganizationId::try_new("   ".to_string()).unwrap_err();
+
+        assert!(matches!(error, TurnkeyValueError::EmptyOrganizationId));
+    }
+
+    #[test]
```
JuaniRios, line 1291: should fix: both new negative tests feed `"not-a-p256-private-key"`, which is the only input class that returns an error. Every valid-hex wrong-length input panics instead, so the suite reports the validator as working while the dominant real failure mode aborts the process. The `wallet_kind_turnkey_*` tests also assert only `matches!(error, EvmError::WalletConfigParse(_))`, and an unrelated missing field produces the same variant, so they do not pin down why parsing failed. Please add cases for `"deadbeef"`, an empty string, and a 31-byte hex string, asserting the specific error variant. In the lib.rs tests, assert on the boxed error's message so the assertion distinguishes key validation from a generic serde error.

crates/evm/src/turnkey.rs
```diff
@@ -42,33 +43,89 @@ use crate::{Evm, EvmError, TryIntoWallet, Wallet, WalletCtx};
...
+
+    #[deprecated(note = "use TurnkeyApiPrivateKey::try_new")]
     pub fn new(value: String) -> Self {
         Self(value)
     }
 }
 
+impl TryFrom<String> for TurnkeyApiPrivateKey {
+    type Error = TurnkeyValueError;
+
```
JuaniRios, line 108: should fix: `.map_err(|_| TurnkeyValueError::InvalidApiPrivateKey)?` throws away the underlying `StamperError`, which carries the specific reason (bad hex, wrong byte length, invalid scalar). `InvalidApiPrivateKey` is a unit variant with no `#[source]`, so every failure mode collapses into the same static message. AGENTS.md calls this out under Error Type Design: prefer `#[from]` and `?`, preserve chains with `#[source]`, and use bare `.map_err` only when `From` cannot adapt the source. Operators lose the concrete cause of a startup failure, which works against this PR's own goal of clearer config-boundary errors. Please make the variant carry the source, for example `InvalidApiPrivateKey(#[source] turnkey_api_key_stamper::StamperError)`, and map with `.map_err(TurnkeyValueError::InvalidApiPrivateKey)?`.

crates/evm/src/turnkey.rs
```diff
@@ -42,33 +43,89 @@ use crate::{Evm, EvmError, TryIntoWallet, Wallet, WalletCtx};
...
+    ///
+    /// # Errors
+    ///
+    /// Returns [`TurnkeyValueError::EmptyOrganizationId`] when `value`
+    /// contains only whitespace.
+    pub fn try_new(value: String) -> Result<Self, TurnkeyValueError> {
+        value.try_into()
+    }
+
+    #[deprecated(note = "use TurnkeyOrganizationId::try_new")]
```
JuaniRios: should fix: `TurnkeyOrganizationId::new` and `TurnkeyApiPrivateKey::new` stay `pub` with zero callers. The diff migrated every call site to `try_new` and `test_organization_id()`, and `st0x-evm` is workspace-internal, so there is nothing to stay compatible with. The PR's stated motivation is that malformed values allowed invalid states to exist in memory; these two functions still allow exactly that, and `TurnkeyApiPrivateKey::new` is the path that keeps the panicking parse reachable. A deprecation warning is easy to allow away. Please delete both functions in this PR, so these types are constructible only through the validated path.

crates/evm/src/turnkey.rs
```diff
@@ -42,33 +43,89 @@ use crate::{Evm, EvmError, TryIntoWallet, Wallet, WalletCtx};
...
     pub fn new(value: String) -> Self {
         Self(value)
     }
 }
 
+impl TryFrom<String> for TurnkeyOrganizationId {
+    type Error = TurnkeyValueError;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
```
JuaniRios, line 75: should fix: `try_from` trims and rejects empty strings, but it still accepts non-UUID values such as `"org-test-123"`. Turnkey organization IDs are UUID-shaped in the production configs, and the generated Turnkey schema types them as UUIDs. So a typoed org ID still deserializes, passes startup, and fails only on the first signing request. That is the delayed failure this PR is meant to remove. Please parse the trimmed value as a UUID, store the canonical form, and add a `WalletKind::Turnkey` test that rejects a non-empty non-UUID id. Update the invalid-key test to use a valid UUID so it keeps testing the key path.

crates/evm/src/lib.rs
```diff
@@ -1542,6 +1546,69 @@ mod tests {
         );
     }
 
+    #[cfg(feature = "turnkey")]
+    #[tokio::test]
```
JuaniRios, line 1703: minor: both new tests call `Anvil::new().spawn()` and build a live HTTP provider, but `try_into_wallet` fails during settings and credentials parsing, before the provider is touched. The Anvil process is pure overhead on two tests whose subject is JSON deserialization. This follows the pre-existing sibling tests, so it is not introduced here, but the two new cases are a good place to stop paying for it. Please factor a helper that builds `WalletCtx` for parse-only failures without spawning Anvil.

crates/evm/src/turnkey.rs
```diff
@@ -4,9 +4,12 @@
 //! enclaves for low-latency signing (50-100ms). Like
 //! [`RawPrivateKeyWallet`](super::local::RawPrivateKeyWallet), it wraps
 //! the base provider with a [`WalletFiller`] -- the only difference is
-//! the signer: `TurnkeySigner` (remote signing via Turnkey API) instead
+//! the signer: `TracingTurnkeySigner` (remote signing via Turnkey API) instead
 //! of `PrivateKeySigner` (local key).
 
+use std::sync::Arc;
```
JuaniRios: nit: this patch puts a blank line between the std imports and the alloy imports. AGENTS.md puts std and external crates in one group with no empty lines inside it. rustfmt does not catch this. Please remove the blank line.

JuaniRios (commented):
The direction is right, but the validator does the opposite of what it promises for the most likely bad input. A valid-hex key of the wrong length panics inside generic-array rather than returning an error, and because this runs in a serde Deserialize impl it aborts the process at boot. validate-config cannot catch it, since the deploy gate keeps the wallet section untyped, so a bad secret passes the gate and crash-loops the host. The new tests only feed non-hex input, which is the one class that does return an error.

JuaniRios (changes requested):
Requesting changes. The validator does the opposite of what it promises for the most likely bad input:

- A valid-hex key of the wrong length panics inside generic-array's length assertion instead of returning `InvalidApiPrivateKey`. Only non-hex input takes the error path. Reproduced against the locked versions. Because this now runs inside a serde `Deserialize` impl, a truncated or padded secret aborts the process at boot with an assertion that never mentions Turnkey or the config field.
- `validate-config` cannot catch it either, because the deploy gate keeps the wallet section as an untyped `toml::Value`. The deploy passes and the host crash-loops.
- The new tests only feed non-hex input, which is the single class that does return an error, so the suite reports the validator as working.

Please also delete the deprecated infallible constructors (zero callers, workspace-internal crate) and preserve the parser error as a `#[source]`.

rouzwelt (approved):

crates/evm/src/turnkey.rs
0xgleb: Fixed in the submitted head. The key is hex-decoded, its observed length is checked by converting it to [u8; 32] before the SDK sees it, and the stamper boundary repeats the check. Valid-hex, wrong-length inputs now return a typed error without exposing key material or reaching the panicking conversion.

crates/evm/src/turnkey.rs
0xgleb: Added empty, deadbeef, 31-byte, and 33-byte cases with exact InvalidApiPrivateKeyLength assertions. Wallet parse-path tests also assert the specific key-validation message, so unrelated serde failures cannot satisfy them.

crates/evm/src/turnkey.rs
0xgleb: Fixed. Hex decoding, wrong length, and invalid P-256 scalar are distinct typed variants, while the hex, UUID, and Stamper errors are retained as sources.

crates/evm/src/turnkey.rs
0xgleb: Fixed. The unchecked public constructors were removed; organization IDs and API private keys now enter through fallible, validated constructors.

crates/evm/src/turnkey.rs
0xgleb: Fixed. Organization IDs are parsed as UUIDs, stored in canonical form, and covered by non-UUID and canonicalization tests. Invalid-key tests now use a valid organization UUID.

crates/evm/src/lib.rs
0xgleb: Fixed. Parse-only wallet tests now use a lightweight provider helper and no longer spawn Anvil.

crates/evm/src/turnkey.rs
0xgleb: Fixed. The std and external imports form one uninterrupted group, followed by the internal-import group.

PR #1040, opened 2026-07-16
fix: represent unknown block cutoff as null (RAI-1141)

0xgleb:
## What

[RAI-1141](https://linear.app/makeitrain/issue/RAI-1141/make-block-lag-telemetry-record-an-unknown-cutoff-as-null-not-block-0)

- Persist an unavailable ingestion cutoff as `NULL` instead of block zero.
- Keep lag absent when either the cutoff or checkpoint is unknown.
- Surface the latest unknown-cutoff sample as degraded in the dashboard.
- Remove a mock-server leak and a redemption E2E assertion race exposed by full-suite verification.

## Why

Synthesizing block zero made a paused order-fill monitor report a healthy zero lag. Operators need an unavailable cutoff to remain distinguishable from a caught-up monitor.

## How

`BlockLagSample` now models the cutoff as optional, and the read model uses the latest sample even when its lag is absent. A SQLite migration recreates the telemetry table with a nullable cutoff while preserving existing rows and indexes. The dashboard distinguishes an unknown cutoff from the separate no-checkpoint state.

## Testing

- `cargo nextest run -p st0x-hedge`: 2,556 passed, 4 skipped
- Redemption E2E stress run: 5/5 passed with retries disabled
- `cargo clippy -p st0x-hedge --all-targets --all-features`
- Dashboard tests: 244 passed
- Dashboard typecheck and lint
- Pre-commit hooks

## Screenshots

Not applicable.

## Anything else

The unknown-cutoff sample remains timestamped so the dashboard can show a current degraded state, while it is excluded from numeric lag charts.

PR #1037, opened 2026-07-15
fix: upgrade Turnkey SDK to 0.7.0

0xgleb:
## What

Upgrade the Turnkey API key stamper and client dependencies from 0.4.0 to 0.7.0.

## Why

The signing integration needs the current Turnkey request schema and SDK fixes. Keeping the upgrade isolated from the configuration-validation change makes the external-contract delta explicit and independently reviewable.

## How

- Pin `turnkey_api_key_stamper` and `turnkey_client` to 0.7.0.
- Refresh the lockfile.
- Populate the new optional `generate_app_proofs` request field with `None`.
- Update the signed-request fixture to assert the corresponding `generateAppProofs: null` wire payload.

## Testing

- Existing Turnkey signing and request-stamping tests cover the upgraded client.
- The request-body test verifies the exact 0.7.0 JSON shape.
- GitHub Actions static checks and backend build/test matrix pass.

## Screenshots

Not applicable.

## Anything else

No signing semantics change: transaction bytes, organization ID, signer address, and local signature verification remain unchanged. This PR only adapts the integration to the 0.7.0 SDK contract.

PR #1020, opened 2026-07-14
fix: gate CCTP burn Dropped verdicts on node freshness

0xgleb:
## What

Fixes RAI-1241. Gates every CCTP burn `Dropped` verdict (and the wallet's `wait_for_receipt` drop verdict) on node freshness: an absence reported by a possibly-lagging load-balanced backend never concludes `Dropped`.

## Why

The drop-detection concluded `Dropped` from a single node's `eth_getTransactionByHash == None`, but on dRPC a lagging backend returns `None` for a tx it hasn't indexed yet — including one already mined. In the 2026-07-07 `5c6f3fdd` incident the burn had mined the whole time; the false verdict paged the operator and stalled the transfer ~26 minutes. A synced node never returns `None` for a mined tx, so absence from a possibly-lagging node is not evidence of a drop.

## How

- `CctpEndpoint::burn_status` now takes `submitted_after_block` (the pre-burn chain head the transfer already records for the log scan) and only counts an absence toward the drop threshold when the answering node's head is `SCAN_FINALITY_MARGIN` blocks past it — the same discipline `find_recent_burn` already applies. A lagging node resolves as `Pending` (retryable) after the grace window, never `Dropped`.
- `wait_for_receipt` has no submission block available through the wallet trait, so it anchors on the head observed at the first post-grace absence and requires `dropped_head_advance` (2 blocks) of advance before trusting the miss run — a frozen-head backend can never conclude `Dropped` and falls out via the retryable inclusion timeout.
- The `Bridge::burn_status` trait and `check_pending_burn` thread the recorded `from_block` through.

## Testing

- New: `burn_status_reports_pending_when_node_lags_submission_block` (the incident shape: both lookups `None`, head behind the pre-burn block → `Pending`).
- New: `wait_for_receipt_does_not_drop_when_node_head_is_frozen` and `wait_for_receipt_drops_when_head_advances_past_absent_tx`.
- Existing drop-path fixtures updated to advance the anvil head past the margin (a frozen fixture head now correctly reads as a lagging node).

## Anything else

Stacked under the RAI-1242 cross-check (`DepositForBurn` scan before paging), which closes the second half of the same incident.

PR #1046, opened 2026-07-16
fix: recover exhausted worker circuits (RAI-1051)

0xgleb:
## What

[RAI-1051](https://linear.app/makeitrain/issue/RAI-1051)

Recover supervised durable-job workers automatically after terminal retry exhaustion, while isolating the failed worker from unrelated trading and rebalancing jobs.

## Why

A single terminal worker failure previously stopped the entire bot and required a process restart. One exhausted job should open a bounded circuit for its own worker, alert operators once, and recover without interrupting independent financial workflows.

## How

- Track terminal failures with a per-worker recovery circuit.
- Open the circuit after retry exhaustion and schedule a cooldown.
- Stop and restart only the affected worker.
- Reset the circuit after successful work.
- Align persisted job attempts with retry-middleware accounting.
- Emit one operator alert when the circuit opens and log its scheduled recovery.
- Leave explicitly best-effort workers outside supervision.

## Testing

- `nix run .#ci`
- 3,562 tests passed; one restart E2E used its configured retry.
- Clippy and formatting passed.
- Generated DTO and Bun inputs passed.
- Dashboard lint and Svelte checks passed with zero warnings or errors.

## Screenshots

Not applicable.

## Anything else

The production recovery cooldown is five minutes. Tests use a short deterministic cooldown.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -310,6 +310,17 @@ configured `apalis_finished_job_cleanup_interval_secs` cadence so the durable
 queue cannot grow without bound during normal operation or repeated restarts.
 The cadence must be explicitly configured and non-zero.
 
+Supervised apalis workers isolate terminal failures with a local recovering
+circuit. The middleware and durable queue share a four-attempt limit, so a job
+produces one terminal transition after exhausting its bounded retries and
+remains visible as a terminal Jobs row. The first consecutive terminal failure
+then pauses only that worker, sends an operator alert, and records a structured
+circuit transition with the worker name, error, failure count, and cooldown. The
+process and sibling workers remain live. After a five-minute cooldown, the same
+worker resumes without a process restart and records a structured recovery
+transition. A successful job resets the consecutive-failure count. Best-effort
+workers retain and log terminal jobs without pausing.
+
 Deployed systemd services must use bounded restart loops with a restart delay
 long enough to avoid runaway RPC consumption. After the restart limit is hit,
 the service stays stopped until an operator investigates.
@@ -2436,10 +2447,11 @@ misses) so a still-pending tx is never misclassified as dropped.
 **Bound**: Both revert and timeout redrives count against a shared
 `max_burn_revert_redrives` counter persisted in the job payload (durable across
 restarts). Once the bound is reached, the job propagates a
-`BurnRevertLimitReached` error so the circuit opens and the operator is alerted.
-Genuinely ambiguous failures (where the burn may have landed and `is_revert()`
-returns false, e.g. `MessageSentEventNotFound`) are never reclassified and
-always halt for operator review.
+`BurnRevertLimitReached` error so its worker circuit opens and the operator is
+alerted. Genuinely ambiguous failures (where the burn may have landed and
+`is_revert()` returns false, e.g. `MessageSentEventNotFound`) are never
+reclassified and always pause their worker for operator review before automatic
+recovery.
 
 **Alerting**: A Telegram alert fires at the warn threshold (`max / 2 + 1`
 redrives), at the limit, and before any terminal non-redriven error propagates.
@@ -4080,6 +4092,12 @@ requests to the backend.
 The bot raises out-of-band alerts for conditions an operator must react to
 quickly. Alerting is **optional**: when its configuration is absent the bot runs
 normally with no alert channel and the alert monitors are not spawned.
+
+Worker-circuit signals answer two bounded on-call questions. “Which worker
+paused and why?” is answered by the circuit-open transition and its operator
+alert. “Did it recover without a process restart?” is answered by the matching
+structured recovery transition. Worker names and transition names are bounded
+fields; errors remain event details rather than metric labels.
 
 ### Gas balance monitoring
 
--- a/src/conductor.rs
+++ b/src/conductor.rs
@@ -584,6 +584,7 @@ impl Conductor {
             Err(CtxError::NotRebalancing) => None,
             Err(error) => return Err(error.into()),
         };
+        let notifier = build_operational_notifier(ctx.alerts.as_ref())?;
 
         let PositionAndRebalancing {
             position,
@@ -614,6 +615,7 @@ impl Conductor {
                 vault_registry_projection,
                 schedulers: schedulers.clone(),
                 telemetry: telemetry.clone(),
+                notifier: notifier.clone(),
             },
         )
         .await?;
@@ -699,6 +701,7 @@ impl Conductor {
             wallet_polling,
             tokenizer,
             shutdown_token: shutdown_token.clone(),
+            notifier,
             #[cfg(any(test, feature = "test-support"))]
             failure_injector,
         };
@@ -1149,6 +1152,7 @@ struct RebalancingDeps {
     vault_registry_projection: Arc<Projection<VaultRegistry>>,
     schedulers: RebalancingSchedulers,
     telemetry: TelemetrySender,
+    notifier: Arc<dyn crate::alerts::Notifier>,
 }
 
 /// Position + rebalancing-adjacent infrastructure produced during conductor
@@ -1269,7 +1273,7 @@ impl PositionAndRebalancing {
     }
 }
 
-/// Builds the USDC alerting notifier.
+/// Builds the shared operational alerting notifier.
 ///
 /// Returns `Ok(Arc<NoopNotifier>)` when `[alerts]` is absent — silence is
 /// the correct behaviour for an unconfigured optional channel.
@@ -1279,16 +1283,16 @@ impl PositionAndRebalancing {
 /// an operator who configured `[alerts]` believes alerts are active; silently
 /// falling back to Noop would suppress all redrive-limit and terminal-error
 /// pages with no runtime indication.
-fn build_usdc_notifier(
+fn build_operational_notifier(
     alerts: Option<&st0x_config::AlertsCtx>,
 ) -> Result<Arc<dyn crate::alerts::Notifier>, NotifierError> {
     let Some(alerts) = alerts else {
-        debug!("USDC alerting: [alerts] section absent, using NoopNotifier");
+        debug!("Operational alerting: [alerts] section absent, using NoopNotifier");
         return Ok(Arc::new(NoopNotifier));
     };
     let notifier =
         TelegramNotifier::new(&alerts.bot_token, alerts.chat_id, alerts.message_thread_id)?;
-    info!("USDC alerting: Telegram notifier configured");
+    info!("Operational alerting: Telegram notifier configured");
     Ok(Arc::new(notifier))
 }
 
@@ -1471,7 +1475,7 @@ fn spawn_rebalancing_infrastructure<Chain: Wallet + Clone>(
         let transfer_usdc_to_market_making_queue =
             deps.schedulers.transfer_usdc_to_market_making.clone();
 
-        let usdc_notifier = build_usdc_notifier(deps.ctx.alerts.as_ref())?;
+        let usdc_notifier = deps.notifier.clone();
 
         let rebalancing_service = Arc::new(RebalancingService::new(
             RebalancingServiceConfig {
@@ -8406,6 +8410,7 @@ mod tests {
                 vault_registry_projection,
                 schedulers: RebalancingSchedulers::new(&apalis_pool),
                 telemetry: TelemetrySender::disabled(),
+                notifier: Arc::new(NoopNotifier),
             },
         )
         .await
@@ -10208,7 +10213,7 @@ mod tests {
         // Monitor returns an error after shutdown signal.
         let monitor = tokio::spawn(async move {
             apalis_token_clone.cancelled().await;
-            Err(MonitorTaskError::TerminalJobFailure)
+            Err(MonitorTaskError::UnexpectedExit { source: None })
         });
 
         let job_cleanup = tokio::spawn(pending::<()>());
@@ -10238,10 +10243,13 @@ mod tests {
     #[test]
     fn check_monitor_drain_result_propagates_monitor_error() {
         let error =
-            check_monitor_drain_result(Ok(Err(MonitorTaskError::TerminalJobFailure))).unwrap_err();
+            check_monitor_drain_result(Ok(Err(MonitorTaskError::UnexpectedExit { source: None })))
+                .unwrap_err();
 
         assert!(
-            error.to_string().contains("Apalis worker failed"),
+            error
+                .to_string()
+                .contains("Apalis monitor exited unexpectedly"),
             "should propagate MonitorTaskError, got: {error}"
         );
     }
@@ -10674,18 +10682,18 @@ mod tests {
         );
     }
 
-    /// When `[alerts]` is absent, `build_usdc_notifier` returns a `NoopNotifier`
+    /// When `[alerts]` is absent, `build_operational_notifier` returns a `NoopNotifier`
     /// that silently discards notifications without error.
     #[tokio::test]
-    async fn build_usdc_notifier_returns_ok_noop_when_alerts_absent() {
-        let notifier = build_usdc_notifier(None).unwrap();
+    async fn build_operational_notifier_returns_ok_noop_when_alerts_absent() {
+        let notifier = build_operational_notifier(None).unwrap();
         notifier
             .notify("test message")
             .await
             .expect("NoopNotifier must not error on notify");
     }
 
-    /// When `[alerts]` IS present, `build_usdc_notifier` constructs a real
+    /// When `[alerts]` IS present, `build_operational_notifier` constructs a real
     /// Telegram notifier and returns `Ok` -- it must NOT fail startup for a
     /// well-formed config, and it must NOT silently fall back to `NoopNotifier`
     /// (which would suppress every redrive-limit and terminal-error page).
@@ -10697,7 +10705,7 @@ mod tests {
     /// `notify()` is deliberately not exercised: the present-branch notifier posts
     /// to the live Telegram API, which a unit test must never reach.
     #[tokio::test]
-    async fn build_usdc_notifier_returns_ok_telegram_when_alerts_present() {
+    async fn build_operational_notifier_returns_ok_telegram_when_alerts_present() {
         let alerts = st0x_config::AlertsCtx {
             chat_id: 123,
             bot_token: "test-bot-token".to_string(),
@@ -10707,7 +10715,7 @@ mod tests {
             message_thread_id: Some(42),
         };
 
-        build_usdc_notifier(Some(&alerts))
+        build_operational_notifier(Some(&alerts))
             .expect("a well-formed [alerts] config must yield a notifier, not a startup error");
     }
 }
--- a/src/conductor/job.rs
+++ b/src/conductor/job.rs
@@ -11,21 +11,232 @@ use apalis_core::backend::TaskSinkError;
 use apalis_core::backend::poll_strategy::{BackoffConfig, IntervalStrategy, StrategyBuilder};
 use apalis_core::worker::context::WorkerContext;
 use apalis_core::worker::event::Event;
-use apalis_sqlite::{Config, SqliteContext, SqlitePool, SqliteStorage, SqlxError};
+use apalis_sqlite::{Config, SqliteContext, SqlitePool, SqliteStorage, SqlxError, TaskBuilderExt};
 use serde::Serialize;
 use serde::de::DeserializeOwned;
 use std::fmt;
+use std::num::NonZeroU32;
 use std::sync::Arc;
-#[cfg(any(test, feature = "test-support"))]
 use std::sync::Mutex;
-use std::time::Duration;
-use tracing::{debug, error, warn};
+use std::time::{Duration, Instant};
+use tracing::{debug, error, info, warn};
 
-/// Recovery timeout for the fail-stop circuit breaker. Effectively
-/// infinite for any plausible bot uptime; chosen to be finite so
-/// `last_failure + recovery_timeout` cannot overflow if apalis
-/// changes its check from `elapsed() >=` to addition.
-pub(crate) const FAIL_STOP_RECOVERY_TIMEOUT: Duration = Duration::from_secs(60 * 60 * 24 * 365);
+use crate::alerts::Notifier;
+
+/// Production policy for workers whose terminal systemic failures must pause
+/// new work and recover without restarting the process.
+#[cfg(not(any(test, feature = "test-support")))]
+pub(crate) const WORKER_CIRCUIT_POLICY: WorkerCircuitPolicy = WorkerCircuitPolicy::new(
+    NonZeroU32::MIN,
+    Duration::from_secs(5 * 60),
+    Duration::from_secs(60 * 60),
+);
+
+#[cfg(any(test, feature = "test-support"))]
+pub(crate) const WORKER_CIRCUIT_POLICY: WorkerCircuitPolicy = WorkerCircuitPolicy::new(
+    NonZeroU32::MIN,
+    Duration::from_millis(10),
+    Duration::from_secs(60 * 60),
+);
+
+pub(crate) const JOB_RETRIES: usize = 3;
+const JOB_MAX_ATTEMPTS: u32 = 4;
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct WorkerCircuitPolicy {
+    failure_threshold: NonZeroU32,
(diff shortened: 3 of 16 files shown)


0xgleb:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1046?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#936** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/936?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#939](https://github.com/ST0x-Technology/st0x.liquidity/pull/939) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/939?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1020](https://github.com/ST0x-Technology/st0x.liquidity/pull/1020) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1020?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#930** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/930?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#929** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/929?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#924** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/924?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#923** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/923?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#922** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/922?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1042** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1042?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1039** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1039?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1024** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1024?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1023** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1023?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1030** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1030?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1028** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1028?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1046** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1046?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1046?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1040** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1040?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1061** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1061?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1055** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1055?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1046).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/conductor/job.rs
```diff
@@ -128,7 +339,10 @@ impl<Task: Serialize + DeserializeOwned + Send + Sync + Unpin + 'static> JobQueu
     }
 
     pub(crate) async fn push(&mut self, task: Task) -> Result<(), QueuePushError> {
```
JuaniRios, line 383: should fix: `push` and `push_with_delay` set `max_attempts = JOB_MAX_ATTEMPTS` only on rows they enqueue. Jobs rows that are already durable at deploy time keep the old apalis value. When the in-memory retry middleware stops after 4 attempts, apalis writes such a row as `Failed`, not `Killed`, because `attempts < max_attempts` is still true, and `fetch_next.sql` re-selects it. The new handler treats every `Event::Error` as terminal, so one old row can open the circuit, alert, and re-run its side effects on each cooldown cycle. That breaks the one-terminal-transition contract this PR states. Please normalize `Jobs.max_attempts` for non-terminal rows at startup, or add a migration, and leave true dead letters alone.

src/conductor/job.rs
```diff
@@ -937,33 +1137,162 @@ mod tests {
...
+        assert!(
+            worker.is_paused(),
+            "two consecutive failures must open the circuit"
+        );
+    }
+
+    /// A terminal systemic failure pauses the worker, alerts, and then resumes
+    /// sibling processing after the configured cooldown.
+    #[tokio::test]
+    async fn supervised_worker_recovers_after_terminal_failure() {
```
JuaniRios, line 1440: should fix: the main benefit this PR claims is that only the failing worker pauses while sibling workers stay live. No test proves that. `supervised_worker_recovers_after_terminal_failure` uses one worker and one job type, so it only shows the same worker resumes. The e2e tests assert `!bot.is_finished()`, which shows the process is alive but not that a healthy job type kept making progress. A regression that reintroduces a shared stop signal passes all of them. Please add a test that registers two independent supervised workers on one monitor, drives one to a terminal failure, and asserts the other keeps draining its own queue during the cooldown.

src/conductor/job.rs
```diff
@@ -11,21 +11,232 @@ use apalis_core::backend::TaskSinkError;
...
+                Err(error) => warn!(
+                    worker = %worker.name(),
+                    ?error,
+                    "Worker circuit cooldown elapsed but the worker could not resume"
+                ),
+            }
+        });
+    }
+
+    fn should_alert(&self, now: Instant) -> bool {
```
JuaniRios, line 191: should fix: `should_alert` writes `*last_alerted = Some(now)` and returns true before delivery happens. `alert` then spawns a detached task, and a failed `notify()` only produces a warning. There is no retry and no rollback. Production uses a 5 minute cooldown and a 1 hour re-alert interval, so a systemic failure reopens the circuit about 12 times per hour, and the timestamp installed by the failed send suppresses every one of them. One transient Telegram error therefore gives zero operator pages for an hour while the worker keeps killing a job per cycle. Please write `last_alerted` only after `notify()` succeeds, or reset it to `None` when delivery fails.

src/conductor/job.rs
```diff
@@ -758,6 +953,11 @@ mod tests {
         );
     }
 
+    #[test]
```
JuaniRios: should fix: SPEC.md and docs/conductor.md state the shared four-attempt limit as fact, but it depends on how the apalis release-candidate crates interpret `max_attempts`. If a later rc treats that value as additional retries, apalis writes `Failed` with `attempts(4) < max_attempts(5)` and `fetch_next.sql` re-selects the same row. This test compares two local constants, so it cannot detect that. Please assert the real durable outcome in `supervised_worker_recovers_after_terminal_failure`: read the row through `job_rows_for_assertion` and check `status == Status::Killed.to_string()` and `attempts == JOB_MAX_ATTEMPTS`.

src/conductor/job.rs
```diff
@@ -758,6 +953,11 @@ mod tests {
...
+    fn durable_attempt_limit_matches_middleware_retry_budget() {
+        assert_eq!(usize::try_from(JOB_MAX_ATTEMPTS).unwrap(), JOB_RETRIES + 1,);
```
JuaniRios: nit: this test reads two literal constants that are defined a few lines apart and compares them to each other. It never exercises the retry middleware or the durable `max_attempts` column. If both numbers change together for the wrong reason, the test still passes. Please either delete it, because the integration test already covers the real invariant, or raise it to an integration assertion on the `attempts` column at the moment the row goes terminal.

SPEC.md
```diff
@@ -310,6 +310,17 @@ configured `apalis_finished_job_cleanup_interval_secs` cadence so the durable
 queue cannot grow without bound during normal operation or repeated restarts.
 The cadence must be explicitly configured and non-zero.
 
+Supervised apalis workers isolate terminal failures with a local recovering
```
JuaniRios, line 313: should fix: this paragraph says the exhausted job remains visible as a terminal Jobs row, but the paragraph above says the runtime deletes terminal rows on the `apalis_finished_job_cleanup_interval_secs` cadence. `spawn_finished_job_cleanup` deletes `Killed` rows and excludes only the four transfer job types. Aligning `max_attempts` to 4 makes an exhausted job land in `Killed`, so it now matches that cleanup predicate. In production the row is swept after the configured interval, so the diagnostic artifact the spec promises is gone when on-call looks for it. Please either exclude these terminal rows from cleanup, or state in the spec that the row survives only until the cleanup cadence.

tests/e2e/chaos_alpaca.rs
```diff
@@ -589,25 +591,34 @@ async fn broker_outage_with_submitted_order_halts_bot_and_restart_recovers() ->
...
+        !bot.is_finished(),
+        "a terminal status-poll job must not stop the bot",
     );
 
     // Broker comes back; let the order fill on the next poll.
     infra
         .broker_service
         .set_symbol_fill_delay(Symbol::new(equity_symbol)?, 0);
     latency.restore().await?;
+    bot.abort();
```
JuaniRios: should fix: the test used to await the bot's real exit before it started the replacement, so only one conductor touched the database at a time. `bot.abort()` cancels the outer `JoinHandle` only. `run_bot_session_inner` spawns the server supervisor and `run_conductor_session`, and dropping those handles detaches them instead of aborting them. The first bot's monitor, supervised workers, and inventory poller keep running against the same database. Under `test-support` the cooldown is 10 ms, so the first bot's paused worker resumes almost at once. The `order_count == 1` assertion is then decided by a race between two live conductors, and it can pass because the old bot recovered. Please shut the first bot down through its real path and await its handle.

tests/e2e/poll.rs
```diff
@@ -518,6 +518,51 @@ pub async fn poll_for_running_job(
...
+
+        let Ok(pool) = SqlitePool::connect_with(connect_opts.clone()).await else {
+            assert!(
+                tokio::time::Instant::now() < deadline,
+                "Timed out after {timeout:?} waiting for {context} (database not ready)",
+            );
+            continue;
+        };
+
+        let terminal: Result<(i64,), _> = sqlx::query_as(
```
JuaniRios, line 553: minor: this query hard-codes apalis-sqlite's private schema and the `'Killed'` and `'Failed'` string literals. The repo already has a typed precedent: `wait_for_terminal_test_job` in src/conductor/job.rs binds `Status::Killed.to_string()`. If an rc bump renames a column or a status variant, this predicate silently matches nothing and the test fails as a timeout instead of a schema mismatch. Please bind the status values from `apalis_core::task::status::Status`.

src/conductor/job.rs
```diff
@@ -11,21 +11,232 @@ use apalis_core::backend::TaskSinkError;
...
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct WorkerCircuitPolicy {
+    failure_threshold: NonZeroU32,
+    recovery_timeout: Duration,
+    realert_interval: Duration,
+}
+
+impl WorkerCircuitPolicy {
+    pub(crate) const fn new(
```
JuaniRios: minor: `recovery_timeout` and `realert_interval` are adjacent `Duration` parameters with very different meanings, and every call site passes two `Duration::from_secs(...)` literals positionally. A swapped pair compiles and silently inverts the policy: recovery after 60 minutes and re-alerts every 5 minutes. Please use two newtypes, or named-field construction, so a swap becomes a compile error.

src/conductor/job.rs
```diff
@@ -11,21 +11,232 @@ use apalis_core::backend::TaskSinkError;
...
+        *state = WorkerCircuitState::Open;
+        drop(state);
+        Some(consecutive_failures)
+    }
+
+    fn schedule_recovery(&self, worker: WorkerContext) {
+        let recovery_timeout = self.policy.recovery_timeout;
+        let state = self.state.clone();
+        tokio::spawn(async move {
+            tokio::time::sleep(recovery_timeout).await;
```
JuaniRios, line 147: minor: the recovery task sets the state to `Closed` before it calls `worker.resume()`. `resume()` returns an error when the worker is not paused or is shutting down. On that path the code only warns and returns, so the circuit is `Closed` but the worker is not running. A worker that does not run never dequeues a job, so it never emits another `Event::Error` and never reopens the circuit. It stays latched idle with one warning line, no alert, and no `worker_circuit_transition` field. That is the failure mode this PR removes elsewhere. Please set `Closed` only after `resume()` succeeds, and on the error path keep the circuit open, log at error level, and page the operator.

src/conductor/job.rs
```diff
@@ -11,21 +11,232 @@ use apalis_core::backend::TaskSinkError;
...
+        };
+
+        let consecutive_failures = consecutive_failures.saturating_add(1);
+        if consecutive_failures < self.policy.failure_threshold.get() {
+            *state = WorkerCircuitState::Closed {
+                consecutive_failures,
+            };
+            return None;
+        }
+
```
JuaniRios, line 117: nit: this explicit `drop(state)` has no effect. The guard is not used again, and the alert, pause, and recovery calls all happen in the caller. It reads as if early release matters here, which it does not. Please remove the line and let the guard fall out of scope.

JuaniRios (commented):
Solid direction and the recovery path is well tested in isolation, but two gaps matter before merge. Jobs rows already durable at deploy time keep the old max_attempts, so an exhausted old row lands in Failed rather than Killed and re-runs its side effects on every cooldown, which breaks the one-terminal-transition contract. And nothing proves the headline claim: no test shows a sibling worker still making progress while another worker's circuit is open.

JuaniRios (changes requested):
Requesting changes. Blocking items, both in the inline comments:

- Jobs rows already durable at deploy time keep the old `max_attempts`, so an exhausted old row lands in `Failed` rather than `Killed`, gets re-fetched, and re-runs its side effects on every cooldown. That breaks the one-terminal-transition contract this PR states.
- No test proves the headline claim. Nothing shows a sibling worker still making progress while another worker's circuit is open, so a regression to a shared stop signal would pass CI.

The alerting gap (`should_alert` records the page before delivery, so one failed send suppresses operator pages for an hour) and the SPEC contradiction about the terminal Jobs row surviving cleanup should be resolved in the same pass.

src/conductor/job.rs
0xgleb: Addressed. Startup now normalizes max_attempts for every pre-existing Pending, Queued, Running, or retryable Failed row while preserving true terminal rows. The regression fixture includes a legacy Failed row already at the new four-attempt limit and verifies it is normalized.

src/conductor/job.rs
0xgleb: Addressed. The new two-worker monitor test drives one queue to Killed, enqueues work on an independent healthy queue during the failing worker cooldown, and proves that sibling completes while the monitor remains live.

src/conductor/job.rs
0xgleb: Addressed. Failed notification delivery now clears the reserved last_alerted timestamp, so the next circuit opening can page immediately. A failing-notifier test waits for the rollback and verifies the alert becomes eligible again.

src/conductor/job.rs
0xgleb: Addressed. The supervised worker integration test now reads the durable Jobs row and asserts Status::Killed plus attempts and max_attempts both equal the configured four-attempt limit.

src/conductor/job.rs
0xgleb: Addressed. The constant-only test was removed; the retry invariant is now verified against the durable apalis row in the integration test.

SPEC.md
0xgleb: Addressed. SPEC now states that the terminal Jobs row remains visible only until the configured cleanup cadence, matching the cleanup implementation.

tests/e2e/chaos_alpaca.rs
0xgleb: Addressed. The chaos test now starts both bot sessions with an explicit cancellation token routed through the production graceful-shutdown lifecycle. It cancels and awaits the first conductor and server drain before starting the replacement, and gracefully drains the replacement at test end. The session drop guard is also disarmed after a successful drain so it cannot issue a duplicate shutdown.

tests/e2e/poll.rs
0xgleb: Addressed. The shared e2e polling helper now binds apalis Status::Killed and Status::Failed rather than hard-coded status strings.

src/conductor/job.rs
0xgleb: Addressed. WorkerCircuitPolicy is now constructed with named fields everywhere, so recovery_timeout and realert_interval cannot be silently swapped positionally.

src/conductor/job.rs
0xgleb: Addressed. Recovery now closes the circuit only after resume succeeds or the worker is confirmed running. A real resume failure leaves it open, logs a structured recovery_failed transition at error level, and alerts the operator; the new test verifies all three outcomes.

src/conductor/job.rs
0xgleb: Addressed. The explicit drop was removed by scoping the lock guard to the state transition block, so it falls out of scope naturally before returning.

src/conductor/job.rs
0xgleb: Addressed by normalizing max_attempts during apalis startup for Pending, Queued, Running, and retryable Failed rows while preserving true terminal rows. The startup regression covers stale retryable and terminal rows and asserts only live work is moved to JOB_MAX_ATTEMPTS.

src/conductor/job.rs
0xgleb: Addressed with supervised_worker_failure_does_not_pause_a_healthy_sibling_worker. It registers independent failing and healthy workers on one Monitor, waits for the first job to become terminal, queues healthy work during the failing worker’s five-second cooldown, and proves the sibling drains it while the monitor remains live.

src/conductor/job.rs
0xgleb: Addressed by clearing last_alerted when notifier delivery fails, so the next circuit opening can page immediately rather than waiting for the re-alert interval. worker_circuit_delivery_failure_allows_immediate_alert_retry pins the failed-delivery path.

src/conductor/job.rs
0xgleb: Addressed in supervised_worker_recovers_after_terminal_failure: the test now reads the durable Jobs row and asserts Status::Killed plus attempts == JOB_MAX_ATTEMPTS and max_attempts == JOB_MAX_ATTEMPTS after the real retry middleware exhausts.

src/conductor/job.rs
0xgleb: Addressed by deleting the constant-vs-constant test. The retry-budget invariant is now exercised through the supervised worker integration test and asserted on the durable Killed row’s attempts and max_attempts columns.

SPEC.md
0xgleb: Addressed in SPEC.md: an exhausted row is explicitly documented as remaining visible only until the configured terminal-job cleanup cadence, consistent with spawn_finished_job_cleanup.

tests/e2e/chaos_alpaca.rs
0xgleb: Addressed by replacing the outer-handle abort in this restart test with spawn_bot_until_cancelled, cancellation through the real shutdown signal, and awaiting the first session to finish before starting its replacement. SessionTaskGuard also aborts conductor/server children if the session future itself is cancelled.

tests/e2e/poll.rs
0xgleb: Addressed by binding apalis_core::task::status::Status::Killed and Status::Failed string values in poll_for_terminal_job instead of hard-coding private status literals; the query remains scoped to the concrete job_type.

src/conductor/job.rs
0xgleb: Addressed by removing the positional constructor and using named WorkerCircuitPolicy fields at every call site, making recovery_timeout and realert_interval explicit rather than adjacent anonymous Duration arguments.

src/conductor/job.rs
0xgleb: Addressed by transitioning the circuit to Closed only after resume succeeds (or the worker is already running). A genuine resume failure leaves it Open, logs worker_circuit_transition=recovery_failed at error level, and pages; the dedicated recovery-failure test pins all three outcomes.

src/conductor/job.rs
0xgleb: Addressed without restoring the misleading explicit drop: record_failure now uses a lexical scope for the mutex guard, so it releases before constructing the return value and also satisfies the denied significant-drop-tightening lint.

src/conductor.rs
```diff
@@ -1279,16 +1297,16 @@ impl PositionAndRebalancing {
...
-        debug!("USDC alerting: [alerts] section absent, using NoopNotifier");
+        debug!("Operational alerting: [alerts] section absent, using NoopNotifier");
         return Ok(Arc::new(NoopNotifier));
     };
     let notifier =
         TelegramNotifier::new(&alerts.bot_token, alerts.chat_id, alerts.message_thread_id)?;
-    info!("USDC alerting: Telegram notifier configured");
+    info!("Operational alerting: Telegram notifier configured");
     Ok(Arc::new(notifier))
 }
```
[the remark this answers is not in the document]
0xgleb: Verified against the current contract and not changed: SPEC.md’s Operational Alerting section explicitly defines alerting as optional and says the bot runs normally with no channel when configuration is absent. AlertsCtx::assemble returns None only when both config and secrets are absent, rejects either half-configured shape, and NoopNotifier makes intentional absence typed and startup-logged. Making [alerts] mandatory would contradict that accepted contract and this review’s sibling request to make circuit-alert claims conditional. Configured Telegram initialization already fails startup through NotifierError.

src/conductor/job.rs
```diff
@@ -11,21 +11,244 @@ use apalis_core::backend::TaskSinkError;
...
+        let mut state = self
+            .state
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        if matches!(*state, WorkerCircuitState::Closed { .. }) {
+            *state = WorkerCircuitState::Closed {
+                consecutive_failures: 0,
+            };
+        }
+    }
```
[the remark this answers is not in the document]
0xgleb: Replaced the production matches! check with an exhaustive Closed/Open match; only Closed resets its failure count.

src/conductor/job.rs
```diff
@@ -11,21 +11,244 @@ use apalis_core::backend::TaskSinkError;
...
+
+#[cfg(any(test, feature = "test-support"))]
+pub(crate) const WORKER_CIRCUIT_POLICY: WorkerCircuitPolicy = WorkerCircuitPolicy {
+    failure_threshold: NonZeroU32::MIN,
+    recovery_timeout: Duration::from_millis(10),
+    realert_interval: Duration::from_secs(60 * 60),
+};
+
+pub(crate) const JOB_RETRIES: usize = 3;
+pub(crate) const JOB_MAX_ATTEMPTS: u32 = 4;
```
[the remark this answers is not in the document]
0xgleb: Derived JOB_MAX_ATTEMPTS from JOB_RETRIES; retries now use a non-lossy u16-to-usize conversion under strict clippy.

PR #1048, opened 2026-07-16
docs: supersede guard-only USDC recovery ADR (RAI-1324)

0xgleb:
## What

[RAI-1324](https://linear.app/makeitrain/issue/RAI-1324)

Accept ADR 0015 for reconstructing pre-burn Alpaca-to-Base USDC inventory before resume, supersede ADR 0003’s guard-only decision, and update the ADR index.

## Why

ADR 0003 assumed interrupted USDC transfers had no automatic resume consumer. Startup recovery now re-arms resumable transfers, so guard-only recovery can resume without the original owner and source reservation, causing a double debit or terminal inflight underflow.

## How

The accepted decision:

- gates pre-burn resume on a fresh Hedging observation;
- restores exact ownership and inflight state with replacement rather than additive accounting;
- treats `None` as a fresh transfer;
- permits conservative source-reservation reconstruction only before a recorded burn transaction;
- fails closed on missing observations, conflicting owners, ambiguous candidates, and post-burn uncertainty; and
- requires explicit transaction adoption before any potentially repeated burn submission.

## Testing

Documentation-only change:

- repository Markdown formatting passes;
- pre-commit checks pass; and
- the ADR contains a state-by-state recovery and acceptance matrix consumed by #1050’s regression coverage.

## Screenshots

Not applicable.

## Anything else

ADR 0015 is **Accepted** and ADR 0003 is **Superseded**. The implementation is provided by the next PR in the stack, #1050. Post-burn reconstruction remains a separate follow-up requiring transaction-status and destination-settlement evidence.

PR #1050, opened 2026-07-16
fix: reconstruct resumed USDC reservation (RAI-1324)

0xgleb:
## What

[RAI-1324](https://linear.app/makeitrain/issue/RAI-1324)

Implement ADR 0015 by reconstructing the exact pre-burn Alpaca-to-Base USDC source reservation before a resumed transfer re-enters the transfer manager.

## Why

Startup recovery already re-armed resumable USDC jobs, but their in-memory Hedging reservation and ownership were lost across restart. Resuming against a fresh broker snapshot could debit the source twice, conflict with another owner, or settle with an inflight underflow.

## How

- Derive a typed reservation-recovery context from supported Alpaca-to-Base aggregate states.
- Wait for a fresh Hedging inventory observation before releasing the resume gate.
- Restore the exact owner, tracking context, available balance, and inflight amount with replacement semantics rather than an additive debit.
- Fail closed on conflicting owners, mismatched amounts, multiple recovery candidates, source-observation failures, and recorded burn submissions.
- Retire stale resume gates after timeout so later recovery can proceed.

## Testing

Added focused coverage for:

- fresh snapshot callbacks even when the snapshot value is deduplicated;
- replacement of stale available/inflight inventory;
- timeout cleanup of the recovered resume gate;
- no second debit after `ConversionComplete` recovery;
- owner and amount conflicts;
- multiple reservation candidates;
- terminal settlement without inflight underflow;
- transfer-job blocking until reservation recovery completes; and
- exclusion of `BridgingSubmitting` with a recorded burn transaction.

Verification:

- `cargo fmt --all -- --check`
- GitHub Actions backend build and partitioned test matrix

## Screenshots

Not applicable.

## Anything else

ADR 0015 is Accepted and supersedes the earlier guard-only recovery decision in ADR 0003. Recovery beyond the burn-submission boundary remains fail-closed and requires separate transaction and destination-settlement evidence.

### DIFF
--- a/src/inventory/polling.rs
+++ b/src/inventory/polling.rs
@@ -9,7 +9,7 @@
 use alloy::primitives::{Address, B256, TxHash};
 use alloy::providers::RootProvider;
 use async_trait::async_trait;
-use chrono::Utc;
+use chrono::{DateTime, Utc};
 use futures_util::future::try_join_all;
 use rain_math_float::FloatError;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
@@ -50,6 +50,18 @@ pub(crate) trait PendingRequestOwnership: Send + Sync {
     async fn pending_request_ownership(&self) -> PendingRequestOwnershipSnapshot;
 }
 
+/// Consumer for a freshly fetched, reserve-adjusted offchain cash balance.
+/// Unlike snapshot reactors, this callback runs even when the aggregate
+/// deduplicates an unchanged balance.
+#[async_trait]
+pub(crate) trait FreshOffchainUsdObserver: Send + Sync {
+    async fn observe_fresh_offchain_usd(
+        &self,
+        available: Usdc,
+        observed_at: DateTime<Utc>,
+    ) -> Result<(), Box<dyn std::error::Error + Send + Sync>>;
+}
+
 /// Error type for inventory polling operations.
 #[derive(Debug, thiserror::Error)]
 pub(crate) enum InventoryPollingError<ExecutorError> {
@@ -78,6 +90,8 @@ pub(crate) enum InventoryPollingError<ExecutorError> {
     SharesConversion(#[from] SharesConversionError),
     #[error(transparent)]
     Reserve(#[from] ReserveError),
+    #[error("fresh offchain USD observer failed: {0}")]
+    FreshOffchainUsdObserver(#[source] Box<dyn std::error::Error + Send + Sync>),
 }
 
 /// Errors that can occur when computing available cash after subtracting
@@ -92,6 +106,8 @@ pub(crate) enum ReserveError {
     CentsConversion(#[from] UsdToCentsError),
     #[error("failed to convert broker cents {cents} to Usd")]
     BrokerCentsConversion { cents: i64 },
+    #[error("failed to convert available broker cents {cents} to Usdc")]
+    AvailableUsdcConversion { cents: i64 },
 }
 
 pub(crate) struct WalletPollingCtx {
@@ -120,6 +136,7 @@ where
     wallet_polling: Option<WalletPollingCtx>,
     tokenizer: Option<Arc<dyn Tokenizer>>,
     pending_request_ownership: Option<Arc<dyn PendingRequestOwnership>>,
+    fresh_offchain_usd_observer: Option<Arc<dyn FreshOffchainUsdObserver>>,
     external_pending_warnings: Mutex<HashSet<TokenizationRequestId>>,
     unconfigured_symbol_warnings: Mutex<HashSet<Symbol>>,
     retired_equity_vault_warnings: Mutex<HashSet<(Address, B256)>>,
@@ -157,6 +174,7 @@ where
             wallet_polling,
             tokenizer,
             pending_request_ownership: None,
+            fresh_offchain_usd_observer: None,
             external_pending_warnings: Mutex::new(HashSet::new()),
             unconfigured_symbol_warnings: Mutex::new(HashSet::new()),
             retired_equity_vault_warnings: Mutex::new(HashSet::new()),
@@ -191,6 +209,14 @@ where
         pending_request_ownership: Arc<dyn PendingRequestOwnership>,
     ) -> Self {
         self.pending_request_ownership = Some(pending_request_ownership);
+        self
+    }
+
+    pub(crate) fn with_fresh_offchain_usd_observer(
+        mut self,
+        observer: Arc<dyn FreshOffchainUsdObserver>,
+    ) -> Self {
+        self.fresh_offchain_usd_observer = Some(observer);
         self
     }
 
@@ -540,6 +566,18 @@ where
                 },
             )
             .await?;
+
+        if let Some(observer) = &self.fresh_offchain_usd_observer {
+            let available = Usdc::from_cents(available_usd_cents).ok_or(
+                ReserveError::AvailableUsdcConversion {
+                    cents: available_usd_cents,
+                },
+            )?;
+            observer
+                .observe_fresh_offchain_usd(available, Utc::now())
+                .await
+                .map_err(InventoryPollingError::FreshOffchainUsdObserver)?;
+        }
 
         self.snapshot
             .send(
@@ -996,6 +1034,21 @@ mod tests {
     impl PendingRequestOwnership for TestPendingRequestOwnership {
         async fn pending_request_ownership(&self) -> PendingRequestOwnershipSnapshot {
             self.0.clone()
+        }
+    }
+
+    #[derive(Default)]
+    struct RecordingFreshOffchainUsdObserver(Mutex<Vec<Usdc>>);
+
+    #[async_trait]
+    impl FreshOffchainUsdObserver for RecordingFreshOffchainUsdObserver {
+        async fn observe_fresh_offchain_usd(
+            &self,
+            available: Usdc,
+            _observed_at: DateTime<Utc>,
+        ) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+            self.0.lock().unwrap().push(available);
+            Ok(())
         }
     }
 
@@ -1457,6 +1510,56 @@ mod tests {
         assert_eq!(
             *usd_balance_cents, 25_000_000,
             "Cash balance mismatch: expected $250,000.00"
+        );
+    }
+
+    #[tokio::test]
+    async fn fresh_offchain_usd_observer_runs_when_snapshot_deduplicates_balance() {
+        let pool = setup_test_db().await;
+        let provider = mock_provider();
+        let raindex_service = create_test_raindex_service(provider.clone());
+        let (orderbook, order_owner) = test_addresses();
+        let observer = Arc::new(RecordingFreshOffchainUsdObserver::default());
+        let executor = MockExecutor::new().with_inventory(Inventory {
+            positions: vec![],
+            usd_balance_cents: 25_000_000,
+            cash_buying_power_cents: Some(25_000_000),
+            alpaca_usdc: None,
+            cash_withdrawable_cents: None,
+        });
+        let service = InventoryPollingService::new(
+            raindex_service,
+            executor,
+            Arc::new(test_store::<VaultRegistry>(pool.clone(), ())),
+            InventorySnapshotId {
+                orderbook,
+                owner: order_owner,
+            },
+            order_owner,
+            Arc::new(test_store(pool.clone(), ())),
+            None,
+            None,
+            Usd::ZERO,
+        )
+        .with_fresh_offchain_usd_observer(observer.clone());
+
+        service.poll_and_record().await.unwrap();
+        service.poll_and_record().await.unwrap();
+
+        let observations = observer.0.lock().unwrap().clone();
+        assert_eq!(
+            observations.as_slice(),
+            &[Usdc::new(float!(250000)), Usdc::new(float!(250000))],
+            "each successful broker fetch must cross the recovery freshness barrier",
+        );
+        let snapshot_events = load_snapshot_events(&pool, orderbook, order_owner).await;
+        assert_eq!(
+            snapshot_events
+                .iter()
+                .filter(|event| matches!(event, InventorySnapshotEvent::OffchainUsd { .. }))
+                .count(),
+            1,
+            "the recovery callback must not disable persisted snapshot deduplication",
         );
     }
 
--- a/src/rebalancing/trigger/mod.rs
+++ b/src/rebalancing/trigger/mod.rs
@@ -40,8 +40,9 @@ use crate::inventory::projection::InventoryProjectionError;
 use crate::inventory::snapshot::{InventorySnapshot, InventorySnapshotEvent};
 use crate::inventory::view::InFlightEquityLocation;
 use crate::inventory::{
-    BroadcastingInventory, ImbalanceThreshold, Inventory, InventoryView, InventoryViewError,
-    Operator, PendingRequestOwnership, PendingRequestOwnershipSnapshot, TransferOp, Venue,
+    BroadcastingInventory, FreshOffchainUsdObserver, ImbalanceThreshold, Inventory, InventoryView,
+    InventoryViewError, Operator, PendingRequestOwnership, PendingRequestOwnershipSnapshot,
+    TransferOp, Venue,
 };
 use crate::position::{Position, PositionEvent};
 use crate::rebalancing::equity::{
@@ -49,6 +50,7 @@ use crate::rebalancing::equity::{
     TransferEquityToMarketMakingJobQueue,
 };
 use crate::rebalancing::usdc::{
+    AlpacaToBaseResumePreparation, AlpacaToBaseResumePreparationError, PrepareAlpacaToBaseResume,
     TransferUsdcToHedging, TransferUsdcToHedgingJobQueue, TransferUsdcToMarketMaking,
     TransferUsdcToMarketMakingJobQueue,
 };
@@ -60,8 +62,8 @@ use crate::unwrapped_equity_recovery::{
     UnwrappedEquityRecoveryJob, UnwrappedEquityRecoveryJobQueue,
 };
 use crate::usdc_rebalance::{
-    InterruptedUsdcRebalances, RebalanceDirection, UsdcRebalance, UsdcRebalanceEvent,
-    UsdcRebalanceId, interrupted_usdc_rebalance_ids,
+    AlpacaToBaseReservationRecovery, InterruptedUsdcRebalances, RebalanceDirection, UsdcRebalance,
+    UsdcRebalanceEvent, UsdcRebalanceId, interrupted_usdc_rebalance_ids,
 };
 use crate::vault_registry::{VaultRegistry, VaultRegistryId};
 use crate::wrapped_equity_recovery::aggregate::WrappedEquityRecoveryId;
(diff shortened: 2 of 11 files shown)


0xgleb:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1050?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1050** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1050?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1050?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1050).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/rebalancing/trigger/mod.rs
```diff
@@ -2029,6 +2082,107 @@ impl RebalancingService {
...
+        match self.alpaca_to_base_resume_gate.read().await.as_ref() {
+            Some(AlpacaToBaseResumeGate::AwaitingFreshHedgingSnapshot {
+                id: recovered_id, ..
+            }) if recovered_id == id => {
+                Ok(AlpacaToBaseResumePreparation::AwaitingFreshHedgingSnapshot)
+            }
+            Some(AlpacaToBaseResumeGate::ReservationRestored {
+                id: recovered_id, ..
+            }) if recovered_id == id => Ok(AlpacaToBaseResumePreparation::Ready),
+            Some(gate) => Err(
```
JuaniRios: should fix: this arm returns `ConflictingRecoveryOwner` for any id that is not the gate owner. The gate covers one aggregate's source reservation, but `TransferUsdcToMarketMaking::perform` runs this check for every Alpaca-to-Base job, including the post-burn jobs that `recover_usdc_guard` re-arms in the same startup pass. `resumable_post_burn_transfer` returns `Some` for `Bridging`, `AwaitingAttestation`, and `Attested`, and `rearm_stranded_transfers` pushes the Alpaca-to-Base ones onto the same queue. So one aggregate stranded at `AwaitingAttestation` plus one pre-burn aggregate that owns the gate means the post-burn job fails on every attempt, burns its retry budget, and lands in a terminal `Failed` row that is never re-armed. Its USDC is already burned in CCTP. This case does not trip `MultipleAlpacaToBaseReservationRecoveries`, because only one of the two is a reservation candidate. Please return `Ready` for any id that does not own the gate. Owner conflicts are already failed closed against the inventory in `observe_fresh_offchain_usd`.

src/rebalancing/trigger/mod.rs
```diff
@@ -2029,6 +2082,107 @@ impl RebalancingService {
...
+        *inventory = inventory
+            .clone()
+            .update_usdc(
+                Inventory::restore_reservation(Venue::Hedging, available, recovery.amount),
+                observed_at,
+            )?
+            .set_active_usdc_rebalance(id.clone());
+        drop(inventory);
+
+        let stage = match recovery.stage {
```
JuaniRios: should fix: this maps a recovered `BridgingSubmitting` to `UsdcRebalanceStage::BridgingInitiated`, but that recovery only fires when `pending_burn_tx` is `None`, so no burn is submitted yet. `UsdcRebalanceTracking::is_post_burn()` treats `BridgingInitiated` as post-burn for `AlpacaToBase`. This fabricates a post-burn tracking stage for an aggregate that is durably still pre-burn, and it contradicts the invariant stated in `post_burn_failure`: pre-burn `BridgingFailed` is always pre-burn, so absent tracking correctly clears. That invariant held before this PR because pre-burn states were never seeded with tracking. Please map this case to `UsdcRebalanceStage::WithdrawalConfirmed`, the last accurate pre-burn stage, and let the real `BridgingInitiated` event set the post-burn stage through the normal reactor path.

src/rebalancing/trigger/mod.rs
```diff
@@ -3883,6 +4042,27 @@ impl RebalancingService {
             }
         }
 
+        if reservation_recoveries.len() > 1 {
```
JuaniRios: should fix: this returns an error before `rearm_stranded_transfers` and before `usdc_in_progress.store(true)`, and the caller in src/conductor.rs propagates it with `?` out to `main`. The process exits, so under systemd this is a permanent crash loop with no hedging, no fill accounting, and no dashboard. The precondition is reachable. The pre-burn timeout sweep in `cleanup_timed_out_usdc_rebalance` abandons a transfer in memory only: it clears inflight and tracking but never sends a command to the store, so the aggregate stays durably in `ConversionComplete` or `BridgingSubmitting` and `interrupted_usdc_rebalance_ids` keeps returning it on every boot. One such stale aggregate plus one new interrupted pre-burn transfer bricks startup. Please keep this fail-closed inside the USDC subsystem instead: log and alert with the ids, leave the gate so `prepare_alpaca_to_base_resume` never returns `Ready` for them, still re-arm and still latch the guard, and return `Ok(())`.

src/rebalancing/trigger/mod.rs
```diff
@@ -1037,6 +1079,11 @@ impl RebalancingService {
                         "USDC transfer timed out; clearing trigger guard and inventory inflight"
```
JuaniRios, line 1186: should fix: the timeout sweep retires a resume gate only for ids that are already in `usdc_tracking`, but `recover_usdc_guard` installs an `AwaitingFreshHedgingSnapshot` gate and the tracking entry is not inserted until `observe_fresh_offchain_usd` succeeds. If the offchain observation keeps failing or never arrives, there is no tracked candidate, so the sweep never sees the gate and never retires it. The delayed job path just keeps pushing another delayed job. USDC rebalancing stays latched, and every other Alpaca-to-Base job is rejected with `ConflictingRecoveryOwner` until an operator intervenes. This contradicts the timeout cleanup the PR describes. Please time the gate out from the recovery `last_progress_at`, or seed enough recovery tracking at gate creation for the sweep to make a stage-aware decision.

src/usdc_rebalance.rs
```diff
@@ -1537,6 +1555,80 @@ impl UsdcRebalance {
             | Self::Reconciled { .. } => None,
         }
     }
+
+    /// Returns the exact source reservation that an Alpaca-to-Base pre-burn
+    /// aggregate must regain after restart, or `None` when the aggregate is not
+    /// in a recoverable source-reservation state.
+    pub(crate) fn alpaca_to_base_reservation_recovery(
```
JuaniRios, line 1565: should fix: of the four recoverable pre-burn stages, three got tests that drive `recover_usdc_guard` and `observe_fresh_offchain_usd` and then assert the resulting tracking stage and restored inflight. `WithdrawalComplete` did not. Its two existing tests are untouched by this diff, and neither calls `prepare_alpaca_to_base_resume` or `observe_fresh_offchain_usd`, nor asserts the restored inflight, the ownership, or the stage. `WithdrawalConfirmed` appears in this diff only inside the production match arm, never in an assertion. A wrong mapping here compiles, keeps every test green, and shows up in production the next time a restart happens in that state. Please add a test that mirrors `recover_usdc_guard_rearms_bridging_submitting_alpaca_to_base` for `WithdrawalComplete{AlpacaToBase}`.

src/inventory/polling.rs
```diff
@@ -541,6 +567,18 @@ where
             )
             .await?;
 
+        if let Some(observer) = &self.fresh_offchain_usd_observer {
```
JuaniRios: should fix: the observer call sits in the middle of `poll_offchain` with a hard `?`, after the `OffchainUsd` send but before the `OffchainCashBuyingPower`, `OffchainCashWithdrawable`, and `AlpacaUsdc` sends. `poll_and_record` only warns on a `poll_offchain` error and continues, so this silently truncates the snapshot group. The error the observer returns, `ConflictingActiveUsdcRebalance`, is not transient: the gate stays awaiting and every later poll tick fails at the same point. `buying_power_cents` feeds equity hedge sizing and `withdrawable_cash_cents` gates the Alpaca-to-Base trigger, so a USDC ownership conflict freezes unrelated equity hedging inputs indefinitely. Please move the observer call after every `snapshot.send(...)` in the function and log its error at error level.

src/rebalancing/trigger/mod.rs
```diff
@@ -2029,6 +2082,107 @@ impl RebalancingService {
...
+            Some(gate) => Err(
+                AlpacaToBaseResumePreparationError::ConflictingRecoveryOwner {
+                    requested: id.clone(),
+                    owner: gate.id().clone(),
+                },
+            ),
+            None => Ok(AlpacaToBaseResumePreparation::Ready),
+        }
+    }
+}
```
JuaniRios, line 2280: minor: this `None` arm gates every Alpaca-to-Base transfer that was not interrupted by a restart, which is nearly all production traffic, and no test exercises it on the real service. The job tests use the `ReadyAlpacaToBaseResume` stub, so the stub's `Ok(Ready)` runs instead. The tests that build a real `RebalancingService` all seed a recovery first, or use a mismatched id to hit the conflict arm. Please add a small unit test that builds a real service with no recovery seeded and asserts `Ready` for a fresh id.

src/rebalancing/trigger/mod.rs
```diff
@@ -2029,6 +2082,107 @@ impl RebalancingService {
...
+
+#[async_trait]
+impl FreshOffchainUsdObserver for RebalancingService {
+    async fn observe_fresh_offchain_usd(
+        &self,
+        available: Usdc,
+        observed_at: DateTime<Utc>,
+    ) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+        let event_sync_guard = self.usdc_event_sync.lock().await;
+        let awaiting = match self.alpaca_to_base_resume_gate.read().await.as_ref() {
```
JuaniRios: minor: this restores inventory only while the gate is `AwaitingFreshHedgingSnapshot`. Once it flips to `ReservationRestored`, later fresh broker observations return early, and `Inventory::on_snapshot` cannot compensate because it skips all snapshots while any venue has inflight. If broker cash changes after the first restore but before the transfer settles, Hedging `available` stays pinned to the first observed value. Terminal settlement then clears the guard and enqueues a new imbalance check against stale source cash. Please treat `ReservationRestored` for the same id as an idempotent replacement: apply `restore_reservation` again with the latest available balance, without adding inflight or changing ownership. Add a test where the second observation carries a different balance.

src/rebalancing/trigger/mod.rs
```diff
@@ -538,6 +558,27 @@ pub(crate) struct RebalancingService {
     usdc_store: RwLock<Option<Arc<Store<UsdcRebalance>>>>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
```
JuaniRios: minor: `Clone`, `PartialEq`, and `Eq` are derived here but nothing clones the whole value or compares two gates with `==`. Call sites match out and clone individual fields, or call `.id()` and compare the inner id, or check `.is_none()`. Please keep `Debug` only.

src/rebalancing/usdc/job.rs
```diff
@@ -153,6 +157,50 @@ pub(crate) trait ResumeAlpacaToBase: Send + Sync + 'static {
...
+/// Restores restart-only source inventory ownership before the transfer
+/// manager resumes. Fresh transfers return [`AlpacaToBaseResumePreparation::Ready`].
+#[async_trait]
+pub(crate) trait PrepareAlpacaToBaseResume: Send + Sync + 'static {
+    async fn prepare_alpaca_to_base_resume(
+        &self,
+        id: &UsdcRebalanceId,
+    ) -> Result<AlpacaToBaseResumePreparation, AlpacaToBaseResumePreparationError>;
+}
+
```
JuaniRios: minor: `ReadyAlpacaToBaseResume` and its trait impl are `#[cfg(test)]` items at module scope, between the trait definition and the production impl, but they are used only by `market_making_ctx()` inside the `mod tests` block in this same file. docs/feature-flags.md prescribes `#[cfg(test)] mod tests` for a helper used in one file, and reserves module-scope test doubles for helpers shared across files. Please move the struct and its impl into the `mod tests` block, next to `AwaitingAlpacaToBaseResume`.

JuaniRios (commented):
Five high findings here, two of which can move or strand money: the resume gate rejects every non-owner Alpaca-to-Base job, including post-burn jobs re-armed in the same startup pass, and ambiguous recovery propagates out to main and crash-loops the process instead of failing closed on USDC alone. The BridgingSubmitting mapping also fabricates a post-burn tracking stage before any burn exists.

JuaniRios (changes requested):
Requesting changes. Two of the five high findings can move or strand money:

- `prepare_alpaca_to_base_resume` returns `ConflictingRecoveryOwner` for every id that is not the gate owner, including post-burn jobs that `recover_usdc_guard` re-arms in the same startup pass. A transfer whose USDC is already burned in CCTP exhausts its retries and lands in a terminal row that is never re-armed.
- Ambiguous reservation recovery returns `Err` before the guard latches, and the caller propagates it out to `main`. Under systemd that is a crash loop with no hedging, and the precondition is reachable through the in-memory timeout sweep.

Also blocking: the `BridgingSubmitting` mapping fabricates a post-burn tracking stage before any burn exists, the awaiting gate never expires when the fresh snapshot never arrives, and `WithdrawalComplete{AlpacaToBase}` recovery has no test at all.

src/rebalancing/trigger/mod.rs
0xgleb: Addressed: resume preparation now returns Ready for every ID not represented by the recovery gate, so unrelated post-burn jobs keep their retry budget. The fallible preparation error was removed entirely, and a real-service regression test covers the non-owner path.

src/rebalancing/trigger/mod.rs
0xgleb: Addressed: BridgingSubmitting without a pending burn now reconstructs WithdrawalConfirmed, the last accurate pre-burn stage. The focused recovery test asserts that stage explicitly.

src/rebalancing/trigger/mod.rs
0xgleb: Addressed: multiple candidates now install a fail-closed ConflictingReservations gate, log and alert with all IDs, continue rearming jobs, latch the USDC guard, and return Ok. A regression test verifies startup remains live, both owners stay gated, both jobs are rearmed, and the alert names both IDs.

src/rebalancing/trigger/mod.rs
0xgleb: Addressed: installing a single recovery gate now seeds usdc_tracking immediately from the durable recovery stage and last_progress_at, before any fresh broker observation. This makes the existing stage-aware timeout sweep able to select the recovery while the freshness gate is still awaiting.

src/usdc_rebalance.rs
0xgleb: Addressed: the WithdrawalComplete Alpaca-to-Base startup test now drives the freshness observation and asserts Awaiting then Ready, WithdrawalConfirmed tracking, exact Hedging inflight, refreshed available balance, and active ownership.

src/inventory/polling.rs
0xgleb: Addressed: the observer now runs only after all offchain snapshot sends, and its failure is logged at error level without failing poll_offchain. A regression test proves buying power, withdrawable cash, and Alpaca USDC are all persisted when the observer fails.

src/rebalancing/trigger/mod.rs
0xgleb: Addressed: added a real RebalancingService unit test with no recovery gate that asserts a fresh Alpaca-to-Base ID is Ready.

src/rebalancing/trigger/mod.rs
0xgleb: Addressed: ReservationRestored observations now idempotently replace the Hedging snapshot with each latest available balance while retaining the exact inflight amount and owner. The new two-observation regression test verifies the changed balance.

src/rebalancing/trigger/mod.rs
0xgleb: Addressed: AlpacaToBaseResumeGate now derives Debug only.

0xgleb: Addressed: AlpacaToBaseResumeGate now derives Debug only.

src/rebalancing/usdc/job.rs
0xgleb: Addressed: ReadyAlpacaToBaseResume and its impl now live inside mod tests beside AwaitingAlpacaToBaseResume.

src/rebalancing/trigger/mod.rs
```diff
@@ -137,6 +138,19 @@ pub(crate) enum RebalancingServiceError {
...
     #[error("failed to re-arm a stranded USDC transfer job at startup: {0}")]
     RearmEnqueue(#[from] QueuePushError),
+    #[error("cannot restore Alpaca-to-Base rebalance {recovered}: inventory is owned by {active}")]
+    ConflictingActiveUsdcRebalance {
+        recovered: UsdcRebalanceId,
+        active: UsdcRebalanceId,
+    },
+    #[error(
+        "recovered Alpaca-to-Base reservation for {id} is {reserved}, but Initiated carries {initiated}"
+    )]
```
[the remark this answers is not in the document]
0xgleb: Fixed in 47973d89: wrapped the RecoveredReservationAmountMismatch attribute string to the repository 100-column limit.

src/rebalancing/trigger/mod.rs
```diff
@@ -2029,6 +2120,84 @@ impl RebalancingService {
...
+
+        if let Some(recovery) = initial_recovery {
+            self.usdc_tracking
+                .write()
+                .await
+                .insert(id.clone(), tracking_from_reservation_recovery(recovery));
+            *self.alpaca_to_base_resume_gate.write().await =
+                Some(AlpacaToBaseResumeGate::ReservationRestored { id, amount });
+        }
+        drop(event_sync_guard);
```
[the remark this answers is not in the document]
0xgleb: Fixed in 47973d89: timeout cleanup now retires the recovery gate and in-progress guard inside the usdc_event_sync critical section. Added a deterministic interleaving regression proving a stale observation cannot restore the timed-out reservation.

src/rebalancing/usdc/job.rs
```diff
@@ -70,6 +70,10 @@ const SETTLEMENT_REDRIVE_DELAY: Duration = Duration::from_secs(30);
 /// (Ethereum block timing) so the two can evolve independently.
 const WITHDRAWAL_POLL_REDRIVE_DELAY: Duration = Duration::from_secs(30);
 
+/// Delay before retrying a recovered Alpaca-to-Base job whose source
+/// reservation is still waiting for a fresh hedging inventory observation.
+const INVENTORY_RECOVERY_REDRIVE_DELAY: Duration = Duration::from_secs(30);
```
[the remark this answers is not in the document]
0xgleb: Fixed in 47973d89: inventory_recovery_redrive_delay_secs is now a required, non-zero rebalancing setting, validated at startup and threaded through RebalancingCtx into TransferUsdcToMarketMakingCtx. Public deployment/example configs and missing/zero-value tests were updated.

src/rebalancing/usdc/job.rs
```diff
@@ -1506,12 +1548,78 @@ mod tests {
...
+        };
+
+        let before = Utc::now().timestamp();
+        job.perform(&ctx).await.unwrap();
+        let after = Utc::now().timestamp();
+
+        assert!(
+            transfer.captured.lock().unwrap().is_none(),
+            "the transfer manager must not run before recovery restores source inventory"
+        );
```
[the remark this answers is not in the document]
0xgleb: Fixed in 47973d89: the blocked-resume regression now captures the value and asserts it equals None.

src/usdc_rebalance.rs
```diff
@@ -1537,6 +1555,80 @@ impl UsdcRebalance {
...
+                direction: RebalanceDirection::AlpacaToBase,
+                amount,
+                initiated_at,
+                pending_burn_tx: None,
+                ..
+            } => (
+                *amount,
+                AlpacaToBaseReservationStage::BridgingSubmitting,
+                *initiated_at,
+            ),
```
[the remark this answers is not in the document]
0xgleb: Fixed in 47973d89: pre-burn recovery keeps the nominal reservation amount, Withdrawing recovery now requires TransferRef::AlpacaId, and both cases have regressions.

PR #1051, opened 2026-07-16
fix: resolve RustSec audit findings

0xgleb:
## What

Make RustSec auditing an enforced backend CI check and resolve the currently actionable dependency findings.

## Why

Dependency advisories were not enforced by CI, so a vulnerable or abandoned direct dependency could enter the workspace without blocking review. The workspace also retained an unused direct `rsa` dependency that unnecessarily exposed its advisory surface.

## How

- Remove the unused direct `rsa` workspace and test dependencies.
- Add `cargo-audit` to the Nix backend development/CI environments.
- Run `cargo audit` in the backend CI job.
- Document and narrowly ignore two transitive advisories whose affected code is not built by this SQLite-only application: the MySQL path in `sqlx-apalis` and the unused `ark-relations` tracing path.
- Refresh `Cargo.lock` after removing the direct dependency.

## Testing

- `cargo audit` passes with only the documented transitive ignores.
- `cargo metadata` succeeds.
- GitHub Actions static checks and backend build/test matrix pass.

## Screenshots

Not applicable.

## Anything else

The advisory ignores are scoped in `.cargo/audit.toml` with the reason each affected dependency path is unreachable. CI will continue to fail on every other RustSec advisory.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,6 @@
+[advisories]
+# `sqlx-apalis` includes the MySQL driver in Cargo.lock, but this SQLite-only
+# application never enables or builds that driver.
+# `ark-relations` includes its tracing subscriber in Cargo.lock, but no selected
+# dependency path builds `ark-relations`.
+ignore = ["RUSTSEC-2023-0071", "RUSTSEC-2025-0055"]
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -143,6 +143,9 @@ jobs:
               -E "test(compile_fail_tests)"
           '
 
+      - name: Audit Rust dependencies
+        run: nix develop .#ci-backend -c cargo audit
+
       - name: Upload nextest archive
         uses: actions/upload-artifact@v4
         with:
--- a/flake.nix
+++ b/flake.nix
@@ -540,6 +540,7 @@
             # and they bloat the closure.
             backendInputs = [
               rustToolchain
+              pkgs.cargo-audit
               pkgs.openssl
               pkgs.sqlite
               pkgs.pkg-config
@@ -570,6 +571,7 @@
                     bun
                     sccache
                     sqlx-cli
+                    cargo-audit
                     cargo-nextest
                     ragenixPkg
                     packages.secret
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2137,9 +2137,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+checksum = "2d6914041f254d6e9176c01941b21115dcfb7089e55135a35411081bd106ef3f"
 dependencies = [
  "crossbeam-utils",
 ]
@@ -4795,9 +4795,9 @@ dependencies = [
 
 [[package]]
 name = "quinn-proto"
-version = "0.11.14"
+version = "0.11.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "434b42fec591c96ef50e21e886936e66d3cc3f737104fdb9b737c40ffb94c098"
+checksum = "4fcb935c5bec503c2f0e306bdd3e58bb9029dcb14fa8d9ac76e3a5256ac0763e"
 dependencies = [
  "bytes",
  "getrandom 0.3.4",
@@ -6709,7 +6709,6 @@ dependencies = [
  "rand 0.8.6",
  "regex",
  "reqwest 0.12.28",
- "rsa",
  "serde",
  "serde_json",
  "serial_test",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,7 +87,6 @@ reqwest = { version = "0.12.4", features = [
   "gzip",
   "blocking",
 ] }
-rsa = { version = "0.9.10", features = ["pem"] }
 serde = { version = "1.0.197", features = ["derive"] }
 serde_json = "1.0.140"
 serial_test = "3.2.0"
@@ -257,7 +256,6 @@ httpmock.workspace = true
 proptest.workspace = true
 rain-math-float.workspace = true
 regex = "1.12.3"
-rsa.workspace = true
 serial_test.workspace = true
 st0x-bridge = { workspace = true, features = ["mock"] }
 st0x-event-sorcery = { workspace = true, features = ["test-support"] }


0xgleb:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1051?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1051** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1051?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1051?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1051).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

.github/workflows/ci.yaml
```diff
@@ -143,6 +143,9 @@ jobs:
               -E "test(compile_fail_tests)"
           '
 
+      - name: Audit Rust dependencies
+        run: nix develop .#ci-backend -c cargo audit
```
JuaniRios: should fix: bare `cargo audit` fails the job only on vulnerability advisories. Informational advisories (`unmaintained`, `unsound`, `notice`) and yanked crates go into the warnings list, and that list fails the run only when a kind is in the deny set. The deny set comes from `[output] deny` in the config, which defaults to empty, and no `--deny` flag is passed here. So an abandoned or unsound direct dependency can enter the workspace and CI stays green with a warning buried in the backend-build log. The PR's stated goal is to block a "vulnerable or abandoned" dependency, so half of that is not delivered. Please run `cargo audit --deny warnings`, or add `[output] deny = ["unmaintained", "unsound", "yanked"]` to `.cargo/audit.toml`. Expect this to surface existing warnings; triage each into a documented `ignore` entry. If blocking on unmaintained is deliberately out of scope, please correct the PR description instead.

.cargo/audit.toml
```diff
@@ -0,0 +1,6 @@
+[advisories]
+# `sqlx-apalis` includes the MySQL driver in Cargo.lock, but this SQLite-only
+# application never enables or builds that driver.
+# `ark-relations` includes its tracing subscriber in Cargo.lock, but no selected
+# dependency path builds `ark-relations`.
+ignore = ["RUSTSEC-2023-0071", "RUSTSEC-2025-0055"]
```
JuaniRios: should fix: `cargo audit` ignores by advisory id, workspace-wide. There is no way to scope an ignore to a dependency path, so these two entries are not narrow. If a later PR reintroduces a direct `rsa` dependency, or makes the `ark-relations` path reachable, CI passes for the same advisory. That mutes the exact advisory this PR just cleaned up, so the cleanup can silently regress. Please add a guard that keeps the exception honest, for example a small check that runs `cargo audit --json` and fails if an ignored advisory appears outside the expected package path, or a dependency-graph check that proves `rsa` is not a direct dependency.

.github/workflows/ci.yaml
```diff
@@ -143,6 +143,9 @@ jobs:
               -E "test(compile_fail_tests)"
           '
 
+      - name: Audit Rust dependencies
```
JuaniRios: minor: this step sits between the workspace build and the nextest archive upload. Steps default to `if: success()`, so an audit failure skips the upload, fails `backend-build`, skips all four `backend-test` partitions, and fails the required `backend` gate. Unlike clippy and nextest, `cargo audit` is not hermetic: it refreshes the advisory database from github.com at run time, so a newly published upstream advisory or a transient fetch failure discards a full workspace compile and produces zero test results for an unrelated push. Please move the audit into its own job, or into the cheap `hooks` job, so it reports independently. If it must stay here, put it after the archive upload.

flake.nix
```diff
@@ -570,6 +571,7 @@
                     bun
                     sccache
                     sqlx-cli
+                    cargo-audit
```
JuaniRios, line 662: minor: `cargo-audit` is now in the CI shell, but the `nix run .#ci` wrapper still runs only check, nextest, clippy, and fmt for the backend. The project docs describe that wrapper as the local mirror of CI, so an engineer can run it, get a clean result, and still fail the new required gate on GitHub. Please add the audit to the backend block of the `ci` package, with the same command and config as the workflow.

.cargo/audit.toml
```diff
@@ -0,0 +1,6 @@
+[advisories]
+# `sqlx-apalis` includes the MySQL driver in Cargo.lock, but this SQLite-only
```
JuaniRios: nit: these justifications name a crate that does not exist. `Cargo.lock` has `apalis-sql` (no sqlx dependency) and `apalis-sqlite` (which depends on `sqlx`), but no `sqlx-apalis`. RUSTSEC-2023-0071 is against `rsa`, reached through `sqlx`'s optional `sqlx-mysql` feature, and neither crate is named. RUSTSEC-2025-0055 is against `tracing-subscriber` 0.2.25, referred to only as "its tracing subscriber". The next engineer who has to revalidate an ignore will grep for `sqlx-apalis`, find nothing, and leave a stale entry in place. Please name the affected crate, its version, and the exact path in each comment.

JuaniRios (commented):
Making the audit a blocking gate is the right call, but as configured it does not deliver the stated goal. Bare cargo audit fails only on vulnerability advisories; an abandoned or unsound dependency stays a warning because the deny list is empty. The ignores are also workspace-wide by advisory id, so the rsa advisory this PR just cleaned up is now permanently muted and can silently regress.

JuaniRios (changes requested):
Requesting changes. The gate does not deliver its stated goal as configured:

- Bare `cargo audit` fails only on vulnerability advisories. Unmaintained, unsound, and yanked findings go to the warnings list, which fails the run only when a kind is in the deny set, and that set is empty. So an abandoned direct dependency still enters the workspace with CI green.
- The ignores are workspace-wide by advisory id, with no way to scope them to a path. The `rsa` advisory this PR just cleaned up is now permanently muted and can silently regress.

Please also move the audit out of `backend-build`, or below the archive upload, so a non-hermetic advisory-database fetch cannot discard a full workspace compile.

.github/workflows/ci.yaml
0xgleb: Addressed. The audit config now denies every unmaintained and unsound advisory, while vulnerabilities remain fatal by default. The pinned cargo-audit cannot ignore a single yanked crate, so scripts/audit-rust.sh parses the JSON report and permits only spin 0.9.8; any additional yanked dependency fails the check. Every current advisory is either upgraded or documented with an exact guarded path.

.cargo/audit.toml
0xgleb: Addressed. scripts/audit-rust.sh now validates every ignored package with cargo tree across all targets and normal/development edges. It asserts exact versions and immediate dependents, and proves the lock-only rsa, tracing-subscriber, and derivative packages have no selected path. Any new path or version fails CI and forces the exception to be reviewed.

.github/workflows/ci.yaml
0xgleb: Addressed. The audit was removed from backend-build and moved into the existing lightweight hooks job, before the other hook validations. Backend compilation, archive upload, and all test partitions now proceed independently of the live advisory fetch.

flake.nix
0xgleb: Addressed. nix run .#ci now invokes the same nix develop .#ci-audit -c ./scripts/audit-rust.sh command used by GitHub Actions, so local and hosted audit behavior share one implementation and config.

.cargo/audit.toml
0xgleb: Addressed. Every exception comment now names the RustSec ID, affected crate and version, and exact dependency path. The rsa entry specifically identifies rsa 0.9.10 as lock-only through sqlx-mysql 0.8.6; the tracing entry identifies tracing-subscriber 0.2.25 through ark-relations 0.5.1.

scripts/audit-rust.sh
```diff
@@ -0,0 +1,86 @@
...
+for required_policy in \
+  '[database]' \
+  'url = "https://github.com/RustSec/advisory-db"' \
+  'fetch = true' \
+  'stale = false' \
+  '[yanked]' \
+  'enabled = true' \
+  'update_index = true'; do
+  assert_config_line "$required_policy"
+done
```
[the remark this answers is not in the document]
0xgleb: Fixed in cebbc862: the audit script now requires the [advisories] and [output] sections, asserts the exact deny categories, extracts and compares the complete normalized RUSTSEC ignore set, and fails with the expected/actual policy when either set drifts. The exact ci-audit command and shellcheck hook pass.

PR #1053, opened 2026-07-17
fix: surface failed counter-trades

JuaniRios:
## What

- Fixes [RAI-1418](https://linear.app/makeitrain/issue/RAI-1418/dashboard-does-not-surface-failed-counter-trades).
- Surfaces failed counter-trades in both initial dashboard history and live updates.
- Shows the terminal error and the filled, unfilled, and excess portions of failed broker orders.

## Why

Failed hedges were only visible through logs or direct database queries. That made it too easy to miss unhedged exposure from the dashboard's main operational view.

## How

- Adds a typed terminal outcome to the trade DTO and uses the terminal outcome timestamp for ordering.
- Converts failed offchain orders into trade entries without hiding partial fills or clamping broker overfill anomalies.
- Makes the history loaders and `/trades` endpoint fail closed on malformed persisted data or invalid filters.
- Broadcasts terminal offchain outcomes and reconciles live updates with paginated history without duplicate rows, stale responses, or count rollbacks.
- Keeps initial and live history ordering aligned, including numeric Raindex log-index tie breaks.
- Renders failed outcomes and their quantities in the trade-history table and detail view.

## Testing

- `cargo check --workspace` and `cargo check --workspace --all-features`
- `cargo nextest run --workspace --all-features`: 3,621 passed, 6 skipped
- `cargo clippy --workspace --all-targets --all-features`
- `cargo fmt -- --check`, Nix formatting, and changed-file pre-commit hooks
- Dashboard lint, `svelte-check`, and Vitest: 258 passed
- `nix build .#st0x-dashboard`
- 7 full review-loop iterations, plus focused review-feedback verification and clean re-reviews

## Screenshots

Not included. The dashboard change is covered by mounted Svelte component tests.

## Anything else

The trade WebSocket statement changes from `trade_fill` to `trade_update` because it now carries both fills and failures. Rolling-version compatibility is tracked separately from this fix.

## Summary by CodeRabbit

* **New Features**
  * Trade History now shows both completed and failed trades, including outcome status plus failure reasons and filled, unfilled, and excess quantities.
  * The history table and detail view now include an explicit trade status (“Filled” / “Failed”) and use the trade occurrence time.
  * Live updates continue to reconcile correctly with the current filtered/paginated history.

* **Bug Fixes**
  * Invalid trade-history filters now produce clear client-facing errors.
  * Terminal outcomes now reliably appear in both the initial dashboard state and subsequent live updates without stale or duplicated rows.

* **Documentation**
  * Updated Trade History panel documentation to clarify on-chain vs counter-trade failure representation and quantity accounting.

### DIFF
--- a/crates/dto/src/trade.rs
+++ b/crates/dto/src/trade.rs
@@ -4,7 +4,7 @@ use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use ts_rs::TS;
 
-use st0x_finance::{FractionalShares, Symbol};
+use st0x_finance::{FractionalShares, NonNegative, Symbol};
 
 /// Where a trade was executed.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, TS)]
@@ -118,20 +118,78 @@ pub struct InvalidDirectionError {
     direction_provided: String,
 }
 
-/// A completed trade fill.
+/// Terminal outcome of a dashboard trade entry.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, TS)]
+#[serde(
+    tag = "status",
+    rename_all = "snake_case",
+    rename_all_fields = "camelCase"
+)]
+pub enum TradeOutcome {
+    Filled,
+    Failed {
+        error: String,
+        #[ts(type = "string")]
+        filled_shares: NonNegative<FractionalShares>,
+        #[ts(type = "string")]
+        remaining_shares: NonNegative<FractionalShares>,
+        /// Shares filled beyond the broker-accepted order quantity. This is
+        /// separate from remaining shares so anomalous broker state is never
+        /// clamped away.
+        #[ts(type = "string")]
+        excess_shares: NonNegative<FractionalShares>,
+    },
+}
+
+/// A completed onchain fill or terminal offchain counter-trade.
 #[derive(Debug, Clone, Serialize, TS)]
 #[serde(rename_all = "camelCase")]
 pub struct Trade {
     /// Unique identifier for deduplication on reconnect.
     /// Onchain: `"tx_hash:log_index"`. Offchain: offchain order aggregate ID.
     pub id: String,
-    pub filled_at: DateTime<Utc>,
+    pub occurred_at: DateTime<Utc>,
     pub venue: TradingVenue,
     pub direction: Direction,
     #[ts(type = "string")]
     pub symbol: Symbol,
+    /// Quantity submitted for this counter-trade. Before broker acceptance this
+    /// is the requested quantity; afterward it is the quantity the broker
+    /// accepted. Failed outcomes split this order quantity into filled and
+    /// remaining portions in [`TradeOutcome::Failed`].
     #[ts(type = "string")]
     pub shares: FractionalShares,
+    pub outcome: TradeOutcome,
+}
+
+/// Sorts dashboard trades newest-first with a stable cross-loader tie-breaker.
+pub fn sort_trades_newest_first(trades: &mut [Trade]) {
+    trades.sort_by(|left, right| {
+        right
+            .occurred_at
+            .cmp(&left.occurred_at)
+            .then_with(|| compare_trade_ids(left, right))
+    });
+}
+
+fn compare_trade_ids(left: &Trade, right: &Trade) -> std::cmp::Ordering {
+    if left.venue == TradingVenue::Raindex
+        && right.venue == TradingVenue::Raindex
+        && let (Some((left_hash, left_index)), Some((right_hash, right_index))) = (
+            parse_onchain_trade_id(&left.id),
+            parse_onchain_trade_id(&right.id),
+        )
+        && left_hash == right_hash
+    {
+        return right_index.cmp(&left_index);
+    }
+
+    left.id.cmp(&right.id)
+}
+
+fn parse_onchain_trade_id(id: &str) -> Option<(&str, u64)> {
+    let (tx_hash, log_index) = id.rsplit_once(':')?;
+    Some((tx_hash, log_index.parse().ok()?))
 }
 
 #[cfg(test)]
@@ -158,14 +216,15 @@ mod tests {
     }
 
     #[test]
-    fn trade_serializes_all_fields() {
+    fn filled_trade_serializes_all_fields() {
         let trade = Trade {
             id: "test-order-id".to_string(),
-            filled_at: DateTime::from_timestamp(1_700_000_000, 0).unwrap(),
+            occurred_at: DateTime::from_timestamp(1_700_000_000, 0).unwrap(),
             venue: TradingVenue::Alpaca,
             direction: Direction::Sell,
             symbol: Symbol::new("TSLA").unwrap(),
             shares: FractionalShares::new(float!(5.5)),
+            outcome: TradeOutcome::Filled,
         };
         let json = serde_json::to_value(&trade).expect("serialization should succeed");
         assert_eq!(json["id"], json!("test-order-id"));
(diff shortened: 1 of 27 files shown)


JuaniRios:
* **#1062** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1062?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1058** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1058?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1053** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1053?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1053?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1053).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

0xgleb (commented):
agent misfire

src/api.rs
```diff
@@ -646,131 +615,232 @@ impl TradeFilter {
...
+    .await?;
 
     rows.into_iter()
-        .filter_map(|(view_id, payload_str)| {
-            let payload: serde_json::Value = serde_json::from_str(&payload_str).ok()?;
-            let filled = payload.get("Live")?.get("Filled")?;
+        .map(|(view_id, payload)| parse_offchain_trade_row(view_id, &payload, filter))
+        .filter_map(Result::transpose)
+        .collect()
+}
```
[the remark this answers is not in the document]
JuaniRios: Agreed on the growth, and it is real: both loaders do `fetch_all` with no LIMIT and page in memory. Deferring rather than doing it here, because it needs SQL-side filters plus an index migration, which is out of scope for this PR. Tracked in RAI-1526.

findolor (approved):

src/api.rs
```diff
@@ -646,131 +615,224 @@ impl TradeFilter {
...
-                id: aggregate_id,
-                filled_at: filled_at.to_rfc3339(),
-                venue: "raindex".to_string(),
-                direction: direction.to_string(),
-                symbol: trade_symbol.to_string(),
-                shares: amount.to_string(),
-            })
-        })
-        .collect()
+    Ok(Some(result))
```
rouzwelt, line 764: One undecodable row takes the whole Trade History panel offline.

parse_*_trade_row promotes any per-row decode failure to TradeHistoryError, and the handler maps that to 500 (asserted by offchain_trade_history_rejects_malformed_failed_orders). A single legacy/corrupt payload — exactly the shape this PR's new Failed fields introduce for pre-upgrade rows — therefore hides all trades indefinitely, which contradicts SPEC's "rather than … making history unavailable" intent for anomalous broker state.

Prefer fail-open per row with loud observability, keeping infrastructure errors (sqlx::Error) fatal:

```
-    rows.into_iter()
-        .map(|(view_id, payload)| parse_offchain_trade_row(view_id, &payload, filter))
-        .filter_map(Result::transpose)
-        .collect()
+    Ok(rows
+        .into_iter()
+        .filter_map(|(view_id, payload)| {
+            parse_offchain_trade_row(view_id, &payload, filter)
+                .inspect_err(|error| {
+                    warn!(target: "dashboard", ?error, "Skipping unparseable trade-history row");
+                })
+                .ok()
+                .flatten()
+        })
+        .collect())
```

If failing closed is the deliberate choice, surface it as a partial-data warning in the response (the DTO already has a warnings channel) rather than an opaque 500, and document it in SPEC.

Also applies to: 816-836

dashboard/src/lib/components/trade-history-panel.svelte
```diff
@@ -142,19 +185,22 @@
...
-        entries.update((prev) => [...prev, ...data.entries])
-      } else {
-        entries.update(() => data.entries)
-      }
+      const nextEntries = isLoadMore
+        ? mergeLiveTrades([...entries.current, ...data.entries])
+        : mergeLiveTrades(data.entries)
+      const reconciledTotal = Math.max(total.current, data.total)
+      total.update(() => reconciledTotal)
+      hasMore.update(() => data.hasMore || nextEntries.length < reconciledTotal)
```
rouzwelt, line 206: total is monotonically increasing, so a full refresh can never correct live-merge drift — and hasMore inherits the drift.

reconciledTotal = Math.max(total.current, data.total) means the server total is only ever used to raise the count, never to lower it. Every live merge that adds an entry bumps total by one (line 151), so once a live trade is also returned by a later server page the dedupe drops the row but the inflated count stays. Because hasMore is data.hasMore || nextEntries.length < reconciledTotal, the "Load older trades" button then stays enabled forever while each click appends nothing.

A 'replace' fetch is an authoritative snapshot: it should adopt data.total and re-add cached live trades on top, rather than clamping to the previously accumulated value. Only the append path needs the max.

```
-      const nextEntries = isLoadMore
-        ? mergeLiveTrades([...entries.current, ...data.entries])
-        : mergeLiveTrades(data.entries)
-      const reconciledTotal = Math.max(total.current, data.total)
+      const merged = isLoadMore ? [...entries.current, ...data.entries] : data.entries
+      const nextEntries = mergeLiveTrades(merged)
+      // A replace is an authoritative snapshot: trust the server total and add
+      // back only the live trades the snapshot has not caught up with yet.
+      const liveExtras = Math.max(0, nextEntries.length - merged.length)
+      const reconciledTotal = isLoadMore
+        ? Math.max(total.current, data.total)
+        : data.total + liveExtras
       total.update(() => reconciledTotal)
       hasMore.update(() => data.hasMore || nextEntries.length < reconciledTotal)
       entries.update(() => nextEntries)
```

crates/dto/src/trade.rs
```diff
@@ -118,20 +118,78 @@ pub struct InvalidDirectionError {
...
+    {
+        return left_index.cmp(&right_index);
+    }
+
+    left.id.cmp(&right.id)
+}
+
+fn parse_onchain_trade_id(id: &str) -> Option<(&str, u64)> {
+    let (tx_hash, log_index) = id.rsplit_once(':')?;
+    Some((tx_hash, log_index.parse().ok()?))
```
rouzwelt, line 192: Tie-breaker orders log indices oldest-first inside a "newest-first" sort.

For two fills in the same transaction at the same occurred_at, left_index.cmp(&right_index) puts the lower log index first. Within a transaction a higher log index is strictly later, so this is the reverse of the documented newest-first ordering (the tests at Line 269 and in src/dashboard/trade_loader.rs lock in the inverted direction).

```
-        return left_index.cmp(&right_index);
+        return right_index.cmp(&left_index);
```

src/dashboard/mod.rs
```diff
@@ -86,7 +86,13 @@ async fn drain_client_messages(stream: &mut futures_util::stream::SplitStream<We
 async fn send_initial_state(sink: &mut SplitSink<WebSocket, Message>, state: &AppState) -> bool {
     let inventory_dto = state.inventory.read().await.to_dto();
     let transfers = transfer_loader::load_transfers(&state.pool).await;
-    let trades = trade_loader::load_trades(&state.pool).await;
+    let trades = match trade_loader::load_trades(&state.pool).await {
+        Ok(trades) => trades,
+        Err(error) => {
+            warn!(target: "dashboard", %error, "Failed to load initial trade history");
+            return false;
+        }
```
rouzwelt, line 100: Failing trade load now silently closes the socket, unlike every other initial-state loader.

transfer_loader::load_transfers and load_positions degrade gracefully, but a trade-history error aborts handle_socket with no close reason. Since the client auto-reconnects, a persistent DB error turns into a tight reconnect loop with nothing visible to the operator. Prefer either sending the initial state with an empty/partial trade list plus a warnings entry, or closing with an explicit reason so the client can surface it and back off.

rouzwelt (changes requested):

src/api.rs
JuaniRios: Fixed in ec095eef. Both loaders now skip an unparseable row with a warn log and keep sqlx errors fatal, as suggested. I applied the same pattern to the websocket-path loaders in trade_loader.rs, since they had the same fail-closed shape. The test that asserted the 500 now asserts skip-and-continue.

dashboard/src/lib/components/trade-history-panel.svelte
JuaniRios: Fixed in ec095eef with your suggested split: replace adopts the server total plus live extras, append keeps the max. Added a panel test where an inflated total previously kept "Load older trades" enabled forever.

crates/dto/src/trade.rs
JuaniRios: Fixed in ec095eef. Flipped to `right_index.cmp(&left_index)` and also flipped the mirrored comparator in `dashboard/src/lib/trade.ts`; updated the three tests that locked in the old order.

src/dashboard/mod.rs
JuaniRios: Fixed in ec095eef. A trade-load failure now sends the initial state with an empty trade list plus a `TradeHistoryUnavailable` warnings entry instead of closing the socket. With the per-row skip from your other comment, this path is now only reachable on real infrastructure errors.

rouzwelt (approved):

PR #675, opened 2026-05-15
fix: correct equity redemption inventory races

JuaniRios:
## What

Addresses [RAI-583](https://linear.app/makeitrain/issue/RAI-583/use-actual-withdraw4-received-amount-before-unwrapping-equity) and [RAI-584](https://linear.app/makeitrain/issue/RAI-584/prevent-stale-onchain-fill-deltas-from-inflating-inventory-after).

Fixes equity redemption and inventory accounting races that could leave the rebalancing view tracking the wrong quantity.

This PR records actual Raindex withdrawal amounts from transaction receipts, carries actual unwrapped quantities through redemption state, and prevents stale inventory snapshots from overwriting newer absolute balances while still allowing fill deltas to apply.

## Why

Raindex withdrawals can return less than the requested wrapped-token amount. The redemption flow previously kept using the requested quantity after confirmation, which could overstate in-flight equity and make stuck-redemption recovery re-trigger work for inventory the bot no longer holds.

Inventory snapshots also used timestamps like causal watermarks for all updates. That let stale snapshots overwrite newer snapshot values, and could suppress legitimate fill deltas whose exchange timestamp was earlier than a later snapshot fetch.

## How

- `Raindex::confirm_tx_receipt` now returns the confirmed transaction receipt, while `confirm_tx` remains a convenience wrapper.
- `EquityRedemptionEvent::WithdrawnFromRaindex` stores `actual_wrapped_amount`, decoded from matching ERC-20 `Transfer` logs in the withdrawal receipt.
- `EquityRedemptionEvent::TokensUnwrapped` stores the actual fractional `quantity` derived from the unwrapped amount.
- `EquityRedemption` uses the actual withdrawn/unwrapped amount for later redemption states, while keeping event fields backward-compatible with `serde(default)`.
- `symbols_with_stuck_redemptions` recovers quantities from the latest sent/unwrapped/withdrawn event instead of blindly using the originally requested quantity, and now errors on invalid recovered amounts instead of silently skipping bad financial data.
- `InventorySnapshot` persists per-snapshot fetched timestamps for onchain equity, onchain USDC, offchain equity, and offchain USD, and ignores older snapshot events for the same snapshot kind.
- `InventoryView` tracks per-symbol equity snapshot watermarks, applies absolute equity snapshots only when they are newer and safe to apply, and continues applying position fill deltas independently of snapshot timestamps.
[...]

PR #1057, opened 2026-07-20
fix: skip offchain equity snapshots while a hedge order is open

agryaznov:
## What

Proper reconciliation between the inventory poller and the offchain order
reactor.

Two uncoordinated writers set `InventoryView`'s Hedging (offchain) equity
balance: the ~60s poller snapshot (absolute **set**) and the offchain fill
delta (**add/remove**). The polled `qty_available` moves before the bot
observes anything — held at order acceptance (sells), filled at execution — so
a poll landing in the open-order window can already contain the fill, and the
delta then applies it a second time.

- **Loud variant** (fill > remaining): underflow → `InsufficientAvailable`.
  The MSTR incident, 2026-05-26. PR #690 stopped the resulting deadlock but
  left the race.
- **Silent variant** (fill ≤ remaining): the subtraction succeeds, the balance
  is short by one fill with no error, and the rebalancer reads the wrong number
  — firing needless mints/redemptions/bridges. The more common case.

Full design, alternatives, and accepted tradeoffs: [ADR 0015](https://github.com/agryaznov/st0x.liquidity/blob/fix/reconcile/adrs/0015-single-writer-reconciliation-for-hedging-equity-balance.md).

## Why

Wrong offchain balances make imbalance detection fire rebalancing transfers
that were not needed — wasted onchain gas and broker fees — every time a poll
lands within ~100ms of a fill, which under real load is often enough to matter.

## How

Give the balance a **single writer at any instant** by extending
`equity_snapshot_would_apply` with two guards (Hedging venue only):

1. **Ownership** — skip offchain snapshots for a symbol while a hedge order is
   open; the fill delta owns the balance until the order terminates.
2. **Reverse-ordering** — skip a snapshot fetched before the last applied fill
   (`fetched_at < last_offchain_fill_applied_at`).

Both compare a single clock (or none), avoiding the broker-vs-host cross-clock
defect a `broker_timestamp` comparison would have. Supporting changes:

- **Single-owned gate**: `pending_offchain_order_symbols` moves from
  `RebalancingService` into `InventoryView` (one copy under one lock, not two
  hand-synced copies — the class of bug behind PR #1013). The trigger reads it
  via `has_pending_offchain_order`.
- **`fetched_at` captured before the broker read** (in the poller, threaded
  through the command like `InflightEquity`), not stamped at command-handling
  time — so a pre-fill read cannot be stamped after the fill and slip past
  guard 2.
- **Recovery paths respect ownership**: `on_snapshot_recovery` preserves the
  gate across its view reset, and the force-apply path holds guard 1 (still
  bypasses the staleness guards, by design).
- **Boot ordering made structural**: `restore_inventory_at_boot` hydrates the
  view *before* seeding the gate — otherwise open-order symbols boot with an
  uninitialized offchain balance.
- **Gate-only mutations skip the dashboard broadcast**
  (`write_without_broadcast`); the gate is not in the DTO.
- Startup lifecycle-failure catch-up relocated before the resume workers and
  retry-wrapped for transient `SQLITE_BUSY`.

**Schema**: untouched — no new persisted fields, no `SCHEMA_VERSION` bump, no
migration.

## Testing

New unit + reactor-level coverage:

- The race: the silent variant converges instead of double-subtracting, and a
  pre-fill read handled after the fill does not resurrect the pre-fill balance
  (driven through the production stamping path).
- The guards: skips don't advance the watermark, the strict `<` boundary
  holds, `MarketMaking` snapshots are unaffected.
- Recovery: the reset keeps the gate and each gated symbol's delta-owned
  balance; the force path applies ungated symbols but holds guard 1.
- Boot: a symbol with an open hedge order both gates and hydrates.
- Gate-only writes don't broadcast dashboard snapshots.

Backend CI matrix green (`cargo check` both feature profiles, workspace
nextest, clippy `--all-targets --all-features`, `fmt --check`); no dashboard
changes in this PR.

## Anything else

**Accepted residuals** (all self-healing, detailed in the ADR's Tradeoffs):

- While a hedge order is open, that symbol's offchain balance is frozen against
  external truth (manual trades, corporate actions) — the same starvation the
  existing `has_inflight()` arm already accepts during rebalancing.
- A restart across an open order re-applies the ambiguous mid-order snapshot
  and double-counts once; the drift heals when the broker value next changes
  or on restart re-hydration. A retry mechanism that would bound this to one
  poll (forget-on-clear) was descoped in review to a follow-up PR.

**Open question** (in the ADR): whether guard 1 should carry a staleness bound
so a leaked gate cannot starve a symbol's snapshots indefinitely. Shipped
unbounded; a `warn!` after N consecutive skips would make it observable.

## Follow-up

1. **Skipped-snapshot retry (forget-on-clear)** — descoped in review after edge
   cases surfaced (poller-stream targeting under split owners among them);
   needs its own design pass. Until it lands, a snapshot skipped during an open
   order is retried only when the broker value changes or a restart
   re-hydrates. Prior art in git history: commits `be897734`/`48346b7c` and
   descendants; tests
   `skipped_offchain_snapshot_heals_after_gate_clears_despite_dedupe`,
   `restart_with_mid_order_snapshot_heals_within_one_poll`.
2. **The offchain cash leg** is double-counted in the same window and is not
   guarded here (it is venue-level, not per-symbol; needs a different
   mechanism).
3. **Long-term**: consume Alpaca's account-activities stream (an ordered op
   log), which dissolves the ambiguity at source and retires these guards
   entirely.

## Summary

- **Bug Fixes**
  - Improved offchain hedging equity accuracy by preventing double-counting when fills and snapshots overlap.
  - Added stronger handling for stale offchain equity snapshots, with correct retry/re-application once hedge conditions clear.
  - Preserved hedge-order gating state across startup, recovery, and restarts to support reliable balance “healing.”

- **Improvements**
  - Enhanced offchain inventory polling by including a fetch timestamp to improve snapshot ordering.
  - Streamlined boot-time inventory restoration when rebalancing is enabled.

- **Documentation**
  - Added an ADR explaining the hedging equity reconciliation approach.


src/conductor.rs
```diff
@@ -628,6 +628,13 @@ impl Conductor {
         // database. Without this, the first post-restart poll may emit
         // no events (unchanged values are deduplicated), leaving the
         // view empty and potentially causing incorrect rebalancing.
+        //
+        // Ordering: this must run before
```
JuaniRios: critical: this comment says hydrate runs before recover, but the actual order is reversed. `recover_pending_offchain_order_symbols` runs inside `PositionAndRebalancing::setup` (awaited up at line 600), so the pending set is already populated by the time `hydrate_inventory_from_snapshot` runs down here at 638. that means for any symbol with an open hedge order at restart, guard 1 skips its snapshot during hydration and the view boots with an uninitialized offchain balance, which is exactly the trap the ADR warns about. fix is to pull the recover call out of `setup` and run it right after `hydrate_inventory_from_snapshot` instead.

src/inventory/view.rs
```diff
@@ -1086,6 +1152,34 @@ impl InventoryView {
...
+                    target: "inventory",
+                    %symbol,
+                    ?fetched_at,
+                    "Skipping offchain equity snapshot: hedge order still open",
+                );
+                return Ok(false);
+            }
+
+            if self
+                .last_offchain_fill_applied_at
```
JuaniRios: should fix: guard 2 compares `fetched_at` against `last_offchain_fill_applied_at`, but `fetched_at` isn't the broker read time. `poll_offchain` sends `OffchainEquity` with no timestamp, and `InventorySnapshot::transition` stamps `fetched_at = Utc::now()` when the command is handled, which is after the read and after the fill can already be applied. so a stale pre-fill snapshot gets `fetched_at > filled_at` and the guard lets it through, overwriting the correctly decremented balance. the reverse-race the PR claims to close stays open, the tests only pass because they hand-pick the timestamp. to actually fix it, capture the read time in `poll_offchain` and carry it through the command into the event.

src/inventory/view.rs
```diff
@@ -598,11 +598,28 @@ pub(crate) struct InventoryView {
...
     /// Latest absolute equity snapshot timestamp by symbol for the offchain venue.
+    /// Local-clock time, as above.
     #[serde(default)]
     offchain_equity_snapshot_watermarks: HashMap<Symbol, DateTime<Utc>>,
+    /// Symbols with an open offchain (hedge) order. The broker bakes a fill
+    /// into its position the instant it executes, before we observe the fill,
+    /// so offchain snapshots are skipped for these symbols and the fill delta
+    /// owns the balance until the order reaches a terminal state.
+    #[serde(default)]
+    pending_offchain_order_symbols: HashSet<Symbol>,
```
JuaniRios, line 611: should fix: now that this set lives on `InventoryView`, `on_snapshot_recovery` (trigger/mod.rs ~1732) wipes it. that fn does `*inventory = InventoryView::default()` on any snapshot error for any symbol or venue, and nothing repopulates it after startup. so a USDC or onchain-equity recovery clears the open-hedge gate for every symbol, and the next offchain poll can write straight into a balance whose order is still open, re-opening the double-count race this PR is meant to close. either preserve these two fields across the reset, or re-derive them from the Position projection there the way startup recovery does.

src/inventory/view.rs
```diff
@@ -1086,6 +1152,34 @@ impl InventoryView {
...
 
+        // The offchain venue has a second writer (the fill delta) that the
+        // onchain venue does not; yield to it while it owns the balance.
+        if venue == Venue::Hedging {
+            if self.pending_offchain_order_symbols.contains(symbol) {
+                debug!(
+                    target: "inventory",
+                    %symbol,
+                    ?fetched_at,
+                    "Skipping offchain equity snapshot: hedge order still open",
```
JuaniRios: should fix: when a snapshot is skipped here the `InventorySnapshot` aggregate has already recorded it, and `transition` dedupes unchanged offchain positions. so once the gate clears, the same broker truth emits no new event and the skipped value never gets retried. the added test re-delivers the same event by hand but prod won't. if the skipped snapshot held real broker truth the delta didn't capture (or the fill update failed and cleared with `None`), the view stays stale until the position changes again or we restart. worth making skipped offchain snapshots replayable after the gate clears.

src/conductor.rs
```diff
@@ -628,6 +628,13 @@ impl Conductor {
...
+        // `recover_pending_offchain_order_symbols` (below). Hydration replays
+        // offchain equity snapshots through the pending-offchain-order guards,
+        // which are inert only while the pending set is still empty; seeding
+        // the set first would skip hydration for every symbol with an open
+        // hedge order, booting it with an uninitialized offchain balance.
```
JuaniRios: should fix: this ordering is load-bearing now (hydrate must run before recover, or open-order symbols boot with a zeroed offchain balance) but nothing tests it. the ADR even says 'nothing currently enforces the order'. a future refactor that swaps these two calls compiles and passes every other test while silently corrupting the hedging balance at boot. worth a test that drives the real hydrate-then-recover sequence with a persisted snapshot plus an open pending order and asserts the balance is applied. maybe extract the two calls into a small helper so there's a narrow seam to test.

src/rebalancing/trigger/mod.rs
```diff
@@ -2103,20 +2104,27 @@ impl Reactor for RebalancingService {
...
+                        // Also stops offchain snapshots from writing this
+                        // symbol's balance while the order is open: the broker
+                        // bakes a fill into its position the instant it
+                        // executes, which is before we observe the fill, so a
+                        // snapshot landing in that window already contains it.
+                        self.inventory
                             .write()
                             .await
-                            .insert(symbol);
+                            .mark_offchain_order_pending(symbol);
```
JuaniRios, line 2124: minor: `self.inventory` is a `BroadcastingInventory`, so `.write()` broadcasts a full dashboard snapshot on drop. before this move these arms only touched a plain `RwLock<HashSet>` with no broadcast. `pending_offchain_order_symbols` isn't in `to_dto()`, so every place/fail/cancel now fires a snapshot carrying no new info. not a correctness issue, but under active trading that's a lot of empty broadcasts and the side effect is invisible at the call site. either add a non-broadcasting write path for gate-only mutations, or just call it out as accepted.

src/inventory/view.rs
```diff
@@ -1086,6 +1152,34 @@ impl InventoryView {
             return Ok(false);
         }
 
+        // The offchain venue has a second writer (the fill delta) that the
+        // onchain venue does not; yield to it while it owns the balance.
+        if venue == Venue::Hedging {
```
JuaniRios: nit: `equity_snapshot_watermark` right above matches `Venue` exhaustively, but this uses `if venue == Venue::Hedging`. two variants today so it's correct, but a third venue that needed the same guard would compile and silently skip it. a `match` here would force the decision, which is the match-over-boolean thing in the guidelines.

src/rebalancing/trigger/mod.rs
```diff
@@ -2146,12 +2154,26 @@ impl Reactor for RebalancingService {
...
+                    if clear_pending_offchain_order {
+                        // Release the snapshot block in the same critical
+                        // section that applied the delta, so no snapshot can
+                        // land in between and overwrite the fill just recorded.
+                        // Stamp the local clock rather than reusing `timestamp`
+                        // (the event's `broker_timestamp`): this is compared
+                        // against snapshot `fetched_at`, which is stamped from
+                        // this host's clock.
+                        inventory
+                            .clear_offchain_order_pending(&symbol, result.is_ok().then(Utc::now));
```
JuaniRios, line 2176: nit: on a failed fill this clears with `None`, so a stale `last_offchain_fill_applied_at` is left in place. the ADR says that's the safe direction (can only over-block) and I agree, but it's untested on the exact path this PR fixes. a future change to the `.then(Utc::now)` could flip it unsafe with nothing catching it. worth extending `offchain_order_filled_releases_gate_even_when_inventory_update_fails` to seed a prior timestamp and assert a later snapshot still applies.

JuaniRios (changes requested):

src/inventory/view.rs
agryaznov: addressed: preserving the two fields across the reset

src/inventory/view.rs
agryaznov: addressed

src/inventory/view.rs
agryaznov: addressed: on gate clear, the trigger sends a new `ForgetOffchainEquity` command that marks the symbol undelivered in the aggregate's cache (a side-set, keeping the cached map intact for hydration), so the next poll re-emits broker truth with a fresh timestamp even when unchanged — healing the view within one poll while never replaying the ambiguous mid-order value itself

src/inventory/view.rs
agryaznov: nice catch! 
addressed: capture in `poll_offchain`; add `fetched_at` field on the `OffchainEquity` command ; `initialize/transition` pass it through instead of stamping `now` + a new test for this specific case added

src/rebalancing/trigger/mod.rs
agryaznov: addressed:  added `BroadcastingInventory::write_without_broadcast()`

src/rebalancing/trigger/mod.rs
agryaznov: addressed: extended the test

src/conductor.rs
agryaznov: addressed

src/conductor.rs
agryaznov: addressed

src/inventory/snapshot.rs
agryaznov, line 133: ## NOTE 
**`SCHEMA_VERSION` deliberately not bumped for `offchain_equity_forgotten`.**
  A bump makes the startup reconciler clear the aggregate's snapshots, which it
  refuses for a `CompactAfterSnapshot` aggregate (pre-snapshot events are
  already gone) — `ReconcileError::CompactedSnapshotClear` fails the deploy
  closed. The field is `#[serde(default)]` and an empty set is the semantically
  correct reading of every old payload, so no backfill is needed either. Same
  no-bump discipline as the `*_fetched_at` fields
  ([`db75a32f`](https://github.com/ST0x-Technology/st0x.liquidity/pull/675)),
  which additionally required a payload backfill migration because their `None`
  default was misleading for old snapshots; ours is not.

src/rebalancing/trigger/mod.rs
```diff
@@ -21896,6 +22003,482 @@ mod tests {
...
+                error,
+                InventorySnapshotEvent::OnchainUsdc {
+                    usdc_balance: usdc(500),
+                    fetched_at: Utc::now(),
+                },
+            )
+            .await
+            .unwrap();
+
+        let view = trigger.inventory.read().await;
```
JuaniRios: should fix: this code holds the `view` read guard across the three asserts. The guard stays open to the end of the function. In this workspace, `clippy::significant_drop_tightening` is a deny-level nursery lint. This lint fails on that pattern. As a result, `cargo clippy --workspace --all-targets` fails, and CI blocks the merge. Every other guard in this file uses a method chain or a block scope. This guard is the only bare one. Add `drop(view)` before the asserts. Or put the reads in a block.

src/conductor.rs
```diff
@@ -1086,6 +1087,33 @@ async fn compact_inventory_snapshot_events(pool: &SqlitePool) -> Result<u64, sql
...
+/// still empty. Seeding the gate first would skip hydration for every symbol
+/// with an open hedge order, booting it with an uninitialized offchain
+/// balance that imbalance detection could then misread.
+pub(crate) async fn restore_inventory_at_boot(
+    pool: &SqlitePool,
+    inventory: &Arc<BroadcastingInventory>,
+    rebalancing_service: Option<&Arc<RebalancingService>>,
+    position_projection: &Projection<Position>,
+) -> Result<(), ProjectionError<Position>> {
+    hydrate_inventory_from_snapshot(pool, inventory).await;
```
JuaniRios, line 1116: should fix: the hydrate step runs before the recover step. This order is correct for the zero-balance case. But the hydrated value can be wrong. The poller stores the offchain snapshot every cycle, and it ignores the view gate. So the latest stored snapshot can hold a mid-order value that the live view refused correctly. At restart, the code applies this value while the gate is still empty. Then the fill event arrives, and the fill delta subtracts again. The balance is now double-counted. The balance heals within one poll, because the Filled arm calls forget and defeats the dedup. So the error is not permanent. But an equity check runs on the wrong balance for about 60 seconds. This behavior is not new: the same window double-counted before this PR. I flag it because the PR claims to close this race. Add a restart test. In the test, capture the snapshot while the order is open. The current boot test uses a snapshot from before the order.

src/inventory/view.rs
```diff
@@ -1064,6 +1087,68 @@ impl InventoryView {
...
+    ///
+    /// Snapshot-error recovery resets the view before force-applying the
+    /// failed snapshot. The guard state must survive that reset: it is derived
+    /// from the `Position` event stream, not from snapshots, and nothing
+    /// re-seeds it after startup — wiping it would re-admit offchain snapshots
+    /// for symbols whose hedge order is still open, re-opening the
+    /// double-apply race. Every other field is intentionally defaulted, which
+    /// is why this uses functional-update syntax rather than an exhaustive
+    /// literal: a future field should default here unless it is guard state.
+    pub(crate) fn reset_preserving_offchain_order_state(&self) -> Self {
```
JuaniRios, line 1149: should fix: this function keeps the gate state across the reset. This part is correct. But `on_snapshot_recovery` then calls `force_apply_snapshot_event`. That function writes each symbol's Hedging balance from the snapshot, and it does not check `pending_offchain_order_symbols`. So it skips the two guards. One symbol can cause a snapshot error. That error sends every open-order symbol through the force-apply, and the double-count comes back. The balance heals within one poll, as in the restart case. So the error is not permanent. But it breaks the rule on the recovery path, and the recovery path exists for this case. Fix: skip the pending-set symbols in the OffchainEquity arm of force_apply. Use the same rule as guard 1.

JuaniRios (changes requested):

src/rebalancing/trigger/mod.rs
agryaznov: addressed

src/conductor.rs
agryaznov: addressed: admitted in the doc among Tradeoffs as a bounded residual (≤1 poll now); mid-order-snapshot restart test added

src/inventory/view.rs
agryaznov: addressed: force path holds guard 1, bypasses guard 2; + test added

src/rebalancing/trigger/mod.rs
```diff
@@ -1729,7 +1790,11 @@ impl RebalancingService {
             _ => None,
         };
         let mut inventory = self.inventory.write().await;
-        *inventory = InventoryView::default();
+        // Reset everything except the offchain-order guard state: it is fed by
+        // Position events, not snapshots, and nothing re-seeds it after
+        // startup, so a plain default() here would clear the open-hedge block
+        // for every symbol and re-open the double-apply race.
+        *inventory = inventory.reset_preserving_offchain_order_state();
```
JuaniRios, line 1740: should fix: this line resets the view. `force_apply_snapshot_event` then runs at line 1822. The reset keeps only the two guard fields. It clears `equities`. The sibling test `reset_preserves_offchain_order_guard_state_only` asserts this. The new gate skip in the `OffchainEquity` arm then skips a gated symbol. But no balance remains to keep. The symbol ends recovery with no entry. The comment there says the symbol keeps its delta-owned balance. That statement is not correct on this path.

Part of this is my fault. In one round I asked you to keep the guard fields across the reset. In the next round I asked you to skip gated symbols in the force path. I did not check how the two changes interact.

The effect is limited. A later sell fill returns `InsufficientAvailable`. It clears the gate. It waits for the next snapshot. A later buy fill seeds the balance from the delta alone, and it does this silently. While the venue is `None`, imbalance detection returns `Ok(None)`. Rebalancing for that symbol stops. It does not fire incorrectly. The heal then depends on forget-on-clear. That mechanism does not work in production. See my other comment.

The requirement is this: a gated symbol must end recovery with the balance that the fill delta owns. Select the approach that you prefer. Also add a test that runs through `on_snapshot_recovery`. Cover an all-gated snapshot and a mixed snapshot.

src/rebalancing/trigger/mod.rs
```diff
@@ -688,6 +695,56 @@ impl RebalancingService {
...
+                target: "rebalance",
+                %symbol,
+                "No snapshot store wired; a snapshot skipped during the open \
+                 order cannot be retried until the broker value changes"
+            );
+            return;
+        };
+
+        let snapshot_id = InventorySnapshotId {
+            orderbook: self.registry_id.orderbook,
```
JuaniRios: should fix: this code builds the target id from `registry_id`. The owner of `registry_id` is `ctx.vault_owner()`. But the poller writes every `OffchainEquity` snapshot to a different id. That id uses `ctx.order_owner()` as the owner. See `conductor/builder.rs`. The two values come from two different config fields.

Under `inventory_mode = "managed"`, `vault_owner` must be the inventory contract. `[wallet].address` is the signing EOA. So the two addresses are always different.

The forget command then goes to a stream that the poller never writes. On that stream the `initialize` arm returns `Ok(vec![])`. The `transition` arm also returns `Ok(vec![])`. There is no event, no error, and no log. The `warn!` below never runs.

So forget-on-clear never heals the view in production. No other code re-sends the value. The unchanged-value dedupe blocks it, and the forget exists to defeat that dedupe. This also removes the basis for the ADR claim that the restart residual is at most one poll.

The tests do not catch this. They use the same address for both ids. Fix: pass the poller's `InventorySnapshotId` into the service, and use one construction site. Add a test where `vault_owner` is different from `order_owner`.

src/inventory/view.rs
```diff
@@ -2985,6 +3128,216 @@ mod tests {
...
+    }
+
+    /// The recovery force path bypasses the staleness guards by design, but
+    /// it must not bypass the open-hedge gate: a symbol with an open order
+    /// owns its balance via the fill delta, and force-writing the ambiguous
+    /// mid-order snapshot value re-opens the double-count the gate exists to
+    /// prevent -- on the recovery path, which exists precisely for when
+    /// things already went wrong.
+    #[test]
+    fn force_apply_offchain_snapshot_respects_open_hedge_gate() {
```
JuaniRios, line 3282: minor: this test seeds the balance with `with_equity`. It then calls `force_apply_snapshot_event` directly. But the rebalancing-service recovery path (`on_snapshot_recovery`) resets the view first. So a gated symbol has no balance when the force path runs. The result of 100 shares cannot occur on that path. This test therefore cannot find the reset problem.

`InventoryProjection::apply` also force-applies without a reset. But it runs only when rebalancing is off. The gate is never set in that mode.

Fix: add a test that runs through `on_snapshot_recovery` with a gated symbol. Cover an all-gated snapshot and a mixed snapshot.

src/rebalancing/trigger/mod.rs
```diff
@@ -2103,20 +2171,28 @@ impl Reactor for RebalancingService {
                         )
                     }
                     OffChainOrderPlaced { .. } => {
-                        self.pending_offchain_order_symbols
-                            .write()
+                        // Also stops offchain snapshots from writing this
+                        // symbol's balance while the order is open: the broker
+                        // bakes a fill into its position the instant it
```
JuaniRios: nit: this comment says that the broker adds a fill to its position at the moment of execution. But the snapshot reads the `qty_available` field. See `crates/execution/src/alpaca_broker_api/positions.rs:31`. For a sell hedge, that value decreases when the broker accepts the order and holds the shares. This occurs before execution.

The guards still work correctly. Guard 1 covers the full window from `OffChainOrderPlaced` to the terminal event. That window includes the acceptance-time decrease. Only the comment needs a correction.

JuaniRios (changes requested):

src/rebalancing/trigger/mod.rs
agryaznov: fixed with this approach: `reset_preserving_offchain_order_state()` now preserves delta-owned Hedging available balances (its inflight and its onchain venue balances are still reset); tests updated

src/rebalancing/trigger/mod.rs
agryaznov: fixed

src/conductor.rs
```diff
@@ -1527,6 +1561,13 @@ fn spawn_rebalancing_infrastructure<Chain: Wallet + Clone>(
             )
             .await;
 
+        rebalancing_service
+            .set_snapshot_store(
```
JuaniRios: should fix: this call now spans three lines, and that tips `spawn_rebalancing_infrastructure` over the `clippy::too_many_lines` limit at 202/200. The lint is deny-level here, so `cargo clippy -p st0x-hedge --all-targets` fails and CI blocks the merge.

AGENTS.md asks you to extract a genuine logical chunk first, and to treat `#[allow]` as a last resort that needs explicit permission. This function does several separable things, so one extraction should bring it under the limit.

This is the second round where a fix commit broke a deny-level lint. Please run `cargo clippy -p st0x-hedge --all-targets` (or `nix run .#ci`) before you push.

src/inventory/snapshot.rs
```diff
@@ -269,12 +283,26 @@ impl EventSourced for InventorySnapshot {
                     fetched_at: now,
                 }])
             }
-            OffchainEquity { positions } => {
-                if self.offchain_equity == positions {
+            OffchainEquity {
+                positions,
+                fetched_at,
+            } => {
+                if self.offchain_equity == positions && self.offchain_equity_forgotten.is_empty() {
```
JuaniRios: scope call, not a bug: please remove `ForgetOffchainEquity` from this PR. The command, the event, the `offchain_equity_forgotten` set, and the wiring can all come out. I found more edge cases surrounding this part of the PR, and since it wasn't part of the original scope better to drop it.

Don't worry though, it's on me to suggest addressing this here since I thought it would be a oneoff fix, but turned out to be a bigger change than it looked, and can be addressed in a later PR with a better plan.

JuaniRios (changes requested):

src/inventory/snapshot.rs
agryaznov: this one was moved out of the PR scope anyway, closing

src/conductor.rs
agryaznov: fixed

src/inventory/snapshot.rs
agryaznov: rolled that part back, moving the solution for the problem you nailed in [this comment](https://github.com/ST0x-Technology/st0x.liquidity/pull/1057#discussion_r3617339485) out of the scope of this PR (mentioned in the follow-up section now)

JuaniRios (approved):

PR #1056, opened 2026-07-20
fix: preserve dashboard trade protocol compatibility

JuaniRios:
## What

[RAI-1429](https://linear.app/makeitrain/issue/RAI-1429/preserve-dashboard-trade-protocol-compatibility-across-rolling-deploys)

Keeps trade history working across mixed backend/dashboard versions while failed counter-trades roll out.

## Why

RAI-1418 expands the trade payload from fills to terminal outcomes. Without an explicit compatibility contract, an old dashboard can render failures as fills, while a new dashboard can lose fills when connected to an old backend.

## How

- Defaults REST and WebSocket clients to the legacy filled-only protocol.
- Lets the new dashboard opt into `terminal_outcomes_v1`.
- Emits canonical `trade_update` messages plus legacy `trade_fill` messages for fills.
- Normalizes legacy REST, snapshot, and live rows in the new dashboard.
- Uses full RFC 3339 precision and the same deterministic ID tie-breaker in backend and frontend ordering.

## Testing

- Added mixed-version REST and WebSocket snapshot/live coverage in both directions.
- Added sub-millisecond and exact-tie ordering coverage.
- `cargo nextest run -p st0x-dto`
- Focused `st0x-hedge` compatibility tests
- Focused dashboard Vitest suite
- `cargo clippy -p st0x-hedge --all-targets --all-features`
- `bun run check`
- `bun run lint`

## Screenshots

N/A - protocol compatibility only.

## Anything else

Stacked on #1053.

## Summary by CodeRabbit

- **New Features**
  - Added support for terminal trade outcomes (including filled and failed) with protocol-aware REST and WebSocket trade messaging.
  - Introduced an opt-in query parameter to enable terminal-outcome trade protocol in newer dashboards.
  - Legacy dashboards continue to receive filled trades in the legacy trade-fill format.

- **Bug Fixes**
  - Trade history normalization now correctly maps legacy payloads to the modern UI.
  - Ordering now preserves sub-millisecond timestamps with stable tie-breaking.
  - Duplicate trades are de-duplicated while maintaining newest-first ordering.

PR #1058, opened 2026-07-20
fix: make terminal trade broadcasts durable

JuaniRios:
## What

[RAI-1431](https://linear.app/makeitrain/issue/RAI-1431/make-terminal-trade-broadcasts-durable-and-replayable)

Makes terminal dashboard trade delivery durable and replayable across transient failures and process restarts.

## Why

- The domain event could persist while the in-process broadcast failed. Active dashboard sessions then missed a terminal trade until a reload.
- A restart could make that live-stream gap permanent.

## How

- Persists terminal deliveries in an idempotent ledger and Apalis job queue keyed by trade ID.
- Retries failed reactor handoffs in-process from a bounded queue (64 entries, 3 attempts with exponential backoff).
- An exhausted or poison handoff cannot block later trades: a periodic authoritative reconciliation rebuilds every undelivered terminal trade from the event log, and refreshes stale job payloads in place.
- A handoff whose trade can never convert fail-stops the monitor through the supervisor instead of retrying forever.
- Reconciles missing, in-flight, failed, and exhausted deliveries at startup.
- Publishes the canonical and legacy events before marking delivery complete, while the dashboard deduplicates retries by trade ID.
- Wires both production trade stores through the durable broadcaster.

## Testing

- Replayed all 11,358 persisted aggregates from a consistent production snapshot through the exact migration tree with zero failures.
- Added coverage for transient failures, restart recovery, exhausted jobs and handoffs, poison-handoff isolation, deterministic conversion failures, duplicate delivery, stale payload refresh, and both trade stores.
- `cargo nextest run -p st0x-hedge dashboard::event` is green on every branch of the stack; full CI runs remotely.

## Screenshots

Not applicable.

## Anything else

- Stacked on #1056.
- The bounded-queue and periodic-reconciliation hardening moved down from #1063 into this PR, so the delivery mechanism lives in one place.

## Summary by CodeRabbit

* **New Features**
  * Added reliable live dashboard updates for completed trades, including retry and restart recovery.
  * Added reconciliation to restore undelivered trade updates after interruptions.
  * Expanded trade history loading before applying display limits.
* **Bug Fixes**
  * Trade payloads can now be safely saved and restored without losing field names or timestamp precision.
* **Documentation**
  * Documented dashboard trade delivery and updated migration verification commands.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -286,8 +286,8 @@ create and migrate the local database before any cargo commands.
 **Verifying a migration against real data**: `sqlx db reset -y` only proves a
 migration applies to an empty schema, not that it behaves correctly against real
 prod/staging data or that legacy events still deserialize under current code.
-While developing a migration, run `nix run .#prod-verify-migrations` (or
-`.#staging-verify-migrations`) -- it takes a _consistent_ server-side
+While developing a migration, run `nix run .#prodVerifyMigrations` (or
+`.#stagingVerifyMigrations`) -- it takes a _consistent_ server-side
 `VACUUM INTO` snapshot of the live database (`<env>-db-snapshot` in
 `infra/default.nix`; a plain `scp` of a live WAL-mode SQLite file can race the
 bot's writes and download a torn/corrupt copy, so never `scp` the `.db` file
--- a/src/conductor/builder.rs
+++ b/src/conductor/builder.rs
@@ -34,6 +34,10 @@ use super::monitor::gas::{GasMonitor, ProviderBalanceReader};
 use super::monitor::inventory::InventoryMonitor;
 use super::monitor::order_fills::OrderFillMonitor;
 use crate::alerts::TelegramNotifier;
+use crate::dashboard::{
+    DashboardTradeDeliveryCtx, DashboardTradeDeliveryJobQueue, DashboardTradeHandoffMonitor,
+    DeliverDashboardTrade,
+};
 use crate::inventory::{
     BroadcastingInventory, InventoryDivergenceRecoveryCtx, InventoryPollingService,
     InventorySnapshot, InventorySnapshotId, WalletPollingCtx,
@@ -156,6 +160,9 @@ pub(crate) fn spawn<Prov, Exec>(
     context: ConductorCtx<Prov, Exec>,
     job_queue: DexTradeAccountingJobQueue,
     backfill_queue: BackfillJobQueue,
+    dashboard_trade_delivery_queue: DashboardTradeDeliveryJobQueue,
+    dashboard_trade_delivery_ctx: Arc<DashboardTradeDeliveryCtx>,
+    dashboard_trade_handoff_monitor: DashboardTradeHandoffMonitor,
     hedge_queue: HedgeJobQueue,
     poll_status_queue: PollOrderStatusJobQueue,
     reconcile_queue: ReconcileOrderFillJobQueue,
@@ -373,7 +380,11 @@ where
         .with_base_restart_delay(std::time::Duration::from_secs(1))
         .with_dead_tasks_threshold(Some(0.0))
         .with_task("order-fill-monitor", order_fill_monitor)
-        .with_task("inventory-monitor", inventory_monitor);
+        .with_task("inventory-monitor", inventory_monitor)
+        .with_task(
+            "dashboard-trade-handoff-monitor",
+            dashboard_trade_handoff_monitor,
+        );
 
     log_optional_task_status("executor maintenance", maintenance_interval.is_some());
 
@@ -404,6 +415,8 @@ where
         job_queue,
         hedge_queue,
         backfill_queue,
+        dashboard_trade_delivery_queue,
+        dashboard_trade_delivery_ctx,
         poll_status_queue,
         reconcile_queue,
         rejection_queue,
@@ -463,6 +476,8 @@ where
     job_queue: DexTradeAccountingJobQueue,
     hedge_queue: HedgeJobQueue,
     backfill_queue: BackfillJobQueue,
+    dashboard_trade_delivery_queue: DashboardTradeDeliveryJobQueue,
+    dashboard_trade_delivery_ctx: Arc<DashboardTradeDeliveryCtx>,
     poll_status_queue: PollOrderStatusJobQueue,
     reconcile_queue: ReconcileOrderFillJobQueue,
     rejection_queue: HandleOrderRejectionJobQueue,
@@ -510,6 +525,8 @@ where
             job_queue,
             hedge_queue,
             backfill_queue,
+            dashboard_trade_delivery_queue,
+            dashboard_trade_delivery_ctx,
             poll_status_queue,
             reconcile_queue,
             rejection_queue,
@@ -541,6 +558,8 @@ where
         #[cfg(any(test, feature = "test-support"))]
         let failure_injector_for_backfill = failure_injector.clone();
         #[cfg(any(test, feature = "test-support"))]
+        let failure_injector_for_dashboard_trade_delivery = failure_injector.clone();
+        #[cfg(any(test, feature = "test-support"))]
         let failure_injector_for_poll = failure_injector.clone();
         #[cfg(any(test, feature = "test-support"))]
         let failure_injector_for_reconcile = failure_injector.clone();
@@ -571,6 +590,7 @@ where
         let failure_notify = Arc::new(tokio::sync::Notify::new());
         let failure_notify_for_hedge = failure_notify.clone();
         let failure_notify_for_backfill = failure_notify.clone();
+        let failure_notify_for_dashboard_trade_delivery = failure_notify.clone();
         let failure_notify_for_poll = failure_notify.clone();
         let failure_notify_for_reconcile = failure_notify.clone();
         let failure_notify_for_rejection = failure_notify.clone();
@@ -589,6 +609,7 @@ where
             .with_recovery_timeout(FAIL_STOP_RECOVERY_TIMEOUT);
         let fail_stop_for_hedge = fail_stop.clone();
         let fail_stop_for_backfill = fail_stop.clone();
+        let fail_stop_for_dashboard_trade_delivery = fail_stop.clone();
         let fail_stop_for_poll = fail_stop.clone();
         let fail_stop_for_reconcile = fail_stop.clone();
         let fail_stop_for_rejection = fail_stop.clone();
@@ -615,6 +636,18 @@ where
                         failure_notify.clone(),
                         #[cfg(any(test, feature = "test-support"))]
                         failure_injector.clone(),
+                    )
+                })
+                .register(move |index| {
+                    build_supervised_worker!(
+                        ::<DashboardTradeDeliveryCtx, DeliverDashboardTrade>,
+                        index,
+                        dashboard_trade_delivery_queue.clone(),
+                        dashboard_trade_delivery_ctx.clone(),
+                        fail_stop_for_dashboard_trade_delivery.clone(),
+                        failure_notify_for_dashboard_trade_delivery.clone(),
+                        #[cfg(any(test, feature = "test-support"))]
+                        failure_injector_for_dashboard_trade_delivery.clone(),
                     )
                 })
                 .register(move |index| {
--- a/src/dashboard/event.rs
+++ b/src/dashboard/event.rs
@@ -1,16 +1,30 @@
 //! Reactor that broadcasts aggregate events to WebSocket dashboard
 //! clients as [`Trade`] fills and [`TransferOperation`] updates.
 
+use anyhow::Context;
 use async_trait::async_trait;
+use serde::{Deserialize, Serialize};
 use sqlx::SqlitePool;
-use tokio::sync::broadcast;
-use tracing::{debug, warn};
+use std::collections::VecDeque;
+use std::sync::Arc;
+#[cfg(test)]
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::time::Duration;
+use task_supervisor::{SupervisedTask, TaskResult};
+#[cfg(test)]
+use tokio::sync::Notify;
+use tokio::sync::{Mutex, broadcast, mpsc};
+use tokio::time::{Instant, interval, sleep_until};
+use tracing::{debug, info, warn};
 
 use st0x_dto::{Statement, Trade, TradeOutcome, TradingVenue};
-use st0x_event_sorcery::{EntityList, Never, Reactor, deps, load_entity};
+use st0x_event_sorcery::{EntityList, Reactor, SendError, deps, load_entity};
 
+use crate::conductor::job::{Job, JobQueue, Label, QueuePushError};
 use crate::equity_redemption::EquityRedemption;
-use crate::offchain::order::{OffchainOrder, OffchainOrderEvent};
+use crate::offchain::order::{
+    OffchainOrder, OffchainOrderEvent, OffchainOrderId, TradeConversionError,
+};
 use crate::onchain_trade::{OnChainTrade, OnChainTradeEvent};
 use crate::position::{Position, PositionEvent};
 use crate::tokenized_equity_mint::TokenizedEquityMint;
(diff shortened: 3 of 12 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1058?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1062** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1062?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1058** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1058?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1058?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1053** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1053?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1058).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

0xgleb (commented):
agent misfire

AGENTS.md
```diff
@@ -286,8 +286,8 @@ create and migrate the local database before any cargo commands.
 **Verifying a migration against real data**: `sqlx db reset -y` only proves a
 migration applies to an empty schema, not that it behaves correctly against real
 prod/staging data or that legacy events still deserialize under current code.
-While developing a migration, run `nix run .#prod-verify-migrations` (or
-`.#staging-verify-migrations`) -- it takes a _consistent_ server-side
+While developing a migration, run `nix run .#prodVerifyMigrations` (or
+`.#stagingVerifyMigrations`) -- it takes a _consistent_ server-side
```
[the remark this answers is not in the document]
JuaniRios: These do resolve. `verifyMigrationsPkgs` is merged into the `packages` output at `flake.nix:521`, and `nix eval .#prodVerifyMigrations.name` returns `prod-verify-migrations`. The attribute is camelCase and the derivation's binary name is kebab-case; this commit changed the docs to the attribute names, which is the form `nix run` needs.

src/conductor/builder.rs
```diff
@@ -605,6 +626,18 @@ where
...
+                        ::<DashboardTradeDeliveryCtx, DeliverDashboardTrade>,
+                        index,
+                        dashboard_trade_delivery_queue.clone(),
+                        dashboard_trade_delivery_ctx.clone(),
+                        fail_stop_for_dashboard_trade_delivery.clone(),
+                        failure_notify_for_dashboard_trade_delivery.clone(),
+                        #[cfg(any(test, feature = "test-support"))]
+                        failure_injector_for_dashboard_trade_delivery.clone(),
+                    )
+                })
```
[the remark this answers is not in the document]
JuaniRios: Confirming this is deliberate. A trade we executed but cannot report is a state we do not want the bot to keep trading through, so dashboard delivery shares the fail-stop path on purpose. SPEC.md and docs/conductor.md both state it. Reconciliation on restart is what makes the stop recoverable rather than lossy.

src/dashboard/event.rs
```diff
@@ -28,193 +42,1582 @@ deps!(
...
+
+    fn next_attempt(pending: &VecDeque<ScheduledDashboardTradeHandoff>) -> Option<Instant> {
+        pending.iter().map(|scheduled| scheduled.next_attempt).min()
+    }
+
+    async fn reconcile_after_exhaustion(&self, needed: &mut bool) -> anyhow::Result<()> {
+        self.enqueuer.reconcile_undelivered(&self.pool).await?;
+        *needed = false;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
JuaniRios: Agreed on the periodic path, and coalescing the triggers is the right fix. Tracked in RAI-1527. Not taking the suggested bound to `delivered_at IS NULL` though: the full replay is what makes reconciliation authoritative, because it also catches trades never registered in the delivery table at all. Bounding to that table would keep missing exactly the trades this exists to recover.

findolor (approved):

src/dashboard/event.rs
```diff
@@ -28,193 +42,1582 @@ deps!(
...
+            "UPDATE Jobs SET job = ? \
+             WHERE job_type = ? AND idempotency_key = ? AND status != 'Done'",
+        )
+        .bind(payload)
+        .bind(std::any::type_name::<DeliverDashboardTrade>())
+        .bind(idempotency_key)
+        .execute(self.queue.pool())
+        .await?;
+
+        Ok(())
```
rouzwelt, line 216: Refresh UPDATE cannot repair a Done job whose ledger row is undelivered.

status != 'Done' means the crash-after-publish window (job completed, mark_delivered never committed) yields: is_delivered false → counted as undelivered → push_idempotent deduped against the Done row (if apalis keys across terminal rows) → refresh matches nothing. The trade is reported as reconciled but nothing runs. Pair this with the push_idempotent verification on src/conductor/job.rs; if dedupe does span terminal rows, the reset in reconcile must also cover Done rows whose ledger entry is undelivered.

src/dashboard/event.rs
```diff
@@ -28,193 +42,1582 @@ deps!(
...
-    sender: broadcast::Sender<Statement>,
-    pool: SqlitePool,
-}
+pub(crate) type DashboardTradeDeliveryJobQueue = JobQueue<DeliverDashboardTrade>;
+
+const HANDOFF_RETRY_INITIAL_DELAY: Duration = Duration::from_secs(1);
+const HANDOFF_RETRY_MAX_DELAY: Duration = Duration::from_secs(30);
+const HANDOFF_RETRY_MAX_ATTEMPTS: usize = 3;
+const HANDOFF_RETRY_QUEUE_CAPACITY: usize = 64;
+const HANDOFF_RECONCILIATION_INTERVAL: Duration = Duration::from_secs(30);
```
rouzwelt, line 51: Move the delivery retry/reconciliation timings into configuration.

Backoff bounds, attempt budget, retry-queue capacity, and the reconciliation interval are all operational tuning knobs, yet they are compile-time constants here while comparable knobs (position_check_interval, inventory_poll_interval, apalis_finished_job_cleanup_interval_secs) come from Ctx. Operators cannot change delivery recovery behavior without a rebuild.

As per coding guidelines: "Operational parameters must be explicitly configured; missing configuration must fail rather than silently using defaults."

src/dashboard/event.rs
```diff
@@ -28,193 +42,1582 @@ deps!(
...
+        if reset > 0 || undelivered > 0 {
+            info!(
+                target: "dashboard",
+                reset,
+                undelivered,
+                "Reconciled durable dashboard trade deliveries",
+            );
+        }
+
+        Ok(undelivered)
```
rouzwelt, line 144: Unconditional attempts = 0 reset can re-drive a poison delivery on every restart.

Failed/Killed rows are reset with a fresh retry budget each startup. Combined with the fail-stop circuit breaker, a permanently undeliverable payload will exhaust retries, trip the monitor, and be re-driven again after the systemd restart — a restart loop with no visibility gate. Consider bounding the redrive (e.g. only reset when the ledger row is undelivered and a redrive counter is below a limit) or emitting an explicit operator alert when a reset row has previously exhausted its budget.

Also: reconcile(&self, pool) takes a pool the delivery store already owns; passing it again invites callers to supply a different database than the ledger writes to.

src/dashboard/event.rs
```diff
@@ -28,193 +42,1582 @@ deps!(
...
+            Self::Persistence(_) | Self::Replay { .. } | Self::Missing { .. } => true,
+            Self::Conversion { source, .. } => matches!(
+                source,
+                TradeConversionError::Pending
+                    | TradeConversionError::Submitted
+                    | TradeConversionError::PartiallyFilled
+                    | TradeConversionError::Cancelling
+            ),
+        }
+    }
```
rouzwelt, line 488: A cancelled offchain order should not fail-stop the handoff monitor.

TradeConversionError::Cancelled is classified non-retryable, which propagates DeterministicConversion and takes down the supervised monitor. But trade_loader::load_offchain_trades deliberately treats Cancelled as "not a deliverable trade" and skips it. A cancellation racing an earlier Filled/Failed handoff therefore kills a live task (and blocks every later handoff until restart) for an outcome the rest of the system considers benign.

```
-    async fn persist_once(
+    async fn persist_once(
         &self,
         handoff: &DashboardTradeHandoff,
-    ) -> Result<(), DashboardTradeHandoffAttemptError> {
+    ) -> Result<(), DashboardTradeHandoffAttemptError> {
```

Then treat the cancelled conversion as a successful no-op instead of an error, e.g. return Ok(()) from persist_once when try_into_trade yields TradeConversionError::Cancelled (logging a warning), and drop Cancelled from the retryable list.

src/dashboard/event.rs
```diff
@@ -28,193 +42,1582 @@ deps!(
...
+        Label::new(format!("DeliverDashboardTrade:{}", self.trade.id))
+    }
+
+    async fn perform(&self, ctx: &DashboardTradeDeliveryCtx) -> Result<Self::Output, Self::Error> {
+        if ctx.store.is_delivered(&self.trade.id).await? {
+            return Ok(());
+        }
+
+        ctx.publish(self.trade.clone());
+        ctx.store.mark_delivered(&self.trade.id).await
```
rouzwelt, line 665: Log the already-delivered short-circuit.

A replayed job that finds delivered_at set returns Ok(()) with no trace, so a redelivery storm is invisible in logs.

As per coding guidelines: "Log a warning or error before silent early returns such as let-else failure paths."

```
         if ctx.store.is_delivered(&self.trade.id).await? {
+            debug!(
+                target: "dashboard",
+                trade_id = %self.trade.id,
+                "Skipping dashboard trade delivery; ledger already records completion",
+            );
             return Ok(());
         }
```

src/dashboard/event.rs
```diff
@@ -28,193 +42,1582 @@ deps!(
...
+                tokio::select! {
+                    handoff = self.receive() => {
+                        let handoff = handoff.ok_or(DashboardTradeHandoffMonitorError::QueueClosed)?;
+                        pending.push_back(ScheduledDashboardTradeHandoff::new(handoff));
+                    }
+                    _ = reconciliation.tick(), if reconciliation_needed => {
+                        self.reconcile_after_exhaustion(&mut reconciliation_needed).await?;
+                    }
+                }
+            }
```
rouzwelt, line 443: nit:

Collapse the three near-identical select! blocks.

The full-queue, timed, and untimed branches duplicate receive/reconcile arms; the full-queue arm even uses QueueClosed for a next_attempt that cannot be None there, which would report a misleading cause if the invariant ever breaks. Extracting a single select! with an optional sleep future (OptionFuture or sleep_until(next_attempt.unwrap_or(far_future))) and a pending.len() < CAPACITY guard on the receive arm removes the duplication and the bogus error path.

rouzwelt (changes requested):

src/dashboard/event.rs
JuaniRios: Confirmed and fixed in ca9acbac. The apalis unique index is `(job_type, idempotency_key)` with `ON CONFLICT DO NOTHING`, so dedupe does span terminal rows. The refresh UPDATE now covers every status and resets the row to Pending; it only runs for ledger-undelivered trades, so reclaiming a Done row is safe. Added a test simulating the crash-after-publish window.

src/dashboard/event.rs
JuaniRios: Agreed these are operational knobs. Deferring to RAI-1579 rather than doing it here: threading five new required config fields through Ctx, the example configs, and the deployed secrets/config is its own change, and I'd rather not grow this PR further. The poison-redrive visibility gap you flagged separately is addressed in this PR.

src/dashboard/event.rs
JuaniRios: Fixed in ca9acbac with the alert flavor: reconcile now counts Failed/Killed rows before resetting them and warns that a repeat of the warning across restarts indicates a poison delivery. A redrive counter would need a schema change, so I kept the bound out for now; the fail-stop still limits each cycle to one budget. Also dropped the `pool` parameter -- reconcile now uses the ledger store's own pool.

src/dashboard/event.rs
JuaniRios: Fixed in ca9acbac. `persist_once` now completes a cancelled conversion as a logged no-op instead of returning a non-retryable error, so the monitor survives. Adapted the fail-stop test into one proving the monitor keeps processing after a cancelled handoff.

src/dashboard/event.rs
JuaniRios: Fixed in ca9acbac with the debug log as suggested.

src/dashboard/event.rs
JuaniRios: Fixed in ca9acbac. Single select! with a capacity guard on the receive arm and a guarded sleep arm; the bogus QueueClosed-for-missing-deadline path is gone.

src/dashboard/event.rs
```diff
@@ -1,220 +1,1726 @@
...
+    }
+
+    fn next_attempt(pending: &VecDeque<ScheduledDashboardTradeHandoff>) -> Option<Instant> {
+        pending.iter().map(|scheduled| scheduled.next_attempt).min()
+    }
+
+    async fn reconcile_after_exhaustion(&self, needed: &mut bool) -> anyhow::Result<()> {
+        self.enqueuer.reconcile_undelivered().await?;
+        *needed = false;
+        Ok(())
```
rouzwelt: A transient database error here kills the monitor and discards pending retries.

reconcile_after_exhaustion returns anyhow::Result<()>, and run propagates it with ? into TaskResult. Any error from reconcile_undelivered therefore terminates the supervised task. reconcile_undelivered performs many SQLite statements, so a single SQLITE_BUSY under write contention is enough.

Two consequences follow:

The restart drops the entire in-memory pending queue. Those handoffs then depend on the next reconciliation pass, which is the operation that just failed.
builder.rs configures with_dead_tasks_threshold(Some(0.0)) and 10 restart attempts. Repeated transient failures escalate to a full bot shutdown.
Treat a failed reconciliation pass as retryable: log it, leave reconciliation_needed set, and let the next tick retry. Reserve task termination for the deterministic conversion failure that attempt already returns.

src/dashboard/event.rs
```diff
@@ -1,220 +1,1726 @@
...
+        self
+    }
+
+    #[cfg(test)]
+    fn exhaustion_notification(&self) -> Arc<Notify> {
+        self.exhaustion_notify.clone()
+    }
+
+    async fn receive(&self) -> Option<DashboardTradeHandoff> {
+        self.receiver.lock().await.recv().await
```
rouzwelt, line 294: Clone on the monitor lets a second consumer deadlock on the receiver.

DashboardTradeHandoffMonitor derives Clone and shares one mpsc::Receiver behind Arc<Mutex<..>>. receive holds the mutex guard across .recv().await. With one running instance this is correct, and builder.rs registers exactly one. With two running instances the design breaks in two ways:

The instance that loses the mutex race blocks on lock() for as long as the winner waits on an empty channel. It never processes its own pending retries, because the select! arm that would fire is the one blocked inside receive.
Each clone keeps a private pending queue and a private reconciliation_needed flag, so retry state and capacity accounting split between them.
Only test code clones today, so this is not currently exploitable. The type does not prevent it though. Make the invariant structural: keep the owned mpsc::Receiver in the monitor, drop Clone, and give the tests a constructor that yields the single instance plus the sender. This also removes the mutex and the guard-across-await pattern.

As per coding guidelines: "Use domain newtypes, enums, ADTs, and typestate to make invalid states unrepresentable."

rouzwelt (changes requested):

src/dashboard/event.rs
JuaniRios: Fixed in 47acf30f. `reconcile_after_exhaustion` no longer returns a `Result`: it logs the failure and leaves `reconciliation_needed` set, so the next tick retries and the pending queue survives. Task termination stays reserved for the deterministic conversion failure that `attempt` returns.

Added `failed_reconciliation_pass_keeps_the_monitor_running`, which closes the pool so every reconciliation pass fails, then asserts the monitor is still running.

src/dashboard/event.rs
JuaniRios: I looked at dropping `Clone`, but the supervisor requires it: `task-supervisor` blankets `DynSupervisedTask` over `T: SupervisedTask + Clone` so it can `clone_box` the task on restart. The `Arc<Mutex<Receiver>>` is what lets a restarted clone resume from the same channel instead of losing the queue.

Only one instance runs at a time, so both failure modes you describe need two concurrently running monitors, which the supervisor never creates. Agreed that the type does not prevent it. Making that structural means moving the receiver outside the supervised task, which is a larger change than this PR.

The related fragility you found in the sibling thread is fixed in 47acf30f: a failed reconciliation pass no longer terminates the monitor, so restarts are much rarer.

rouzwelt (approved):

PR #1060, opened 2026-07-20
fix: validate dashboard trade payloads at runtime

JuaniRios:
## What

[RAI-1433](https://linear.app/makeitrain/issue/RAI-1433/validate-dashboard-trade-payloads-at-runtime)

Validate dashboard trade payloads at runtime before they enter REST or WebSocket state, and encode positive total trade quantity in the Rust DTO.

## Why

Generated TypeScript types only protect compile-time callers. Malformed runtime JSON could previously enter the dashboard cache and render misleading financial data.

## How

- Add one runtime parser for canonical and legacy trade payloads, including timestamps, enum variants, outcome shapes, and decimal quantity invariants.
- Preserve known-good history when REST refreshes, pagination, snapshots, or live updates are rejected.
- Surface WebSocket protocol failures in the dashboard and retain reconnect backoff until a valid message arrives.
- Make `Trade.shares` and `LegacyTrade.shares` positive quantities in Rust, propagating typed conversion failures.
- Keep pagination retries on the failed offset and prevent polling from superseding an in-flight page load.

## Testing

- `nix run .#ci`
- 42 focused dashboard tests covering payload validation, cache preservation, WebSocket errors/backoff, and pagination races
- Final Codex review: no actionable regressions

## Screenshots

Not applicable.

## Anything else

Stacked on #1058.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive validation for trade data received through snapshots, live updates, and paginated history.
  * Invalid trade payloads now display clear errors while preserving the last known-good data.
  * Connection errors now show specific details, retry attempts, and reconnect countdowns.

* **Bug Fixes**
  * Prevented invalid or non-positive trade quantities from entering dashboard state.
  * Improved trade-history pagination retries and prevented polling from interrupting older-page loads.
  * Added validation for timestamps, trade fields, outcomes, and pagination metadata.

PR #1062, opened 2026-07-21
fix: preserve counter-trade quantity provenance

JuaniRios:
## What

[RAI-1430](https://linear.app/makeitrain/issue/RAI-1430)

Preserves requested, broker-accepted, and actually-filled counter-trade quantities through terminal failures, then surfaces that provenance consistently in dashboard history and live updates.

## Why

A failed counter-trade could lose the distinction between what we requested, what the broker accepted, and what actually filled. That made partial fills, overfills, and legacy unknowns look more certain than the persisted evidence supported.

## How

- Persists requested quantity and observed fill evidence on the offchain order aggregate and rejection job.
- Projects nullable provenance for failed outcomes, deriving remaining and excess only when accepted and filled quantities are both known.
- Rebuilds stale offchain order projections safely after interrupted schema upgrades.
- Adds `terminal_outcomes_v2` while keeping v1 stable for rolling deploys, with HTTP and WebSocket fallback and re-probing.
- Validates and renders unknown, partial-fill, and overfill states without inference or clamping.

## Testing

- Added backend coverage for pre-acceptance failure, post-acceptance failure, partial fill, overfill, legacy replay, durable jobs, and interrupted projection upgrades.
- Added dashboard coverage for v1/v2 normalization, REST/WebSocket fallback and re-probing, concurrent requests, runtime validation, and unknown quantity rendering.
- Ran `RUSTC_WRAPPER= nix run .#ci` successfully.

## Screenshots

N/A

## Anything else

Stacked on #1060.

## Summary by CodeRabbit

* **New Features**
  * Trade history now provides accepted, filled, remaining, and excess quantities when available.
  * Dashboard clients use the latest trade protocol and automatically fall back to the legacy protocol when needed.
  * Trade outcome displays identify unknown quantities clearly and show accepted-share information.

* **Bug Fixes**
  * Improved compatibility with older trade-history data and rolling deployments.
  * Preserved accurate fill details for broker-reported failures and overfills.
  * Excluded non-terminal orders from trade history.

### DIFF
--- a/crates/dto/src/trade.rs
+++ b/crates/dto/src/trade.rs
@@ -1,8 +1,9 @@
 //! Trade fill DTOs for completed onchain and offchain trades.
 
 use chrono::{DateTime, Utc};
-use serde::ser::SerializeStruct;
-use serde::{Deserialize, Serialize, Serializer};
+use serde::de::Error as _;
+use serde::ser::{Error as _, SerializeStruct};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use ts_rs::TS;
 
 use st0x_finance::{FractionalShares, NonNegative, Positive, Symbol};
@@ -120,7 +121,7 @@ pub struct InvalidDirectionError {
 }
 
 /// Terminal outcome of a dashboard trade entry.
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, TS)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, TS)]
 #[serde(
     tag = "status",
     rename_all = "snake_case",
@@ -130,16 +131,109 @@ pub enum TradeOutcome {
     Filled,
     Failed {
         error: String,
-        #[ts(type = "string")]
-        filled_shares: NonNegative<FractionalShares>,
-        #[ts(type = "string")]
-        remaining_shares: NonNegative<FractionalShares>,
+        #[ts(type = "string | null")]
+        accepted_shares: Option<Positive<FractionalShares>>,
+        #[ts(type = "string | null")]
+        filled_shares: Option<NonNegative<FractionalShares>>,
+        #[ts(type = "string | null")]
+        remaining_shares: Option<NonNegative<FractionalShares>>,
         /// Shares filled beyond the broker-accepted order quantity. This is
         /// separate from remaining shares so anomalous broker state is never
         /// clamped away.
-        #[ts(type = "string")]
-        excess_shares: NonNegative<FractionalShares>,
+        #[ts(type = "string | null")]
+        excess_shares: Option<NonNegative<FractionalShares>>,
     },
+}
+
+#[derive(Default)]
+enum FieldPresence<T> {
+    #[default]
+    Missing,
+    Present(T),
+}
+
+fn deserialize_present<'de, D, T>(deserializer: D) -> Result<FieldPresence<T>, D::Error>
+where
+    D: Deserializer<'de>,
+    T: Deserialize<'de>,
+{
+    T::deserialize(deserializer).map(FieldPresence::Present)
+}
+
+#[derive(Deserialize)]
+#[serde(
+    tag = "status",
+    rename_all = "snake_case",
+    rename_all_fields = "camelCase"
+)]
+enum TradeOutcomeWire {
+    Filled,
+    Failed {
+        error: String,
+        #[serde(default, deserialize_with = "deserialize_present")]
+        accepted_shares: FieldPresence<Option<Positive<FractionalShares>>>,
+        filled_shares: Option<NonNegative<FractionalShares>>,
+        remaining_shares: Option<NonNegative<FractionalShares>>,
+        excess_shares: Option<NonNegative<FractionalShares>>,
+    },
+}
+
+impl<'de> Deserialize<'de> for TradeOutcome {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        match TradeOutcomeWire::deserialize(deserializer)? {
+            TradeOutcomeWire::Filled => Ok(Self::Filled),
+            TradeOutcomeWire::Failed {
+                error,
+                accepted_shares: FieldPresence::Present(accepted_shares),
+                filled_shares,
+                remaining_shares,
+                excess_shares,
+            } => Ok(Self::Failed {
+                error,
+                accepted_shares,
+                filled_shares,
+                remaining_shares,
+                excess_shares,
+            }),
+            TradeOutcomeWire::Failed {
+                error,
+                accepted_shares: FieldPresence::Missing,
+                filled_shares,
+                remaining_shares,
+                excess_shares,
+            } => {
+                // terminal_outcomes_v1 predates accepted-quantity provenance.
+                // It split an overfill between filledShares and excessShares,
+                // so reconstruct the complete broker fill before discarding the
+                // request-derived remaining/excess values.
+                let filled =
+                    filled_shares.ok_or_else(|| D::Error::missing_field("filledShares"))?;
+                // Presence-only: the request-derived remainder is discarded in
+                // favour of the reconstructed complete fill below.
+                let _ =
+                    remaining_shares.ok_or_else(|| D::Error::missing_field("remainingShares"))?;
+                let excess =
+                    excess_shares.ok_or_else(|| D::Error::missing_field("excessShares"))?;
+                let complete_fill = (filled.inner() + excess.inner()).map_err(D::Error::custom)?;
+                let filled_shares = if complete_fill.inner().is_zero().map_err(D::Error::custom)? {
+                    None
+                } else {
+                    Some(NonNegative::new(complete_fill).map_err(D::Error::custom)?)
+                };
+
+                Ok(Self::Failed {
+                    error,
+                    accepted_shares: None,
+                    filled_shares,
+                    remaining_shares: None,
+                    excess_shares: None,
+                })
+            }
+        }
+    }
 }
 
 /// A completed onchain fill or terminal offchain counter-trade.
@@ -155,10 +249,9 @@ pub struct Trade {
     pub direction: Direction,
     #[ts(type = "string")]
     pub symbol: Symbol,
-    /// Quantity submitted for this counter-trade. Before broker acceptance this
-    /// is the requested quantity; afterward it is the quantity the broker
-    /// accepted. Failed outcomes split this order quantity into filled and
-    /// remaining portions in [`TradeOutcome::Failed`].
+    /// Executed quantity for fills, or requested quantity for a failed
+    /// counter-trade. Failed outcomes carry broker-accepted and fill provenance
+    /// separately when those facts are known.
     #[ts(type = "string")]
     pub shares: Positive<FractionalShares>,
     pub outcome: TradeOutcome,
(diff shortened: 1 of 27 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1062?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1062** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1062?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1062?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1058** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1058?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1053** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1053?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1062).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

0xgleb (commented):
agent misfire

0xgleb (approved):

crates/dto/src/trade.rs
```diff
@@ -130,18 +131,109 @@ pub enum TradeOutcome {
...
+                let filled =
+                    filled_shares.ok_or_else(|| D::Error::missing_field("filledShares"))?;
+                remaining_shares.ok_or_else(|| D::Error::missing_field("remainingShares"))?;
+                let excess =
+                    excess_shares.ok_or_else(|| D::Error::missing_field("excessShares"))?;
+                let complete_fill = (filled.inner() + excess.inner()).map_err(D::Error::custom)?;
+                let filled_shares = if complete_fill.inner().is_zero().map_err(D::Error::custom)? {
+                    None
+                } else {
+                    Some(NonNegative::new(complete_fill).map_err(D::Error::custom)?)
```
rouzwelt, line 224: Make the presence-only check explicit, and confirm no persisted legacy failure ever had null quantities.

Line 215 discards the value while relying on the ? for validation — it reads as a dropped statement. More importantly, this arm now hard-fails deserialization of a legacy Failed payload that carried null for any of the three quantities; since Trade is deserialized from durable job payloads and historical rows, a single such row becomes an unrecoverable load error rather than "unknown".

```
                 let filled =
                     filled_shares.ok_or_else(|| D::Error::missing_field("filledShares"))?;
-                remaining_shares.ok_or_else(|| D::Error::missing_field("remainingShares"))?;
+                // Presence-only: the request-derived remainder is discarded below.
+                let _ = remaining_shares.ok_or_else(|| D::Error::missing_field("remainingShares"))?;
                 let excess =
                     excess_shares.ok_or_else(|| D::Error::missing_field("excessShares"))?;
```

crates/dto/src/trade.rs
```diff
@@ -130,18 +131,109 @@ pub enum TradeOutcome {
     Filled,
     Failed {
         error: String,
-        #[ts(type = "string")]
-        filled_shares: NonNegative<FractionalShares>,
-        #[ts(type = "string")]
-        remaining_shares: NonNegative<FractionalShares>,
+        #[serde(default)]
+        #[ts(type = "string | null")]
```
rouzwelt, line 134: Dead #[serde(default)] attribute on accepted_shares.

TradeOutcome no longer derives Deserialize (deserialization now goes through TradeOutcomeWire, which has its own #[serde(default, deserialize_with = ...)] on the equivalent field at line 174-175). The #[serde(default)] here (134) only affects Deserialize, so on this Serialize-only enum it's vestigial and could mislead a future reader into thinking it governs TradeOutcome's deserialization.

rouzwelt (changes requested):

crates/dto/src/trade.rs
JuaniRios: Fixed in 1829f68a with the explicit `let _ =` and comment. On the legacy concern: the v1 serializer (`TradeOutcome::Failed` before this PR) had all three quantity fields as non-optional `NonNegative`, so every persisted v1 payload carries all three as decimal strings -- null or missing quantities cannot exist in durable rows. Rows older than v1 have no Failed outcomes at all.

crates/dto/src/trade.rs
JuaniRios: Fixed in 1829f68a.

crates/dto/src/trade.rs
```diff
@@ -130,18 +131,111 @@ pub enum TradeOutcome {
...
+                accepted_shares: FieldPresence::Present(accepted_shares),
+                filled_shares,
+                remaining_shares,
+                excess_shares,
+            } => Ok(Self::Failed {
+                error,
+                accepted_shares,
+                filled_shares,
+                remaining_shares,
+                excess_shares,
```
rouzwelt, line 199: The v2 branch does not validate the derived-quantity invariants that the TypeScript parser enforces.

The FieldPresence::Present branch passes accepted_shares, filled_shares, remaining_shares, and excess_shares through without any cross-field check. The dashboard parser in dashboard/src/lib/trade-payload.ts (lines 192-217) rejects the same payload shapes:

remaining_shares or excess_shares non-null while accepted_shares or filled_shares is null.
remaining_shares != max(accepted - filled, 0).
excess_shares != max(filled - accepted, 0).
So the two ends of the same contract disagree on what a valid v2 failed outcome is. The Rust boundary accepts a Trade that the client then rejects with a visible payload error. This matters on the persistence path: trade_roundtrips_through_persistent_job_payload shows Trade is deserialized from a durable job payload, so a corrupted or hand-edited row deserializes successfully here and only fails later in the browser.

Encode the invariant once in the Rust type so it cannot be constructed or deserialized in an inconsistent state, matching how Positive<FractionalShares> already encodes the positive-total invariant. A private constructor on the Failed variant, or a validation step in this branch, both work.

rouzwelt (changes requested):

crates/dto/src/trade.rs
JuaniRios: Fixed in 883450ef, on the child PR #1063.

Added `validate_derived_quantities`, called from both v2 branches, enforcing the same rules as `trade-payload.ts`: remaining and excess must be null when acceptance or fill is unknown, and otherwise must equal `max(accepted - filled, 0)` and `max(filled - accepted, 0)`. Five tests cover both variants.

It landed on #1063 rather than here because the `Cancelled` variant only exists there, so one implementation covers both variants instead of splitting the same check across two branches.

rouzwelt (approved):

crates/dto/src/trade.rs
```diff
@@ -130,18 +131,111 @@ pub enum TradeOutcome {
...
+            }),
```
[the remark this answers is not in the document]
JuaniRios: Fixed in d07de706 on the child PR #1063.

`validate_derived_quantities` is called from both v2 branches and enforces the same rules as `trade-payload.ts`: remaining and excess must be null when acceptance or fill is unknown, and otherwise must equal `max(accepted - filled, 0)` and `max(filled - accepted, 0)`. Five tests cover both the failed and cancelled variants.

It landed there rather than here because the `Cancelled` variant only exists on that branch, so one implementation covers both variants instead of splitting the same check across two branches.

src/offchain/order/poll_status.rs
```diff
@@ -205,10 +211,11 @@ impl PollOrderStatus {
...
             Failed {
                 error_reason,
+                shares_filled,
                 failed_at,
                 ..
             } => {
-                self.enqueue_rejection(ctx, symbol, error_reason, Some(failed_at))
+                self.enqueue_rejection(ctx, symbol, error_reason, shares_filled, Some(failed_at))
                     .await
             }
```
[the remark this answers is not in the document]
JuaniRios: Fixed in d07de706 on the child PR #1063, which adds `validate_broker_terminal_fill` and calls it at the top of `dispatch_for_broker_state`, before any arm runs.

That covers more than this finding asks for: it rejects a negative fill on both terminal `Failed` and terminal `Cancelled`, rather than only the `Failed` catch-all, so no corrupt quantity reaches `enqueue_rejection` or the event stream on either path.

I first added a narrower guard arm here, then dropped it: once restacked it sits behind the pre-dispatch validator and is unreachable, so it would have been dead code in the merged result.

src/api.rs
```diff
@@ -144,7 +144,7 @@ struct StuckTransferInfo {
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
 struct TradeResponse {
-    entries: Vec<Trade>,
+    entries: Vec<serde_json::Value>,
```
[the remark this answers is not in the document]
JuaniRios: Agreed on the substance, deferred rather than done here: RAI-1603.

The erased `Vec<serde_json::Value>` does block `ts-rs` from deriving the client type and does allocate an intermediate `Value` per trade per request. Reshaping the response path is a wider change than the counter-trade outcome work this PR is scoped to, and the wire output is unaffected either way, so it is tracked separately.

src/conductor.rs
```diff
@@ -252,6 +254,45 @@ where
...
+    if has_stale_rows != 0 {
+        info!(
+            "Offchain order projection rows predate quantity provenance; \
+             rebuilding the full projection"
+        );
+        projection.rebuild_all().await?;
+    }
+
+    Ok(())
+}
```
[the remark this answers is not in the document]
JuaniRios: Agreed, deferred rather than done here: RAI-1604.

The scan is correct today, and the cost only becomes real as `offchain_order_view` grows, so it is a follow-up rather than a fix for this PR. The issue records the constraint you identified: whatever marker we persist, an interrupted rebuild must still be detected and completed on the next start.

JuaniRios:
Answering the out-of-diff findings here, since they have no threads to reply in.

**`api.rs:580` reuse `TradeProtocol::includes_trade` (Major).** Correct for this PR. Fixed in d07de706 on #1063, which replaces the inline `matches!` with `all_trades.retain(|trade| query.trade_protocol.includes_trade(trade))`. The same duplication in `includes_statement` is collapsed there too, so the protocol type is the single source of the inclusion rule.

**`trade_loader.rs:101` cancelled orders dropped from history (Trivial).** Confirmed dropped, and deliberately so on this branch: cancelled outcomes are v2-only and v2 does not exist yet here, so `TradeConversionError::Cancelled` is in the skip list on purpose. #1063 removes that variant entirely and renders cancellations as a real outcome, with coverage for a cancelled order that retains a partial fill.

**`trade-payload.ts:20` return `Result` instead of throwing (Major).** Keeping the throwing shape. Both exported entry points are called from already-guarded paths, so no consumer observes a raw exception today. Converting only the four entry points leaves every internal helper throwing anyway, so the gain is a type signature rather than real safety. Worth revisiting if a third consumer appears.

**`event.rs:628` typed identifier in `MissingRecord` (Major).** Not taking this one. `trade_id` is a `String` because `Trade.id` is a `String` everywhere in the DTO: onchain ids are `"tx_hash:log_index"`, offchain ids are the aggregate id. A newtype in this single error variant would not match the type the value actually has at every other site, so it adds an inconsistency rather than removing a class of mistake. If we introduce a `TradeId` newtype it should land on `Trade.id` first, and this variant follows from that.

The remaining inline findings are fixed in 976d7013 on this branch.

PR #1075, opened 2026-07-22
fix: bound PollOrderStatus poll jobs to one per open order

JuaniRios:
## What

- Stop `PollOrderStatus` poll jobs from multiplying without bound for one open offchain order. This is one of three sub-issues of the [RAI-1492](https://linear.app/makeitrain/issue/RAI-1492) production incident. See [RAI-1493](https://linear.app/makeitrain/issue/RAI-1493).
- Add one shared guard, `reconcile_and_check_live_poll_job`, that every push site calls before it enqueues a poll job for an order.
- Route all five push sites through that guard: the periodic `recover_submitted_offchain_orders` sweep, `dispatch_post_place_state` (post-place and per-fill), `recover_claimed_offchain_order`, `recover_pending_poll_status`, and `route_placement_outcome`.
- Collapse pre-existing duplicate `Pending` poll rows for one order, so only the best survivor stays and the rest go `Done`.

## Why

- On 2026-07-22 the `PollOrderStatus` worker ran 39,411 executions in 43 minutes for only 2 orders. That tripped the Alpaca rate limit. It burned the job retry budget. It latched the single-concurrency worker idle. All hedging stopped, silently, for 3 hours and 25 minutes.
- The cause was not one code path. Every `PollOrderStatus` execution self-reschedules. Several push sites also re-pushed a job for an already-open order with no dedup. So each recovery tick, and each new fill, forked a new self-perpetuating chain.
- An order stays open for hours when its hedge is an extended-hours limit order on a thin-book equity that does not fill. That is the state that triggers the runaway.

## How

- The guard runs one atomic `UPDATE`. It keeps the single best `Pending` survivor row for an order and marks the rest `Done`. The survivor is picked with apalis's own dispatch order (`priority DESC, run_at ASC, id ASC`, due rows first), so the row kept is the row apalis would run next. One statement, so no read-then-write race can leave the order with zero live rows.
- The guard then checks if a `Pending` row, a fresh in-flight `Queued`/`Running` row, or a fresh retryable `Failed` row still exists for the order. "Fresh" means locked or failed within a staleness bound. This prevents a duplicate self-reschedule chain, while still allowing at most one fresh in-flight row next to the `Pending` successor.
- Retryable `Failed` counts as live only while fresh. A stale one does not, so a row stuck behind a latched worker still gets re-pushed by the next sweep. That keeps this fix decoupled from the circuit-breaker fix in RAI-1495.
- All five push sites now go through one helper, `push_poll_job_if_absent`, so the check-then-push shape lives in one place.

## Testing

- Some tests drive the real apalis fetch and ack paths, not hand-seeded rows, where the apalis retry contract matters. This caught an earlier wrong assumption that a retryable `Failed` row is inert.
- The atomicity test holds a `BEGIN IMMEDIATE` write lock, commits a successor row while the collapse is blocked, and asserts the order never drops to zero live rows.
- Added guard coverage at every push site, including the per-fill burst case and the `PendingExecution` retry path.
- Ran the scoped suite (`offchain::order`, `position_check`, `conductor`, `hedge`, `trade_accountant`), `cargo clippy --all-targets --all-features`, and `nix run .#ci` locally. All green.

## Anything else

- The observed incident volume is larger than the ~60s recovery sweep alone can explain. The fill-driven push sites are the bigger contributor, so gating them is the core of the fix, not just the sweep.
- Two follow-ups are out of scope here and logged. The other `json_extract(job, ...)` dedup guards in the rebalancing and equity-recovery jobs lack the `json_valid` hardening. A full move to one recurring sweep job, instead of per-order self-reschedule chains, is the larger structural fix.
- No migration. The guard is application-level. A DB `UNIQUE` index was considered and rejected, because it would crash-loop `requeue_trading_orphans` at boot and break apalis's own heartbeat sweep. The reasoning is written up in `docs/conductor.md`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate order-status polling jobs when orders are reconciled, retried, recovered, hedged, or checked repeatedly.
  - Improved recovery of submitted and partially filled orders by maintaining a single active polling chain per order.
  - Added safeguards for stale or malformed polling-job records to keep recovery reliable.

- **Documentation**
  - Expanded guidance on polling-job payloads, deduplication behavior, recovery rules, and handling existing duplicates.

- **Tests**
  - Added coverage for repeated checks, recovery scenarios, concurrency, and duplicate-job prevention.

### DIFF
--- a/src/offchain/order/poll_status.rs
+++ b/src/offchain/order/poll_status.rs
@@ -11,6 +11,7 @@ use std::time::Duration;
 
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
+use tokio::sync::Mutex;
 use tracing::{debug, error, info, warn};
 
 use st0x_event_sorcery::{Projection, Store};
@@ -32,6 +33,25 @@ use crate::offchain::order::{
 use crate::position::Position;
 
 pub(crate) type PollOrderStatusJobQueue = JobQueue<PollOrderStatus>;
+
+/// Serializes the guard's reconcile/check and push as one process-local
+/// critical section. The service owns one apalis queue for this SQLite
+/// database, so this closes every in-process check-then-push race without a
+/// database uniqueness constraint that would reject the legitimate overlap
+/// between a `Running` poll and its delayed `Pending` successor.
+///
+/// Cross-process races are not covered: the CLI binary can run against the
+/// same database while the server is live, and this lock does not span
+/// processes. Such a race forks at most one duplicate `Pending` row, which
+/// the collapse in [`reconcile_live_poll_jobs`] converges back to one on the
+/// next guarded push -- that self-healing, not this lock, is the safety story
+/// for the cross-process case.
+///
+/// The lock is global, so a fill burst across many symbols funnels every
+/// guard+push (three SQL round-trips) through it single-file. Irrelevant at
+/// current push frequency; if it ever shows up in latency, a per-order keyed
+/// lock is the natural refinement.
+static POLL_JOB_PUSH_LOCK: Mutex<()> = Mutex::const_new(());
 
 /// Dependencies [`PollOrderStatus`] needs to query the broker and route the
 /// result. The `Filled` and `Failed` terminal transitions are deferred to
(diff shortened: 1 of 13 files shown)


JuaniRios:
* **#1095** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1095?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1092** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1092?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1075** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1075?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1075?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1075).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor (approved):

JuaniRios:
@coderabbitai full review

JuaniRios:
@coderabbitai full review

JuaniRios:
@coderabbitai full review

JuaniRios:
@coderabbitai full review

src/offchain/order/poll_status.rs
```diff
@@ -33,6 +34,13 @@ use crate::position::Position;
 
 pub(crate) type PollOrderStatusJobQueue = JobQueue<PollOrderStatus>;
 
+/// Serializes the guard's reconcile/check and push as one process-local
+/// critical section. The service owns one apalis queue for this SQLite
+/// database, so this closes every in-process check-then-push race without a
+/// database uniqueness constraint that would reject the legitimate overlap
+/// between a `Running` poll and its delayed `Pending` successor.
+static POLL_JOB_PUSH_LOCK: Mutex<()> = Mutex::const_new(());
+
```
rouzwelt, line 55: the doc claims the process-local lock "closes every in-process check-then-push race", which is true, but the CLI binary can run against the same SQLite database while the server is live, and the lock doesn't span processes. The consequence is bounded — a cross-process race forks at most one duplicate Pending, which the next guarded push's collapse converges back to one — but that self-healing argument is the actual safety story for the cross-process case and it isn't written down here. One sentence on the static's doc ("cross-process races are not covered; the collapse in reconcile_live_poll_jobs converges any duplicate they fork on the next guarded push") would stop a reader from concluding the lock is load-bearing for more than it is.

src/offchain/order/poll_status.rs
```diff
@@ -812,20 +820,315 @@ where
...
 /// the bot crashes between those two writes, the order would otherwise sit in
-/// `Submitted` forever waiting for a poll that never comes.
+/// `Submitted` forever waiting for a poll that never comes. This periodic
+/// variant logs and counts a per-order guard/push failure, then continues
+/// arming the remaining orders; the startup variant fails closed instead.
 pub(crate) async fn recover_submitted_offchain_orders(
     offchain_order_projection: &Projection<OffchainOrder>,
-    poll_status_queue: &mut PollOrderStatusJobQueue,
+    poll_status_queue: &PollOrderStatusJobQueue,
+    executor_type: SupportedExecutor,
```
rouzwelt, line 1246: the global mutex serializes the guard+push across all orders, holding it across three SQL round-trips. At current push frequency that's irrelevant, but a fill burst across many symbols funnels through it single-file; if it ever shows up in latency, a per-order keyed lock is the natural refinement. No change needed now — just don't let the static quietly become a bottleneck assumption.

src/offchain/order/poll_status.rs
JuaniRios: Good catch, the doc did overclaim. Added the cross-process caveat: the CLI can share the DB with a live server, the lock does not span processes, and the collapse in `reconcile_live_poll_jobs` is what converges any duplicate such a race forks. Fixed in 23087d39.

src/offchain/order/poll_status.rs
JuaniRios: Agreed. Recorded it on the static's doc so the assumption stays visible: the lock is global, a fill burst funnels through it single-file, and a per-order keyed lock is the refinement if it ever shows up in latency. Added in 23087d39.

PR #1083, opened 2026-07-22
ci: run CI for fork pull requests behind an ok-to-test label

JuaniRios:
## What

- New workflow `.github/workflows/external-pr-ci.yaml`. It runs CI for a pull request from a fork after a maintainer adds the `ok-to-test` label.
- `ci.yaml` now ignores pushes to `external-pr/**`, so one commit can never start CI twice.

## Why

- A pull request from a fork gets no CI. `ci.yaml` runs on `push`, and a push to a fork starts workflows in the fork, not here.
- The required checks `backend`, `dashboard`, and `hooks` stay at "Expected, waiting for status to be reported". The pull request can never merge.
- [#1057](https://github.com/ST0x-Technology/st0x.liquidity/pull/1057) is blocked on this now. It comes from `agryaznov/st0x.liquidity` and has zero workflow runs.

## How

- The `dispatch` job copies the pull request head commit to a branch named `external-pr/<number>` in this repository. Git keeps the original commit ID.
- It then starts `ci.yaml` on that branch with `workflow_dispatch`. A `GITHUB_TOKEN` can still start that event, so this needs no personal access token.
- GitHub attaches a check to a commit, not to a branch. The real `backend`, `dashboard`, and `hooks` results land on the pull request head commit, and branch protection accepts them.
- The job first checks that the labeler has write access. It also stops if the head commit moved after the label event.
- The `revoke` job removes the label on every new push. Each new revision needs a new review.
- The `cleanup` job deletes the mirror branch when the pull request closes, or when someone removes the label.

## Testing

No automated test. A GitHub Actions trigger cannot run under `nextest`, and the `pull_request_target` path needs a real pull request from a fork. I checked the YAML and ran the repository pre-commit hooks on both files locally. The first real proof is a live run on [#1057](https://github.com/ST0x-Technology/st0x.liquidity/pull/1057) after this merges.

## Screenshots

None, this is CI configuration.

## Anything else

- **Setup step: someone must create the `ok-to-test` label.** The repository does not have it yet, so the workflow cannot fire until then.
- **Security.** The dispatched run executes fork code with this repository's secrets. It also runs the fork's copy of `ci.yaml`, because `workflow_dispatch` reads the workflow file from the branch it runs on. The label is the only gate. Read the whole diff, `.github/` included, before you add the label.
- I rejected a `pull_request` trigger. A fork run gets no secrets, so the Graphite optimizer and the Cachix cache lose their tokens. It also adds a second `backend`, `dashboard`, and `hooks` check to our own pull requests, and a skipped check can count as a pass.
- I rejected a `pull_request_target` trigger that runs the tests itself. That event reports against the base branch commit, so the required checks stay unsatisfied.
- The `main protect` ruleset covers the default branch only, so `external-pr/**` needs no ruleset change.

## Summary by CodeRabbit

* **New Features**
  * Added a controlled “External PR CI” workflow that runs fork pull request checks only after an approval label is applied.
  * CI mirrors eligible fork commits, then dispatches the standard CI workflow against the mirrored reference.
  * Added automated cancellation and cleanup for external PR mirror runs when PR state changes or becomes unlabeled.
* **Bug Fixes**
  * Prevented CI optimizations from incorrectly being interpreted as a successful status for protected branches.
  * Reduced workflow permissions for safer external pull request execution.
* **Tests**
  * Added automated coverage for approval gating, dispatch behavior, cancellation/cleanup, and failure scenarios.

### DIFF
--- a/.github/workflows/external-pr-ci.yaml
+++ b/.github/workflows/external-pr-ci.yaml
@@ -1,0 +1,416 @@
+name: External PR CI
+
+# Pull requests from forks get no CI. `ci.yaml` triggers on `push`, and a push
+# to a fork starts workflows in the fork, never here. The required `backend`,
+# `dashboard`, and `hooks` contexts therefore stay "Expected -- waiting for
+# status to be reported" forever, and the pull request can never merge.
+#
+# This workflow closes that gap. When a maintainer adds the `ok-to-test` label,
+# it mirrors the pull request head commit to an `external-pr/<number>` branch in
+# this repository, then dispatches `ci.yaml` against that branch. Git keeps the
+# original commit SHA when the commit moves between repositories, and GitHub
+# attaches check runs to a commit rather than to a branch
+# (https://docs.github.com/en/rest/checks/runs). The real `backend`,
+# `dashboard`, and `hooks` results therefore land on the pull request head
+# commit and satisfy branch protection.
+#
+# TRUST BOUNDARY. The dispatched run executes fork code, including build
+# scripts, flake.nix, nix/, scripts/, build.rs, and tests. External mirror refs
+# therefore receive only a read-only GITHUB_TOKEN: ci.yaml has no OIDC grant,
+# skips the Graphite optimizer, and configures Cachix without its auth token or
+# push capability on external-pr/** refs. A maintainer must still read the
+# whole diff before labeling because untrusted code can consume runner time and
+# make network requests.
+#
+# `workflow_dispatch` loads ci.yaml from the mirrored ref, so the fork also
+# supplies the workflow definition. The `.github/` integrity gate below
+# requires the fork's entire `.github/` tree to match the CURRENT base tip,
+# not merely the fork's merge base. A stale fork therefore cannot retain an
+# older workflow after master removes a permission, action, or credential.
+# Every new push removes the label, so each revision needs a new review.
+#
+# Rejected alternatives:
+#   - A `pull_request` trigger. It creates a second set of `backend`,
+#     `dashboard`, and `hooks` check runs for internal pull requests, and a
+#     skipped check can satisfy branch protection.
+#   - A `pull_request_target` trigger that runs the tests itself, publishing
+#     `backend`/`dashboard`/`hooks` check runs on the pull request head SHA via
+#     `POST /repos/{owner}/{repo}/check-runs` with `checks: write` (the Checks
+#     API can target any commit, not only `github.sha`). This CAN satisfy the
+#     required checks, and would remove the fork-controlled-workflow-file
+#     problem entirely, since the job logic would then live in the trusted base
+#     branch's workflow instead of the fork's copy. It was not used here
+#     because it means re-implementing, and keeping in sync, the entire
+#     `backend`/`dashboard`/`hooks` job bodies as a second copy, instead of
+#     dispatching the existing `ci.yaml` unchanged.
+#
+# This workflow itself never runs fork code. It only moves refs and calls the
+# API, so `pull_request_target` is safe here.
+
+on:
+  pull_request_target:
+    types: [labeled, unlabeled, synchronize, closed]
+    # Only `master` is ever mergeable here (branch protection covers it, per
+    # AGENTS.md). Without this filter, a fork pull request opened against ANY
+    # other branch in this repository (a stale release branch, a maintainer's
+    # own long-lived branch, etc.) would also get mirrored and dispatched,
+    # which is pointless (it cannot merge) and needlessly widens which
+    # branches this workflow reacts to.
+    branches: [master]
+
+# `dispatch`, `revoke`, and `cleanup` are three independent workflow runs (one
+# per triggering event) that all touch the same external-pr/<n> mirror branch
+# and the same ok-to-test label. Without this, a revoke/cleanup run and a
+# dispatch run for the same pull request can interleave: revoke's
+# cancel-and-delete step can run and find nothing yet (the mirror branch does
+# not exist, no CI is dispatched yet), then dispatch pushes the branch and
+# starts CI anyway, moments later, unrevoked. Group by pull request number so
+# events for the SAME pull request always run one at a time, in the order
+# GitHub delivered them; unrelated pull requests are unaffected.
+#
+# `cancel-in-progress: false`: a queued run must still execute once its turn
+# comes (e.g. a queued revoke must still cancel/delete after an in-progress
+# dispatch finishes), so cancelling a queued run here would silently drop a
+# revocation instead of just delaying it.
+#
+# `queue: max` retains up to 100 pending events instead of replacing an older
+# queued revocation with a newer event. Each dispatch and cleanup also reads
+# the live pull request state before changing anything, so a stale queued event
+# cannot override a later approval decision if GitHub starts the runs out of
+# event-delivery order.
+concurrency:
+  group: external-pr-ci-${{ github.event.pull_request.number }}
+  cancel-in-progress: false
+  queue: max
+
+permissions: {}
+
+# The label name `ok-to-test` is repeated in every gate below. A job-level `if`
+# cannot read workflow `env`, so a shared variable would cover only half the
+# uses and hide the other half.
+
+jobs:
+  dispatch:
+    name: mirror and dispatch
+    # `github.event.label` exists only on labeled and unlabeled events, so the
+    # action check must come first. `state == 'open'` keeps a label added to a
+    # closed or merged pull request from mirroring and dispatching CI for code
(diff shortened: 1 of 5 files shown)


JuaniRios:
* **#1083** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1083?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1083?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1083).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

.github/workflows/external-pr-ci.yaml
```diff
@@ -0,0 +1,523 @@
...
+              exit 1
+              ;;
+          esac
+
+      # `pull_request_target` checks out the base branch, never fork code. The
+      # checkout is here for the git credentials and the origin remote, and
+      # `fetch-depth: 0` (full history, not a shallow clone) is what lets the
+      # `.github/` integrity gate below compute a real `git merge-base` against
+      # the pull request head, instead of trusting a capped, paginated API.
+      - uses: actions/checkout@v4
```
0xgleb: minor: Pin this checkout action to a full commit SHA. This `pull_request_target` job gives it a token with repository-write permissions, while `actions/checkout@v4` is a mutable tag. GitHub documents a full-length SHA as the only immutable action reference, so use the audited v4 commit here.

.github/workflows/external-pr-ci.yaml
```diff
@@ -0,0 +1,523 @@
...
+          set -euo pipefail
+          # `git fetch`, `git merge-base`, and `git diff` here are deliberately
+          # NOT wrapped to tolerate failure: if any of them errors, the script
+          # aborts under `set -e` before the push/dispatch steps ever run, so
+          # this fails closed rather than treating "could not compute" as "no
+          # change".
+          git fetch --no-tags origin "$BASE_REF"
+          base_sha=$(git rev-parse FETCH_HEAD)
+
+          merge_base=$(git merge-base "$base_sha" "$fetched_sha")
```
0xgleb: should fix: This gate proves the contributor did not change `.github/` relative to the old merge base, but it does not prove the dispatched workflow is the currently trusted one. Because `workflow_dispatch` loads `ci.yaml` from the mirror ref, a stale fork can run an older workflow after `master` removes a vulnerable action, permission, or secret. Require the mirrored `.github` tree to match the current base tip, or execute trusted base-branch job logic with the fork SHA as input.

.github/workflows/external-pr-ci.yaml
```diff
@@ -0,0 +1,523 @@
...
+          fi
+
+          echo "Re-confirmed: #$PR_NUMBER still carries ok-to-test at $fetched_sha."
+
+      # `workflow_dispatch` is one of the two events that a GITHUB_TOKEN can
+      # still start
+      # (https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows),
+      # so this needs no personal access token. The push above deliberately
+      # starts nothing on its own.
+      - name: Dispatch CI against the mirror branch
```
0xgleb: should fix: The mirrored run executes fork-controlled `flake.nix`, build scripts, tests, and dependency hooks while `ci.yaml` grants OIDC and configures authenticated cache access. A maintainer label should authorize testing, not let fork code mint this repository's OIDC identity or reuse write credentials. Run external refs with `contents: read`, no OIDC, and no write secrets, or structurally exclude every privileged setup from `external-pr/**`.

.github/workflows/external-pr-ci.yaml
```diff
@@ -0,0 +1,523 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          set -euo pipefail
+          if ! gh workflow run ci.yaml --ref "external-pr/$PR_NUMBER"; then
```
0xgleb: should fix: If dispatch fails after the mirror push, this job leaves both the branch and approval label behind. No later cleanup is guaranteed unless another PR event occurs. Move live validation before the push where possible and add failure cleanup that handles an ambiguously accepted dispatch, removes the label, and deletes the mirror.

.github/workflows/external-pr-ci.yaml
```diff
@@ -0,0 +1,523 @@
...
+          # loop exists to avoid. A failed attempt here counts as "found
+          # nothing yet" and the loop keeps retrying.
+          queued_and_running=""
+          list_failed=false
+          for attempt in 1 2 3 4 5; do
+            if ! queued_and_running=$(gh api \
+              -H "Accept: application/vnd.github+json" \
+              -H "X-GitHub-Api-Version: 2022-11-28" \
+              "/repos/$GITHUB_REPOSITORY/actions/workflows/ci.yaml/runs?branch=external-pr/$PR_NUMBER" \
+              --jq '.workflow_runs[] | select(.status == "queued" or .status == "in_progress") | .id')
```
0xgleb: should fix: Revocation does not prove every secret-bearing run has stopped. GitHub also reports `requested`, `waiting`, and `pending` runs, which this filter omits, and the cancel endpoint returns 202 when cancellation is accepted rather than completed. Cancel every non-completed run and poll until all are terminal before deleting the branch and reporting success.

0xgleb (changes requested):

.github/workflows/external-pr-ci.yaml
JuaniRios: fixed. external refs now run with `contents: read`, no OIDC permission, no Cachix auth token, and `skipPush: true`.

.github/workflows/external-pr-ci.yaml
JuaniRios: fixed. live approval is checked before the mirror push and again before dispatch. any failure after the push attempt removes the label and runs cleanup.

.github/workflows/external-pr-ci.yaml
JuaniRios: fixed. the gate now compares `.github/` against the freshly fetched current base tip. stale forks have to rebase before CI can run.

.github/workflows/external-pr-ci.yaml
JuaniRios: fixed. cleanup now paginates all runs, force-cancels every non-completed run, polls until terminal, and only then deletes the mirror. tests cover late runs, stuck runs, and API failures.

.github/workflows/external-pr-ci.yaml
JuaniRios: good catch. pinned every checkout in this `pull_request_target` workflow to the full v4 commit SHA.

.github/workflows/external-pr-ci.yaml
```diff
@@ -0,0 +1,420 @@
...
+# that a concurrency group holds at most one queued run at a time -- if two
+# events for the same pull request queue up while a third is in progress,
+# GitHub keeps only the most recently queued one and drops the older pending
+# run, `cancel-in-progress: false` notwithstanding. A dropped run is simply
+# never executed (not retried, not logged here). If a mirror branch or a
+# running `ci.yaml` job is ever found for a pull request after its label was
+# removed or the pull request closed, this is the mechanism to suspect first:
+# check `gh api .../actions/workflows/ci.yaml/runs?branch=external-pr/<n>` and
+# whether `external-pr/<n>` still exists, and clean up by hand.
+concurrency:
```
0xgleb, line 81: should fix: Keeping serialization is right, but the default concurrency queue retains only one pending run, so a later event can replace a queued revocation. For example, while dispatch is active, `unlabeled` can queue cleanup and a subsequent `synchronize` with the label already absent can replace it; that run skips `revoke`, leaving fork CI active. Use `queue: max` and reconcile the live approval state before each dispatch or cleanup, rather than disabling concurrency.

0xgleb (changes requested):

.github/workflows/external-pr-ci.yaml
JuaniRios: good catch, fixed in d305b2e1b8bf27f3e2aebbb4101f8b820135a2d3. Added `queue: max`. Dispatch already rechecked the live head and label; cleanup now does too and skips if the PR is open and reapproved. Added regression coverage for reapproval, closed PRs, and API failure.

0xgleb (approved):

PR #1047, opened 2026-07-16
docs: record fresh mark quoting boundary

JuaniRios:
## Summary
- records ADR 0015 for RAI-1405's fresh-mark quote gate request
- documents that this repo does not own Raindex quote lifecycle/order pricing
- points the mitigation at the order-management layer or order expression instead of fill accounting/hedging

## Verification
- doc-only change; pre-commit markdown formatting passed during `gt modify`

PR #1044, opened 2026-07-16
fix: reprice stale extended-hours hedges

JuaniRios:
## What

- Fixes [RAI-1403](https://linear.app/makeitrain/issue/RAI-1403/reprice-unfilled-extended-hours-hedge-limits-to-marketable-after-a).
- Adds the required `[broker].extended_hours_reprice_timeout_secs` setting and configures it to `300` seconds.
- Cancels stale `Submitted` and `PartiallyFilled` extended-hours hedges, so the next scan can place a fresh marketable limit.

## Why

- An extended-hours limit can remain unfilled for hours while its reference price moves.
- The regular-open replacement only fixes orders that survive until regular hours.
- A five-minute timeout is conservative against the observed extended-hours median fill time of about 70 seconds.

## How

- `CheckPositions` scans enabled symbols on every tick and compares each live order's `placed_at` timestamp with the timeout.
- `CancellationReason::ExtendedHoursRepriceTimeout` keeps timeout repricing distinct from regular-open replacement.
- The existing cancellation path reconciles partial fills before it releases the position for another hedge attempt.
- Alpaca configs must set a non-zero timeout. The deployed configs use `300` seconds.

## Testing

- `nix develop .#ci-backend -c cargo check -p st0x-config -p st0x-hedge`
- `nix develop .#ci-backend -c cargo nextest run -p st0x-config alpaca_broker_api_requires_extended_hours_reprice_timeout_config alpaca_broker_api_requires_counter_trade_slippage_config`
- `nix develop .#ci-backend -c cargo nextest run -p st0x-hedge position_check`
- A focused reviewer pass found no issues.

## Screenshots

N/A (backend).

## Anything else

- This PR only handles stale limit repricing. Close flattening and bot gas PnL are separate PRs in this stack.

## Summary by CodeRabbit

* **New Features**
  * Added configurable `extended_hours_reprice_timeout_secs` (300s) for extended-hours order repricing.
  * Extended-hours limit orders that exceed the configured timeout are eligible for cancellation so refreshed orders can be placed.
* **Bug Fixes**
  * Improved extended-hours timeout/cancellation behavior to cancel only stale eligible orders while keeping fresh ones.
  * Cancellation is now aligned to the extended-hours market session behavior.
* **Tests**
  * Added/updated coverage for timeout configuration validation and the stale-vs-fresh cancellation outcomes.

PR #1093, opened 2026-07-23
feat: record bot gas receipt costs from production tx paths

JuaniRios:
## What

[RAI-1520](https://linear.app/makeitrain/issue/RAI-1520/record-bot-gas-receipt-costs-from-production-transaction-paths)

- Records the gas cost of every bot signed transaction into the `BotGasReceiptCost` aggregate that PR #1049 added.
- Adds `src/bot_gas/valuation.rs` (ETH/USD from Pyth), `src/bot_gas/job.rs` (the `RecordBotGasReceiptCost` job, queue, and worker), and `src/bot_gas/redrive.rs` (the shared failure-handling mechanism).
- Threads a `BotGasReceiptCostEnqueuer` into every production path that signs a transaction: equity mint (`src/rebalancing/equity/mod.rs`), equity redemption (`src/equity_redemption.rs`), both recovery aggregates, and the USDC manager (`src/rebalancing/usdc/manager.rs`).
- Adds the `[bot_gas_valuation]` config section. It is required in rebalancing mode only, like `[tokenization]`.

## Why

- PR #1049 built the event model and the `/pnl` reader, but nothing in production ever wrote those events. Bot gas stayed `not_ingested` in every PnL report, so the numbers left out a real cost.
- Every bot signed transaction spends gas: vault deposits and withdrawals, wrap and unwrap, CCTP burns and mints, and the USDC transfer to the Alpaca deposit address. Each one now records its confirmed receipt cost with a price you can reproduce later.

## How

- The worker refetches the receipt, checks the bot paid for it, skips reverted receipts with a warning, and values `gas_used * effective_gas_price` with the Pyth ETH/USD feed on Base. ADR 0017 explains the block pinning: a Base receipt is pinned to its own block by block HASH, so the price and the timestamp share one reorg-safe anchor. That needs an RPC that supports EIP-1898, which our self-hosted Base node does. An Ethereum receipt uses the latest Base block, so it is not reproducible across two attempts.
- Idempotency compares the immutable receipt facts (chain, transaction hash, gas used, effective gas price, and category), not the derived USD cost. Persistence rounds to 8 decimals, so a retry that recomputed the cost at full precision would never match and would raise a false conflict.
- Recording must never break trading, so `src/bot_gas/redrive.rs` holds one shared mechanism for it. Each job's error type implements `BotGasFailureClassifier` with an exhaustive `match`, so a new error variant does not compile until someone classifies it. A bookkeeping failure then redrives the job with a delay instead of burning the retry budget and tripping the fail-stop circuit breaker.
- RPC reads that are not yet visible (the receipt, or its block) redrive on the same path, bounded to 20 attempts at 30 seconds, so about 10 minutes. After that the job dead-letters and logs at `error!`.
- Enqueue order depends on the step before it. After an idempotent confirm, the enqueue runs before the state advancing command. After the non idempotent token send in `EquityRedemption::SendTokens`, it runs after the event and only logs on failure, because a retry there would send the tokens twice.
- `BotGasReceiptCost` is now part of the `verify-migrations` replay checks that gate a deploy.

## Testing

- Unit tests for each enqueue site, for the valuation math, for the idempotency and conflict paths, for the bounded redrive and its expiry, and for the reverted-receipt skip.
- One test pins the real Pyth response shape (real feed id, real field widths, and expo) so the assumed contract cannot drift unnoticed. A config test pins the deployed `pyth_contract` address.
- `tests/e2e/rebalancing.rs` runs the real path against anvil with a deployed `MockPyth`, for both the equity and the USDC flows. It asserts the recorded chain, the transaction hash, a positive USD cost, and the exact set of operation categories.
- CI runs the full suite on this commit.

## Anything else

- Config: `[bot_gas_valuation]` is added to `example.config.toml`, `e2e/config.toml`, and the prod and staging tomls. The values are public (the Pyth contract address and the ETH/USD feed id), so the encrypted secrets file does not change.
- Known gaps, all listed in SPEC.md and filed under [RAI-1531](https://linear.app/makeitrain/issue/RAI-1531): the recorded cost is execution gas only and leaves out the Base L1 data availability fee; gas on a reverted but mined transaction is not recorded; and two paths swallow the cost fact rather than risk the money-moving operation. Those two are `EquityRedemption::SendTokens`, and the recovery aggregates' `DispatchToMint`/`DispatchToRedemption` handoff, which folds the failure into the normal `RecoveryFailed` event because `WrappedEquityRecoveryJob` has no resume arm for the state a redrive would land on.
- This PR went through several review passes. Two findings are worth calling out for a reviewer: the idempotency bug above, where the original tests passed only because they used numbers that round trip exactly, and the risk that a failed bookkeeping write could trip the circuit breaker and stop trading.

## Summary by CodeRabbit

* **New Features**
  * Added automatic recording of bot-paid gas costs from confirmed on-chain transactions.
  * Added ETH/USD valuation using block-pinned Pyth price data, with configurable oracle settings.
  * Integrated cost tracking across rebalancing, transfers, deposits, withdrawals, wraps, unwraps, and recovery flows.
  * Added asynchronous retries, idempotency, conflict detection, and delayed redrive handling.
* **Documentation**
  * Documented bot gas recording, valuation, retry behavior, and configuration requirements.
* **Tests**
  * Added unit, integration, and end-to-end coverage for recording, valuation, recovery, and failure scenarios.


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1093?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1093** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1093?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1093?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1049** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1049?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1045** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1045?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1047](https://github.com/ST0x-Technology/st0x.liquidity/pull/1047) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1047?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1044** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1044?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1093).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

config/prod/st0x-hedge.toml
```diff
@@ -78,6 +78,22 @@ organization_id = "b100145e-7894-4c17-b3e7-160435f84803"
...
+# contract-address registry at
+# https://docs.pyth.network/price-feeds/contract-addresses/evm. Pinned by
+# crates/config's repo_config_bot_gas_valuation_pyth_contract_matches_base_deployment
+# test against BASE_PYTH_CONTRACT_ADDRESS (src/onchain/pyth/mod.rs) -- update
+# both if this deployment address ever changes.
+# eth_usd_feed_id is Pyth's Crypto.ETH/USD feed id, per the registry at
+# https://www.pyth.network/developers/price-feed-ids#pyth-evm-stable (also
+# resolvable via https://pyth.network/price-feeds/crypto-eth-usd).
+[bot_gas_valuation]
+pyth_contract = "0x8250f4aF4B972684F7b336503E2D6dFeDeB1487a"
```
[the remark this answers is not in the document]
JuaniRios: False positive. Each file has exactly one `[bot_gas_valuation]` table.

Checked at the current head: `grep -c '^\[bot_gas_valuation\]'` returns 1 for `config/prod/st0x-hedge.toml`, and 1 for `config/staging/st0x-hedge.toml`. Nothing to fix here.

findolor (approved):

src/bot_gas/job.rs
```diff
@@ -0,0 +1,1315 @@
...
+                // site (e.g. a construction site that forgot
+                // `.with_bot_gas_enqueuer(...)`) is visible rather than
+                // silently dropping the cost fact.
+                debug!(
+                    chain = %job.chain,
+                    tx_hash = %job.tx_hash,
+                    category = %job.category,
+                    "bot-gas cost recording skipped: enqueuer is Disabled"
+                );
+                Ok(())
```
rouzwelt: BotGasReceiptCostEnqueuer::Disabled logs at debug!, but the comment justifies the log as making "an accidentally-Disabled production call site … visible rather than silently dropping the cost fact." Prod log levels typically filter debug, so the accident this exists to catch would in practice be silent. Disabled is only legitimate on CLI paths where noise doesn't matter — bump it to warn! so a mis-wired server site actually surfaces.

rouzwelt (changes requested):

src/bot_gas/job.rs
JuaniRios: Bumped to warn and updated the comment to say why. Fixed in 95520861.

crates/config/src/bot_gas_valuation.rs
```diff
@@ -0,0 +1,85 @@
...
+        let result: Result<BotGasValuationConfig, _> = toml::from_str(
+            r#"eth_usd_feed_id = "0xff61491a931112ddf1bd8147cd1b641375f79f5825126d665480874634fd0ace""#,
+        );
+
+        let error = result.unwrap_err();
+        assert!(
+            error.to_string().contains("pyth_contract"),
+            "expected missing-field error for pyth_contract, got: {error}"
+        );
+    }
```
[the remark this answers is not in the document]
JuaniRios: Keeping the substring asserts: they pin the exact failure kind and field name, and this matches the existing loader.rs pattern for TOML parse errors. Full-representation equality would couple the tests to toml-crate span formatting.

src/bot_gas/job.rs
```diff
@@ -0,0 +1,1317 @@
...
+                // is harmless.
+                warn!(
+                    chain = %job.chain,
+                    tx_hash = %job.tx_hash,
+                    category = %job.category,
+                    "bot-gas cost recording skipped: enqueuer is Disabled"
+                );
+                Ok(())
+            }
+        }
```
rouzwelt, line 85: Disabled warn is the only log in this module without target: "rebalance".

Every other warn!/error! here (Lines 328, 373, 457, 476) sets the target, so this one lands on a different filter — exactly the log the doc comment says must "surface under prod log filtering".

```
    warn!(
+       target: "rebalance",
        chain = %job.chain,
```

src/bot_gas/job.rs
```diff
@@ -0,0 +1,1317 @@
...
+        )
+        .await
+        {
+            Ok(price) => price,
+            Err(error) if is_transient_rpc_error(&error) => {
+                return self
+                    .redrive_or_dead_letter(ctx, "valuation RPC error", Some(&error))
+                    .await;
+            }
+            Err(error) => return Err(error.into()),
```
rouzwelt, line 417: Coverage gap: the transient-vs-terminal split on valuation errors is untested.

is_transient_rpc_error is the switch deciding whether a Pyth failure consumes the 10-minute redrive window or dead-letters immediately, and it's the only branch in perform with no test. The redrive side is covered indirectly by the receipt/block tests; the terminal side (a genuinely bad reading, e.g. an implausible expo) is not covered at all, so a future variant added to EthUsdValuationError on the wrong arm would silently turn a data error into 20 pointless RPC redrives.

src/rebalancing/usdc/mod.rs
```diff
@@ -40,6 +41,8 @@ pub(crate) enum UsdcTransferError {
     AlpacaBrokerApi(#[from] AlpacaBrokerApiError),
     #[error("CCTP bridge error: {0}")]
     Cctp(#[from] Box<CctpError>),
+    #[error("Failed to enqueue bot-gas receipt cost recording: {0}")]
```
rouzwelt, line 52: #[from] QueuePushError makes this variant capture any queue-push failure in the USDC path.

Every other error type in this stack deliberately avoids the blanket conversion for exactly this reason: MintError::BotGasEnqueue in src/rebalancing/equity/mod.rs line 471 is constructed explicitly (documented as "built here (not via #[from])"), and EquityRedemptionError::BotGasEnqueueFailed wraps a BotGasEnqueueFailure rather than a raw QueuePushError.

Here, any ? on a QueuePushError anywhere inside the CrossVenueCashTransfer code paths that return UsdcTransferError becomes BotGasEnqueue, and is_bot_gas_enqueue_failure will then report true for it. The job layer treats that as best-effort bookkeeping: Ok(()) plus an unbounded, budget-free 30 s redrive with no notifier — so an unrelated durable push failure would loop silently instead of paging an operator.

Constructing the variant explicitly at the two enqueue sites (mirroring the mint side) closes this off by construction:

```
-    #[error("Failed to enqueue bot-gas receipt cost recording: {0}")]
-    BotGasEnqueue(#[from] crate::conductor::job::QueuePushError),
+    /// Constructed only at the bot-gas enqueue sites (deliberately NOT
+    /// `#[from]`): a blanket conversion would let any queue-push failure in
+    /// this module classify as best-effort bookkeeping and redrive unbounded.
+    #[error("Failed to enqueue bot-gas receipt cost recording: {0}")]
+    BotGasEnqueue(crate::conductor::job::QueuePushError),
```

Please confirm whether src/rebalancing/usdc/manager.rs (not supplied in this review) has any other QueuePushError that flows into UsdcTransferError.

rouzwelt (changes requested):

src/bot_gas/job.rs
JuaniRios: Fixed in 2f77b4d9. Added `target: "rebalance"` so it lands on the same filter as the rest of the module.

src/bot_gas/job.rs
JuaniRios: Fixed in 2f77b4d9. Added tests for both sides of the switch: `implausible_price_exponent_dead_letters_without_redrive` (a Pyth reading with expo -30 returns a terminal `Valuation(ExponentOutOfRange)` and enqueues no redrive) and `rpc_error_reading_price_redrives_without_terminal_error` (an RPC failure on the price read redrives with an incremented counter).

src/rebalancing/usdc/mod.rs
JuaniRios: Fixed in 2f77b4d9. Dropped the `#[from]` and construct the variant explicitly in `CrossVenueCashTransfer::enqueue_bot_gas_cost` via `map_err`, mirroring `MintError::BotGasEnqueue`. The variant now carries a doc comment explaining why the blanket conversion is deliberately absent.

src/conductor.rs
```diff
@@ -1474,6 +1602,51 @@ async fn build_query_frameworks(
...
+            #[cfg(feature = "test-support")]
+            circle_api_base: rebalancing_ctx.circle_api_base.clone(),
+            #[cfg(feature = "test-support")]
+            token_messenger: rebalancing_ctx.token_messenger,
+            #[cfg(feature = "test-support")]
+            message_transmitter: rebalancing_ctx.message_transmitter,
+        },
+    )
+    .map_err(Into::into)
+}
```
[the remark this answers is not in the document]
JuaniRios: Agreed the pairing should be compiler-checked, but the hazard is pre-existing and unrelated to bot-gas cost recording, so changing the three signatures is scope creep here. Tracked in RAI-1605.

src/conductor.rs
```diff
@@ -1578,15 +1771,18 @@ fn spawn_rebalancing_infrastructure<Chain: Wallet + Clone>(
...
+                raindex_service.clone(),
+                vault_lookup.clone(),
+                tokenizer.clone(),
+                wrapper.clone(),
+                market_maker_wallet,
+                built.mint.clone(),
+                built.redemption.clone(),
+            )
+            .with_bot_gas_enqueuer(bot_gas_enqueuer.clone()),
+        );
```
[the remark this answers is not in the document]
JuaniRios: Agreed a required parameter beats a `warn!` log. Deferring: it touches ~20 construction sites across `src/` and tests, and `CrossVenueCashTransfer` has the same setter, so a consistent fix belongs in its own PR. Tracked in RAI-1606.

config/prod/st0x-hedge.toml
```diff
@@ -82,6 +82,22 @@ organization_id = "b100145e-7894-4c17-b3e7-160435f84803"
...
```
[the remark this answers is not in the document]
JuaniRios: Not reproducible. Both files have exactly one `[bot_gas_valuation]` header: grep returns `config/prod/st0x-hedge.toml:97` and `config/staging/st0x-hedge.toml:83`, one hit each. The same applies to the equity tables flagged in the out-of-diff comment - one header per symbol in both files.

rouzwelt (approved):

PR #1098, opened 2026-07-27
docs: spec multichain operation and settlement topology

0xgleb:
## Motivation

Issuance is now multichain; the liquidity bot is not. Its chain identity is
baked in rather than configured: `BridgeDirection` is a closed enum of
`EthereumToBase`/`BaseToEthereum`, the CCTP domains are compile-time constants,
the wallet demands both a Base and an Ethereum RPC, and a fill's identity is
`(tx_hash, log_index)` with no room for the chain it happened on.

Two things are blocked on that. The bot cannot trade one underlying across
several networks, and — less obviously — it cannot run anywhere the broker also
settles cash, which is why a test-network environment has never been reachable
and staging still runs on production chains.

The framing worth agreeing before planning is that CCTP is a USDC transport, not
a domain concept. Rebalancing's concern is holding the target mix where trading
needs it; whether that requires moving value between two networks is a property
of the deployment. Once the trading network and the broker deposit network can
be the same, the transport drops out of the picture rather than becoming a
no-op.

## Solution

Adds a `## Multi-chain` section to `SPEC.md` describing the intended end state.

- Per-network chain registry replacing the single-chain `EvmCtx`, carrying the cash-token address and transport endpoints so no chain-specific constant survives in code, validated at startup against the chain id and token contract.
- Settlement topology is configured and the transport is selected by it; a same-network deployment performs no bridge operation, and the wrong-network failure is caught by startup validation rather than by the type system.
- Describes how the bridge-anchored crash-safety state machine collapses when there is no burn or mint to anchor recovery on.
- Netting is scoped: exposure nets at the symbol for hedge sizing because there is one broker account, but inventory does not — tokens are deliverable only on the chain holding them, so a net-flat symbol can be untradeable on every network individually. Inventory targets, limits, and the in-flight guard are per `(symbol, network)`.
- Fill identity widens to `(network, tx_hash, log_index)` in the **event store** — the `OnChainTrade` aggregate id and the trade id keying `Position`'s exactly-once dedupe — not only in derived read models.
- Per-network fill monitors and checkpoints, per-network gas thresholds, and assets keyed by `(network, address)`; freeze status stays per-underlying.
- Invariants and rejected alternatives recorded, including per-network `Position` and extending `BridgeDirection` per network pair.

Spec only, no behaviour change. Delivery stages will be tracked in Linear rather
than in the spec.

Pairs with the `st0x.issuance` staging-environment spec, which owns the
environment half of the same picture.

## Checks

- [ ] added comprehensive test coverage for any changes in logic — n/a, spec only
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -4164,3 +4164,234 @@ a fresh direct send and `InitiateDeposit` re-enters the leg from `Bridged`,
 where the scan finds the already-submitted transfer and adopts it instead of
 forwarding the minted USDC twice. Once `InitiateDeposit` is recorded, resume
 re-polls by the recorded send tx without sending again.
+
+## Multi-chain
+
+A **network** here is one EVM chain the bot acts on, carrying its own chain id,
+endpoints, contract addresses, and confirmation semantics, and identified by a
+wire name (`base`, `ethereum`) chosen to match the name the issuer and the
+broker use for the same chain in tokenization requests. It is the unit the chain
+registry below is keyed by.
+
+The bot operates on a configured set of networks rather than on Base
+specifically. Today Base is not merely the default but the only expressible
+choice: the chain-dependent values are single-valued in configuration and the
+USDC transport's endpoints are compile-time constants, so a second network
+cannot be described at all. Two capabilities follow from removing that, and they
+are worth separating because they have different consumers:
+
+- The same underlying can be traded on more than one network at once, with a
+  single hedge against the combined exposure.
+- The whole stack can run on a network that is not Base — including one where
+  the broker also settles cash, which is what makes a test-network environment
+  reachable without a bridge.
+
+What stays singular: one bot process, one database, one broker account, and one
+`Position` per symbol.
+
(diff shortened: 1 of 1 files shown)


0xgleb:
* **#1098** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1098?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1098?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1098).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #1104, opened 2026-07-28
Inventory mismatch verification and reconciliation

ueco-jb:
## What
Self-healing for persistent divergence between the broker and the inventory view: after N diverging polls the view converges to the broker's value on its own.
This adds the required `inventory_divergence_threshold` config field, an `InventoryDivergenceGate` that suppresses equity transfers for diverging symbols, and a `ReconcileOffchainEquity` command and event that force apply the broker's value.

## Why
When a redemption completes, the view credits shares at the Hedging venue based on the tokenization provider's status alone; the position is never verified against the broker's account. If the provider reports `Completed` but the shares never arrive, the view now holds shares the broker does not have. The correct value cannot get back in: the snapshot aggregate emits no event for unchanged polls, the staleness guards reject the events that do arrive, and each failed rebalancing attempt re-arms those guards. Only a restart clears the view.
We hit this in production.

## How
Detection happens at the poll boundary, the only place a fresh reading survives the dedup.
The poller compares each fetched position against the view per symbol:
- a diverging poll increments a counter and engages the gate
- a matching poll resets the counter and releases the gate
- a busy symbol freezes the counter

When the counter reaches N, the poller sends [ReconcileOffchainEquity](https://github.com/ueco-jb/st0x.liquidity/blob/7a5a2f4e5f39601f48c26ecc67dfdbf31b24722d/src/inventory/polling.rs#L692-L704). That command always emits its event and folds into aggregate state, and [the view force applies it with an error witness](https://github.com/ueco-jb/st0x.liquidity/blob/7a5a2f4e5f39601f48c26ecc67dfdbf31b24722d/src/inventory/view.rs#L1786-L1797) after revalidating under the write lock.
The apply aborts if a transfer started in the meantime. The reconcile advances the watermark and does not stamp `last_rebalancing`.

Downsides:
- elevated trust in the broker API (we are accepting the ledger overwrite); that sounds worse than it is, but I wanted to mention that a compromised or just broken broker feed sustained for N polls would overwrite balances
- N is an arbitrary value (10 polls at 60s ~= 10min)
- the poller gains a read handle on the view
- a transfer dispatched before the first diverging poll cannot be prevented (the gate needs evidence)
- an aborted apply leaves the symbol ungated until the next diverging poll

Each of the last two costs **at most** one failed cycle: the transfer runs, fails, and by the next poll the gate is up.
The deployed configs gain the new required field in this PR.

Alternatives:
- emitting events on unchanged polls (extra downstream pipe; the poller can do that locally)
- preflight transfers against the broker (one more API call; the view stays wrong; the check only runs on a transfer)
- suppress transfers per symbol after a failure (pause until a fresh snapshot is applied to the view; but if position doesn't change, the update never comes and the release condition never would fire)

## Testing
An integration test recreates the loop (a phantom credit, the dedup suppressing correct polls, the guard re-stamped between polls) and asserts the view converges at exactly N with the gate engaged throughout.
Other tests cover: the escalation (at exactly N), reset on matching polls, and freeze instead of resetting while the symbol is busy (with one test per source: inflight transfer, active mint, active redemption, open hedge order).
Forced apply tests: aborts without modifying the view if the symbol acquired an inflight transfer, a mint, or a redemption after the escalation was sent; aborts when a fill applied after the escalation's reading was fetched; skips symbols with an open hedge order; a stale reconcile event cannot overwrite fresher data; it does not stamp `last_rebalancing`.
Config parsing rejects a missing or zero threshold.
The existing guard tests are unmodified. 

## Anything else
Out of scope: a bug - a late timestamp lets stale reading pass.
- [ ] The [OnchainEquity](https://github.com/ueco-jb/st0x.liquidity/blob/7a5a2f4e5f39601f48c26ecc67dfdbf31b24722d/src/inventory/snapshot.rs#L281), [OnchainUsdc](https://github.com/ueco-jb/st0x.liquidity/blob/7a5a2f4e5f39601f48c26ecc67dfdbf31b24722d/src/inventory/snapshot.rs#L290), and [OffchainUsd](https://github.com/ueco-jb/st0x.liquidity/blob/7a5a2f4e5f39601f48c26ecc67dfdbf31b24722d/src/inventory/snapshot.rs#L334) cases stamp `fetched_at` when the command is handled instead of [before the read like the equity arm does](https://github.com/ueco-jb/st0x.liquidity/blob/7a5a2f4e5f39601f48c26ecc67dfdbf31b24722d/src/inventory/polling.rs#L559-L563).
A reading taken before a rebalance can carry a timestamp from after it and pass the `last_rebalancing` guard it should fail.

Out of scope: nearby problems of the same class.
This PRs detection and reconciliation covers offchain equity positions only. The same stuck state failure exists for other balances.
- [ ] `OffchainUSD` [drops unchainged polls](https://github.com/ueco-jb/st0x.liquidity/blob/7a5a2f4e5f39601f48c26ecc67dfdbf31b24722d/src/inventory/snapshot.rs#L326-L330), its fold is guarded by the inflight and `last_rebalancing` checks in `Inventory::on_snapshot` (same case as the original report). A phantom USDC credit would leave the view stuck on a balance the broker does not have, exactly like the SPYM case.
- [ ] `OnchainEquity` and `OnchainUsdc`: same dedup - [unchanged polls dropped](https://github.com/ueco-jb/st0x.liquidity/blob/7a5a2f4e5f39601f48c26ecc67dfdbf31b24722d/src/inventory/snapshot.rs#L275-L287). The guards differ by path but share what matters: `OnchainEquity` folds through `apply_equity_snapshot`, `OnchainUsdc` folds through `Inventory::on_snapshot`. Either way, a phantom onchain credit would not get corrected.

---
<a href="https://app.blacksmith.sh/ST0x-Technology/codesmith/st0x.liquidity/pr/1104"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787863634&installation_model_id=19370&pr_number=1104&repository=ST0x-Technology%2Fst0x.liquidity&return_to=https%3A%2F%2Fgithub.com%2FST0x-Technology%2Fst0x.liquidity%2Fpull%2F1104&signature=d335667b7dc3cdec29c9eeb0ed1b72f1dc46a053dd955b648f67ec96a7e6debe"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added bounded self-healing for persistent offchain inventory vs broker equity divergences, with consecutive-poll detection and forced reconciliation escalation.
  * Introduced a divergence gate that suppresses equity transfer/mint-redemption dispatch until reconciliation succeeds.
* **Configuration**
  * Added `inventory_divergence_threshold` (set to `10` in standard configs) with required positive validation.
* **Documentation**
  * Expanded README “Configuration” and extended SPEC with “Broker Divergence Recovery” workflow details.
* **Tests**
  * Added scenarios for threshold escalation, busy-condition aborts/counter freezing, gate suppression/release, and forced reconciliation/event behavior.


src/inventory/polling.rs
```diff
@@ -576,9 +624,159 @@ where
...
+                .iter()
+                .map(|(symbol, fetched)| {
+                    let state = if view.equity_reconciliation_busy(symbol)? {
+                        ObservedLedgerState::Busy
+                    } else {
+                        let ledger = view.equity_available(symbol, Venue::Hedging);
+                        let matches = match ledger {
+                            Some(ledger_value) => {
+                                Float::from(ledger_value).eq(Float::from(*fetched))?
+                            }
```
ueco-jb, line 683: That `Float` is Rain's internal decimal type and I assume we don't get float rounding issues here.
Just raising this for the posterity: if the ledger and the broker ever disagree by the smallest amount (the broker using a difference precision?), every poll would read it as a divergence and fire a reconcile after N polls.

src/inventory/polling.rs
```diff
@@ -128,6 +159,13 @@ where
     configured_equity_vaults: Option<BTreeMap<Address, BTreeSet<B256>>>,
     configured_usdc_vaults: Option<BTreeSet<B256>>,
     reserved_cash: Usd,
+    /// Divergence detection wiring; `None` in tests that do not exercise
+    /// divergence recovery. Production always wires it in the conductor
+    /// builder.
+    divergence_recovery: Option<InventoryDivergenceRecoveryCtx>,
+    /// Consecutive diverging polls per symbol. Kept in memory on purpose: a
+    /// restart resets the count, and restart hydration restores the view.
+    divergence_counters: Mutex<HashMap<Symbol, u32>>,
```
ueco-jb, line 168: Technically if the two `poll_and_record` runs somehow executed concurrently, the counter would not work correctly (we assume 60s interval here).
I verified though that the InventoryMonitor runs a single loop where it awaits `poll()` before the next `interval.tick()`.
https://github.com/ueco-jb/st0x.liquidity/blob/contradiction-reconciliation/src/conductor/monitor/inventory.rs#L31-L37
The monitor is built once in the builder, so the counter advances at most once per poll.

src/rebalancing/trigger/mod.rs
```diff
@@ -2645,6 +2672,22 @@ impl RebalancingService {
...
+        // lifts the gate).
+        if self.divergence_gate.is_engaged(symbol) {
+            warn!(
+                target: "rebalance",
+                %symbol,
+                "Skipped equity trigger: unresolved snapshot divergence \
+                 pending reconciliation"
+            );
+            return Ok(());
+        }
```
ueco-jb, line 2704: Any new mint/redemption is skipped until the inventory divergence is solved.

src/inventory/polling.rs
```diff
@@ -576,9 +624,159 @@ where
...
+                        position: escalation.fetched,
+                        fetched_at,
+                        ledger_position: escalation.ledger,
+                        consecutive_polls: escalation.consecutive_polls,
+                    },
+                )
+                .await?;
+
+            self.lock_divergence_counters().remove(&escalation.symbol);
+            recovery.gate.release(&escalation.symbol);
```
JuaniRios, line 765: should fix: the counter reset and the gate release run when `send` succeeds. But `send` success does not mean the view healed. `reconcile_offchain_equity` has three abort paths that return the view unchanged, and the reactor reports no error for them. On an abort, the view still diverges, the counter is back at zero, and the gate is open. Each abort then costs a full new N-poll cycle (10 minutes in prod config), so convergence is not bounded. After the send, read the view back and compare it to `escalation.fetched`. Only clear the counter and release the gate on a match. On a mismatch, keep the gate engaged and keep the counter at the threshold so the next poll re-escalates at once.

src/inventory/polling.rs
```diff
@@ -576,9 +624,159 @@ where
                 .await?;
         }
 
+        // After the OffchainEquity send: reactors run inline within `send`,
+        // so a poll the pipeline accepted has already reached the view here
+        // and reads back as a match. Whatever still diverges from the view at
+        // this point was suppressed or skipped.
+        self.detect_offchain_divergences(snapshot_id, &positions, fetched_at)
```
JuaniRios, line 631: should fix: two ordering problems combine into a starvation loop that can keep the whole mechanism from ever engaging. First, detection runs after the `OffchainEquity` send on line 578, and that send runs the reactors inline, which enqueue equity checks before any gate exists. Second, the `Busy` arm in `record_divergence_observations` neither counts nor engages the gate. Now trace the incident shape: the phantom credit enqueues a check and the first doomed transfer fires before any poll counts. The attempt marks the symbol busy, so every poll during it is frozen. The failure cleanup re-fires the next transfer within seconds, so the idle window between attempts can be shorter than one poll tick. The counter stays at zero, the gate never engages, and the original loop survives this PR. Convergence then depends on a poll randomly landing in a seconds-wide gap.

Fix: make a diverging observation on a busy symbol engage the gate without counting. The gate then only releases on a matching poll, so after any transfer the symbol needs one clean poll before the next dispatch. The cost is bounded and small: up to one poll interval (60s) added between consecutive rebalancing transfers of the same symbol, on a pipeline that runs for many minutes. Hedge orders never consult this gate, so hedging latency is unaffected. The gain is that convergence stops depending on the timing gap between failed attempts. Please also state the new invariant (one clean poll between consecutive transfers of a symbol) in the gate's doc comment and in SPEC.md.

src/inventory/polling.rs
```diff
@@ -576,9 +624,159 @@ where
...
+    /// Compare each fetched broker position against the view's Hedging
+    /// available balance and drive the divergence counter of each symbol.
+    ///
+    /// Busy symbols (inflight transfer, active mint/redemption, or open
+    /// hedge order) freeze their counter: the comparison is ambiguous while
+    /// balances are legitimately in motion, and resetting would discard the
+    /// diverging polls already counted. A matching poll resets the counter
+    /// and releases the transfer gate; a diverging poll increments it and
+    /// engages the gate; at the configured threshold the symbol escalates
+    /// via `ReconcileOffchainEquity` and resets only after the command was
```
JuaniRios: should fix: this comment documents a deliberate ordering: the counter resets only after the command was sent, so a failed send escalates again on the next poll. No test exercises that ordering. Add a test where the escalation send fails once: assert the counter is not reset, the gate stays engaged, and the next successful poll escalates again.

src/inventory/polling.rs
```diff
@@ -576,9 +624,159 @@ where
...
+                        target: "inventory",
+                        %symbol,
+                        ?ledger,
+                        broker = %fetched,
+                        consecutive_polls = *count,
+                        threshold = recovery.threshold.get(),
+                        "Offchain snapshot diverges from the inventory view"
+                    );
+
+                    if *count >= recovery.threshold.get() {
```
JuaniRios, line 831: should fix: the counter stores only how many polls disagreed, not what the broker reported. At the threshold, the escalation force-applies whatever the Nth poll returned. A flapping broker feed (N different wrong values) still escalates, and the PR description sells N as protection against a broken feed. Store the observed broker value with the count. When a diverging poll reports a different value than the stored one, replace the value and restart the count at 1. Escalate only after N confirmations of the same value.

src/rebalancing/trigger/mod.rs
```diff
@@ -1751,6 +1770,7 @@ impl RebalancingService {
 
             OnchainEquity { .. }
             | OffchainEquity { .. }
+            | OffchainEquityReconciled { .. }
```
JuaniRios, line 1788: should fix: when the normal apply of `OffchainEquityReconciled` returns an error, this recovery path force-applies the event against a reset view. The reset drops `active_mints` and `active_redemptions`, so the busy revalidation inside `reconcile_offchain_equity` passes against empty state. The retry then force-writes the broker value in exactly the situation the first attempt refused. Exclude this event from the force-apply recovery, or make the reset preserve the busy-relevant state so the revalidation stays meaningful.

src/rebalancing/trigger/mod.rs
```diff
@@ -2645,6 +2672,22 @@ impl RebalancingService {
             return Ok(());
         }
 
+        // A pending snapshot divergence means the view's balance for this
+        // symbol is suspect: a transfer sized off it would likely fail, and
+        // a failed attempt arms the guards again and marks the symbol busy,
+        // freezing the very counter that resolves the divergence. Skip
+        // until the poller clears it (a matching poll or a forced reconcile
+        // lifts the gate).
+        if self.divergence_gate.is_engaged(symbol) {
```
JuaniRios, line 2696: minor: this gate check runs once, at the start of the check. The hedge-order guard next to it is re-validated later in the flow, but the divergence gate is not. A poll can engage the gate while `build_equity_operation` awaits, and the transfer dispatches anyway. Re-check `is_engaged` immediately before the enqueue.

src/rebalancing/trigger/mod.rs
```diff
@@ -1638,7 +1653,11 @@ impl RebalancingService {
                 now,
             ),
 
-            OffchainUsd { .. }
+            // `reconcile_offchain_equity` validates busyness and the hedge
+            // gate under the write lock, so the generic apply path is the
+            // correct route here too.
+            OffchainEquityReconciled { .. }
```
JuaniRios, line 1659: minor: with rebalancing enabled, `OffchainEquityReconciled` reaches the view through this arm, not through `InventoryProjection`. The end-to-end test wires the projection only (the rebalancing-disabled topology), so the production route is untested. Add a variant of the self-heal test that wires the `RebalancingService` reactor, so a regression in this arm cannot pass CI.

src/inventory/view.rs
```diff
@@ -895,6 +895,41 @@ impl InventoryView {
...
+            || self
+                .last_offchain_fill_applied_at
+                .get(symbol)
+                .is_some_and(|filled_at| fetched_at < *filled_at))
+    }
+
+    /// Inflight at either venue, or an active mint/redemption owning the
+    /// symbol's inflight slot. The hedge order gate is checked separately
+    /// so the reconcile path can log a distinct abort reason for it.
+    fn equity_transfer_busy(&self, symbol: &Symbol) -> Result<bool, FloatError> {
```
JuaniRios, line 944: minor: `equity_reconciliation_busy` and `reconcile_offchain_equity` re-implement the same three-part busy predicate (transfer busy, pending hedge order, fill after fetch). If a future busy source is added to one and not the other, detection and apply disagree silently. Extract one shared predicate and call it from both places.

src/inventory/view.rs
```diff
@@ -1678,6 +1713,92 @@ impl InventoryView {
...
+            polls: consecutive_polls,
+        };
+
+        let view = self.update_equity(
+            symbol,
+            Inventory::force_on_snapshot(Venue::Hedging, broker_position, witness),
+            now,
+        )?;
+
+        Ok(view.record_equity_snapshot_watermarks(Venue::Hedging, [symbol], fetched_at))
```
JuaniRios, line 1809: minor: the reconcile advances the watermark but never reads it. A fresher `OffchainEquity` snapshot that applied between the escalation send and this apply is overwritten by the older broker reading. The window is small because the reactors run inline in the same send, but nothing structural prevents it. Add a watermark check here: abort when the current Hedging watermark for the symbol is newer than `fetched_at`, same as the normal path does.

src/inventory/polling.rs
```diff
@@ -576,9 +624,159 @@ where
...
+                .collect::<Result<_, FloatError>>()?
+        };
+
+        let escalations = self.record_divergence_observations(recovery, observations);
+
+        for escalation in escalations {
+            self.snapshot
+                .send(
+                    snapshot_id,
+                    InventorySnapshotCommand::ReconcileOffchainEquity {
```
JuaniRios: minor: the `?` on this send aborts the loop. When two symbols escalate in the same poll and the first send fails, the second symbol's escalation is dropped for this cycle even though its counter reached the threshold. Attempt every escalation, collect the failures, and return the first error after the loop.

src/inventory/polling.rs
```diff
@@ -576,9 +624,159 @@ where
...
+                .map(|(symbol, fetched)| {
+                    let state = if view.equity_reconciliation_busy(symbol, fetched_at)? {
+                        ObservedLedgerState::Busy
+                    } else {
+                        let ledger = view.equity_available(symbol, Venue::Hedging);
+                        let matches = match ledger {
+                            Some(ledger_value) => {
+                                Float::from(ledger_value).eq(Float::from(*fetched))?
+                            }
+                            None => false,
```
JuaniRios: minor: `None => false` classifies a never-initialized Hedging venue as a divergence even when the broker reports zero. A newly configured symbol with no positions then counts toward escalation and engages the gate for no reason. Treat `None` with a broker reading of zero as a match, or skip counting until the venue is initialized.

src/inventory/polling.rs
```diff
@@ -4068,4 +4274,485 @@ mod tests {
...
+            inventory
+                .read()
+                .await
+                .equity_available(&spym, Venue::Hedging),
+            Some(phantom),
+            "precondition: the staleness guard must keep the phantom credit in the view"
+        );
+
+        let service =
+            reconciling_service(&pool, snapshot_store, inventory.clone(), gate.clone(), 3);
```
JuaniRios: nit: the threshold is a bare `3` and the step comments narrate it by hand. Bind it to a named local and let the assert messages carry the per-poll expectations. The comments restate what the asserts already say.

src/inventory/divergence.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/// Witness for forcing a broker snapshot over the view's balance.
+///
+/// [`Inventory::force_on_snapshot`] takes the triggering error as a witness
+/// to prevent blind usage; this type records what the poller observed
+/// before the escalation fired.
+///
+/// [`Inventory::force_on_snapshot`]: super::Inventory::force_on_snapshot
+#[derive(Debug, thiserror::Error)]
```
JuaniRios: nit: `force_on_snapshot` only requires `Debug`, and the witness is only logged with Debug formatting. The `thiserror::Error` derive and the `#[error(...)]` message are never used. Drop them and keep `#[derive(Debug)]`, or state in a comment why the Display path exists.

src/inventory/divergence.rs
```diff
@@ -0,0 +1,137 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn symbol(ticker: &str) -> Symbol {
+        Symbol::new(ticker).unwrap()
+    }
+
+    #[test]
+    fn gate_engages_and_releases_symbols_independently() {
```
JuaniRios: nit: these two tests only exercise HashSet insert/remove through the wrappers. The real behavior (engage on divergence, release on match or escalation, suppression in the trigger) is covered by the polling and trigger tests. These add maintenance surface without testing logic. Consider removing them.

JuaniRios (changes requested):

src/inventory/polling.rs
ueco-jb: Good point. Fixed: [0952569](https://github.com/ST0x-Technology/st0x.liquidity/pull/1104/commits/6ec0c8f20a1259e6b6ea9adf24837a29efab6a9f)

src/inventory/polling.rs
ueco-jb: > (...) The failure cleanup re-fires the next transfer within seconds, so the idle window between attempts can be shorter than one poll tick. 

After a terminal mint failure cancels the equity checks: https://github.com/ueco-jb/st0x.liquidity/blob/6ec0c8f20a1259e6b6ea9adf24837a29efab6a9f/src/rebalancing/trigger/mod.rs#L4807-L4815
so nothing scheduled by the failure should re-fire the transfer. The next check should come from the next snapshot event, happening on the next tick.

> The counter stays at zero, the gate never engages, and the original loop survives this PR.

Busy also never releases, it does nothing in both directions. The gate needs just one idle poll to engage, and once engaged it blocks every further mint while the counter climbs to N. Starvation would require every poll forever to land inside a mint attempt, but the next attempt is only triggered by the poll tick itself, right?
And the comparison runs milliseconds before the attempt can mark the symbol busy.

src/inventory/polling.rs
ueco-jb: If the broker API is flapping, it means it can't be used as a source of truth anyway (and that the whole system is borken).
Storing the broker value would also extend each divergence window, locking the asset without any actual reconciliation. 

I think that at some point we just need to accept whatever broker tells us (because we can't create those stocks out of thin air anyway). I don't think that storing the values would improve the situation, while it could make it worse.

src/inventory/polling.rs
ueco-jb: After the meeting - I need to verify if the terminal mint failure cancels only the current check or any future ones.
What happens in a situation from 0 to 1 (after the first read and divergence detection the symbol is locked anyway, so no mint will go through afterwards).

ueco-jb: Okay to summarize:
A terminal mint cancels every queued equity check and enqueues only a USDC one. The next check can only come from the next poll tick's snapshot event - and that same tick runs the divergence comparison first.
So the loop gets exactly one free attempt before any diverging poll is observed; after the first failure the comparison always wins, the gate closes at count 1, and it never reopens while the divergence persists - starvation would need the poll to lose that race every tick forever.

If multiple different transfers fire, they all go through the same flow. And per symbol, there is the `equity_in_progress` guard that allows one active transfer per symbol.

So the one free attempt is per symbol, then it gets gated.

src/inventory/polling.rs
ueco-jb: This got covered already in the verify-heal fix change: [0952569](https://github.com/ST0x-Technology/st0x.liquidity/pull/1104/commits/6ec0c8f20a1259e6b6ea9adf24837a29efab6a9f)

src/rebalancing/trigger/mod.rs
ueco-jb: > Exclude this event from the force-apply recovery, or make the reset preserve the busy-relevant state so the revalidation stays meaningful.

The former. :+1: 55a0efc0e47bd955e006207f95da78a900b4e9c9

src/rebalancing/trigger/mod.rs
ueco-jb: Done d7bacdca975f340b427e26607c7ee522079e9b2a

src/rebalancing/trigger/mod.rs
ueco-jb: Done. 278c7f0620383a5d18a5317e5a926ca063a08ccc

src/inventory/divergence.rs
ueco-jb: Removed. f9f3578bf627873430fcc66b6fd6cd1c310fdb2e

src/inventory/view.rs
ueco-jb: Good point, done. 26cb1c67f93199855b86ed02e54e3ceb617e9fc7

src/inventory/divergence.rs
ueco-jb: Done. e1a17df1109a80d20fc15af67352f562af787348

src/inventory/polling.rs
ueco-jb: Done. fe94563100105c12eccebd4f18cd5f6591ff8cd1

src/inventory/polling.rs
ueco-jb: Good point, done. 2b4e2a7ec642f7d6cc9a24b8be69de99936fe9ba

src/inventory/view.rs
ueco-jb: > Add a watermark check here: abort when the current Hedging watermark for the symbol is newer than fetched_at, same as the normal path does.

Done. 7e2798780bf02cf7679e67988c9ec6bc94675334
Not where you pointed out though, because the abort must run before `force_on_snapshot` overwrites the balance.

src/inventory/polling.rs
ueco-jb: Done: 7fd7957cf115c60cbaef5d15ca248a298c8943ce

src/inventory/snapshot.rs
```diff
@@ -284,6 +302,23 @@ impl EventSourced for InventorySnapshot {
...
+                fetched_at,
+                ledger_position,
+                consecutive_polls,
+            } => Ok(vec![InventorySnapshotEvent::OffchainEquityReconciled {
+                symbol,
+                position,
+                fetched_at,
+                ledger_position,
+                consecutive_polls,
+            }]),
```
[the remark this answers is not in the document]
ueco-jb: I added an error. [04cb203](https://github.com/ST0x-Technology/st0x.liquidity/pull/1104/commits/04cb20305156c0ce110e5018f8a5bfab2d7ca1be)

src/inventory/polling.rs
```diff
@@ -576,9 +624,182 @@ where
...
+    /// fresh N-poll cycle.
+    async fn detect_offchain_divergences(
+        &self,
+        snapshot_id: &InventorySnapshotId,
+        positions: &BTreeMap<Symbol, FractionalShares>,
+        fetched_at: DateTime<Utc>,
+    ) -> Result<(), InventoryPollingError<Exe::Error>> {
+        let Some(recovery) = &self.divergence_recovery else {
+            return Ok(());
+        };
```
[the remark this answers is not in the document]
ueco-jb: 0719e4c7e7a34c50bab3fb0f7c5b8858bff2289b

JuaniRios (approved):
LGTM, well done 👏

PR #1111, opened 2026-08-03
fix: prevent skipped CI from satisfying required checks

findolor:
## What

Closes RAI-1602.

Prevents a pull request head SHA from satisfying the required CI checks when validation work is skipped or a replacement pull request event cancels an older run.

## Why

The Graphite CI optimizer can intentionally skip a middle-of-stack pull request, and GitHub treats skipped required jobs as successful. A concurrent head/base update and metadata edit could therefore cancel the validating run while the surviving run reported `backend`, `dashboard`, and `hooks` green without executing the required validation.

## How

- Remove the Graphite optimizer from the required workflow so every subscribed pull request event performs validation.
- Put all events for the same pull request merge ref in one cancel-in-progress group; a later event replaces the earlier run with equivalent validation.
- Keep expensive backend, dashboard, and hooks work cancellable while routing branch protection through small `always()` gate jobs that fail closed on failed, skipped, cancelled, or unknown dependency states.
- Preserve the docs-only backend/dashboard shortcut only when change classification succeeds and explicitly confirms there are no code changes.
- Avoid an unnecessary checkout when the pull request files API supplies the classifier input.
- Expand the CI invariant harness to verify concurrency, exact required check names, job wiring, classifier configuration, gate enforcement, and dependency-result matrices.
- Document the required-check and replacement-run invariants.

## Testing

- [x] `./scripts/test-fork-pr-ci.sh`
- [x] Full repository pre-commit suite
- [x] `nix run .#ci` — 4,184 tests passed with 6 configured skips; workspace checks, Clippy, rustfmt, DTO regeneration, and dashboard checks passed
- [x] Three-angle simplification review plus two independent round-one reviews and one clean round-two review
- Live GitHub verification: this description update intentionally replaces the synchronize run; the surviving run must execute every backend job and publish `backend`, `dashboard`, and `hooks`

## Screenshots

N/A

## Anything else

Middle-of-stack pull requests now spend the full CI cost instead of relying on another stack branch. The required gate jobs add small runner hops so branch protection receives deterministic, fail-closed check contexts even when validation work is cancelled.

## Summary by CodeRabbit

* **CI Improvements**
  * Improved pull-request validation reliability with clearer cancellation and rerun behavior.
  * Required checks now run independently and fail safely when validation cannot confirm documentation-only changes.
  * Dashboard, hooks, and backend checks provide more consistent build and test coverage.
  * Enhanced safeguards for forked pull requests, caching, permissions, and large changes.

* **Documentation**
  * Clarified when CI runs, what it validates, and how pull-request updates are handled.
  * Updated contributor guidance to reflect current build and test behavior.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,12 +7,12 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review, edited]
 
-# Metadata-only pull request edits (title/body, see `optimize_ci`) get their
-# own group. They skip the real checks, so sharing the group would let a title
-# edit cancel a live run and then report every required check as skipped --
-# which branch protection reads as a pass on code that was never tested.
+# Pull request events for the same merge ref share a group. A later head, base,
+# title, or body update may replace an older run, but every replacement performs
+# the same validation, so cancellation cannot leave behind a green no-op run.
+# Master keeps one group per SHA so independent post-merge runs do not cancel.
 concurrency:
-  group: ${{ github.repository }}-${{ github.workflow }}-${{ github.ref }}-${{ github.event.action == 'edited' && !github.event.changes.base && 'metadata-edit' || '' }}-${{ github.ref == 'refs/heads/master' && github.sha || ''}}
+  group: ${{ github.repository }}-${{ github.workflow }}-${{ github.ref }}-${{ github.ref == 'refs/heads/master' && github.sha || ''}}
   cancel-in-progress: true
 
 # CI validates untrusted pull request code, so its repository access stays
@@ -25,8 +25,8 @@ env:
   # GitHub holds fork pull_request runs for maintainer approval, removes every
   # repository secret, and limits GITHUB_TOKEN to read access. Dependabot runs
   # have the same restrictions despite using a branch in this repository. Keep
-  # an explicit flag as well, so no authenticated optimizer or cache publisher
-  # is even invoked for untrusted code.
+  # an explicit flag as well, so no authenticated cache publisher is invoked
+  # for untrusted code.
   UNTRUSTED_PR: ${{ github.event_name == 'pull_request' && (github.event.pull_request.head.repo.fork || github.event.pull_request.user.login == 'dependabot[bot]') }}
 
 # Nix is installed via nix-quick-install-action and both the shared
@@ -36,52 +36,6 @@ env:
 # which did not share the `rainlanguage` cache that raindex CI populates.
 
 jobs:
-  optimize_ci:
-    runs-on: "blacksmith-2vcpu-ubuntu-2404"
-    timeout-minutes: 5
-    outputs:
-      skip: ${{ steps.metadata_edit.outputs.skip == 'true' && 'true' || steps.check_skip.outputs.skip }}
-    steps:
-      # `edited` is in the trigger list for base retargets: a stacked pull
-      # request keeps its head commit when its parent merges, so only the
-      # retarget event tells CI the diff under test changed. GitHub fires the
-      # same event for title and body edits, which change no code at all, and
-      # `/pr-description` edits both right after submit. `github.event.changes`
-      # carries `base` only on a retarget, so its absence marks the edit as
-      # metadata-only and reuses the optimizer's `skip` contract to stand every
-      # real job down.
-      - name: Detect metadata-only pull request edit
-        id: metadata_edit
-        if: ${{ github.event.action == 'edited' && !github.event.changes.base }}
-        run: echo "skip=true" >> "$GITHUB_OUTPUT"
-
-      # Pinned to the v0.0.9 release commit, not a mutable ref: this action
-      # gates all CI and holds the Graphite token, so the SHA freezes both the
-      # supply chain and the `skip` output contract (set in dist/index.js but
-      # undeclared in the action's action.yml).
-      #
-      # Fork pull requests are not Graphite stacks. The optimizer's `skip`
-      # contract is not defined for them, and a `true` would skip the real
-      # jobs, which branch protection then reads as a pass. Do not invoke the
-      # optimizer or expose its token for untrusted code.
-      #
-      # The STEP carries this condition, not the job, on purpose. The job still
-      # runs and still succeeds, so no dependent job can be cascade-skipped by
-      # a skipped dependency. `skip` is then simply empty, and every gate below
-      # tests `!= 'true'`, so all of them fall into the run-the-real-checks
-      # bucket.
-      - name: Optimize CI
-        id: check_skip
-        if: ${{ env.UNTRUSTED_PR != 'true' && steps.metadata_edit.outputs.skip != 'true' }}
-        uses: withgraphite/graphite-ci-action@9bc969adfd43bb790da3b64b543c78c75cef9689  # v0.0.9
-        with:
-          graphite_token: ${{ secrets.GRAPHITE_CI_OPTIMIZATION_TOKEN }}
-          # Declared required by the action but unused at the pinned version
-          # (it reads only graphite_token/endpoint/timeout; there is no
-          # run-cancel path). Skipping is enforced solely by the `skip` output
-          # and the job-level if-gates.
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-
   # Classifies the push: `code == 'true'` when at least one changed file is
   # NOT a docs file. With `predicate-quantifier: every`, a file matches the
   # `code` filter only if it matches `**` AND every negated docs glob -- i.e.
@@ -90,17 +44,11 @@ jobs:
   # all-docs push yields `code == 'false'`. If this job errors or is
   # skipped, `code` is empty/unknown and the gates require the real checks.
   #
-  # Runs whenever the optimizer did NOT confirm a skip (`skip != 'true'`),
-  # including when `optimize_ci` itself failed (empty skip) -- so a broken
-  # optimizer falls into the run-CI bucket, never the skip bucket.
-  #
   # NOTE: paths-filter uses the Pull Requests API for pull_request runs and git
   # change detection for push/workflow_dispatch, so the checkout is needed for
   # the latter two events.
   changes:
     name: changes
-    needs: optimize_ci
-    if: ${{ always() && needs.optimize_ci.outputs.skip != 'true' }}
     runs-on: blacksmith-2vcpu-ubuntu-2404
     timeout-minutes: 5
     outputs:
@@ -110,6 +58,7 @@ jobs:
       code: ${{ github.event_name == 'pull_request' && github.event.pull_request.changed_files >= 3000 && 'true' || steps.filter.outputs.code }}
     steps:
       - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803  # v6
+        if: ${{ github.event_name != 'pull_request' }}
         with:
           persist-credentials: false
 
@@ -127,8 +76,8 @@ jobs:
 
   backend-build:
     name: backend build
-    needs: [optimize_ci, changes]
-    if: ${{ needs.optimize_ci.outputs.skip != 'true' && needs.changes.outputs.code == 'true' }}
+    needs: changes
+    if: ${{ needs.changes.outputs.code == 'true' }}
     runs-on: blacksmith-8vcpu-ubuntu-2404
     timeout-minutes: 30
     env:
@@ -206,8 +155,8 @@ jobs:
 
   backend-test:
     name: backend test ${{ matrix.partition }} of 4
-    needs: [optimize_ci, changes, backend-build]
-    if: ${{ needs.optimize_ci.outputs.skip != 'true' && needs.changes.outputs.code == 'true' }}
+    needs: [changes, backend-build]
+    if: ${{ needs.changes.outputs.code == 'true' }}
     runs-on: blacksmith-8vcpu-ubuntu-2404
     timeout-minutes: 30
     strategy:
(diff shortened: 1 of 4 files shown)


findolor:
* **#1111** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1111?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1111?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1111).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

.github/workflows/ci.yaml
```diff
@@ -378,14 +327,12 @@ jobs:
...
-  # docs-only pushes -- because pre-commit is the only check that validates
-  # docs files (deno fmt covers markdown). Short-circuiting it on docs-only
-  # pushes would let unformatted docs merge green and then break
-  # `pre-commit run --all-files` for every subsequent push on any branch.
+  # Required check. Runs fully on docs-only pushes because pre-commit is the
+  # only check that validates docs files (deno fmt covers markdown).
+  # Short-circuiting it on docs-only pushes would let unformatted docs merge
+  # green and then break `pre-commit run --all-files` for every subsequent push
+  # on any branch.
   hooks:
```
0xgleb: critical: dropping both `needs` and `if` here leaves `hooks` on the implicit `success()` condition, so if the run is cancelled before this job starts (manual cancel, or a later edited/synchronize event in the now-shared concurrency group), `hooks` resolves to "skipped", which branch protection treats as a pass. That reopens the exact skip-as-pass class this PR closes, just via cancellation instead of the optimizer, and the new `expected_wiring` table even locks `"hooks" => { "needs" => nil, "if" => nil }` in as the correct state. Give `hooks` the same `if: ${{ always() }}` treatment `backend` got and update the wiring assertion to require it.

.github/workflows/ci.yaml
```diff
@@ -299,13 +247,14 @@ jobs:
...
+  # Required check. The real build steps are skipped only on a confirmed
+  # docs-only push (changes succeeded AND code == 'false'), so a docs-only push
+  # reports this check green without spinning up the dashboard toolchain.
   dashboard:
-    needs: [optimize_ci, changes]
-    if: ${{ always() && needs.optimize_ci.outputs.skip != 'true' }}
+    needs: changes
+    # The backend gate fails when the workflow is canceled. Do not make its
+    # replacement wait for an obsolete dashboard build as well.
+    if: ${{ !cancelled() }}
```
0xgleb: should fix: `!cancelled()` is the functional opposite of the old `always()`, this job now skips once the run is cancelled, and a not-yet-started job on a cancelled run reports "skipped", which counts as a green required check. The comment assumes the backend gate covers for it, but `backend` and `dashboard` run independently off `needs: changes` with no ordering guarantee, so a cancellation landing after backend's deps finish but before dashboard runs leaves a fabricated pass on that SHA, with a broken dashboard build able to merge. The remaining safety story depends on GitHub preferring the latest same-named check run, which is stated and tested nowhere. Either give `dashboard` the same `always()` plus explicit result-check split that `backend` has, or document and test that assumption in the harness.

.github/workflows/ci.yaml
```diff
@@ -110,6 +58,7 @@ jobs:
       code: ${{ github.event_name == 'pull_request' && github.event.pull_request.changed_files >= 3000 && 'true' || steps.filter.outputs.code }}
     steps:
       - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803  # v6
+        if: ${{ github.event_name != 'pull_request' }}
```
0xgleb, line 61: minor: this new step-level condition is asserted nowhere in the extended harness, the Ruby block only checks job-level `needs`/`if` wiring. A future edit that flips or drops it would pass every assertion, and an inverted condition would break paths-filter's git-based detection on push events. Worth adding a step-level assertion for the checkout `if`, mirroring the `expected_wiring` pattern.

scripts/test-fork-pr-ci.sh
```diff
@@ -36,6 +45,138 @@ assert_count 1 '    branches: [master]'
...
+  ["README.md"] => false,
+  ["AGENTS.md"] => false,
+  ["src/lib.rs"] => true,
+  ["dashboard/src/app.ts"] => true,
+  ["docs/ci.md", "src/lib.rs"] => true
+}
+classification_cases.each do |paths, expected|
+  actual = paths.any?(&code_file)
+  next if actual == expected
+
```
0xgleb: minor: `code_file` reimplements what dorny/paths-filter should compute, and `classification_cases` only checks that reimplementation against itself, the real filter is never exercised. If the action's glob semantics diverge from this model on an edge case, a code change could classify as docs-only and skip real validation, and the fail-closed path only covers the classifier erroring, not it returning a wrong boolean. Either rely on the fail-closed design plus the literal filter-list assertion above, or run the actual glob engine against these sample paths.

.github/workflows/ci.yaml
```diff
@@ -7,12 +7,12 @@ on:
...
-# edit cancel a live run and then report every required check as skipped --
-# which branch protection reads as a pass on code that was never tested.
+# Pull request events for the same merge ref share a group. A later head, base,
+# title, or body update may replace an older run, but every replacement performs
+# the same validation, so cancellation cannot leave behind a green no-op run.
+# Master keeps one group per SHA so independent post-merge runs do not cancel.
 concurrency:
-  group: ${{ github.repository }}-${{ github.workflow }}-${{ github.ref }}-${{ github.event.action == 'edited' && !github.event.changes.base && 'metadata-edit' || '' }}-${{ github.ref == 'refs/heads/master' && github.sha || ''}}
+  group: ${{ github.repository }}-${{ github.workflow }}-${{ github.ref }}-${{ github.ref == 'refs/heads/master' && github.sha || ''}}
   cancel-in-progress: true
```
0xgleb, line 16: should fix: dropping the `metadata-edit` discriminator puts title and body `edited` events in the same cancel-in-progress group as the run already validating that head SHA. The comment this PR deletes recorded that our own `/pr-description` flow edits both right after submit, and Graphite stacks rewrite bodies routinely, so this is the expected path on every submit rather than a rare event. Each one now cancels the in-flight run and restarts `backend-build` on an 8vcpu runner with a 30 minute timeout, the four way `backend-test` matrix, `dashboard-build`, and `hooks-run` from scratch on identical code, with cold cargo state the cancelled run had already paid for. The safety property is worth having, but the fix for the old hole was making the replacement run validate, not merging the groups. Restoring the suffix while keeping full validation in that run closes the hole and stops a title edit killing a live run. That needs `expected_group` in the harness and the matching paragraph in docs/ci.md updated too.

scripts/test-fork-pr-ci.sh
```diff
@@ -36,25 +45,210 @@ assert_count 1 '    branches: [master]'
...
+    [%w[success true success success], true],
+    [%w[success true failure success], false],
+    [%w[success true skipped success], false],
+    [%w[success true cancelled success], false],
+    [%w[success true success failure], false],
+    [%w[success true success skipped], false],
+    [%w[success true success cancelled], false],
+    [%w[failure unknown skipped skipped], false],
+    [%w[skipped unknown skipped skipped], false],
+    [%w[cancelled unknown skipped skipped], false]
```
0xgleb, line 215: should fix: neither gate matrix ever pairs a non-success `needs.changes.result` with `code == 'false'`, since every failure, skipped, and cancelled case uses the `unknown` sentinel for `code`. That makes the `[ "${{ needs.changes.result }}" = "success" ] &&` guard non load bearing: weakening it to `!= "failure"` or anything admitting skipped or cancelled still passes all eleven backend cases and all eight dashboard ones. The `expressions.each` presence check only catches deleting the reference outright, not weakening the comparison. The matrices also never exercise `code == ''`, which is what a failed or cancelled `changes` job actually produces, so that path is untested. Adding `[%w[failure false skipped skipped], false]`, the skipped and cancelled equivalents, and `[["success", "", "skipped", "skipped"], false]` to backend, plus the three element versions for dashboard, would make the conjunct actually load bearing.

scripts/test-fork-pr-ci.sh
```diff
@@ -36,25 +45,210 @@ assert_count 1 '    branches: [master]'
...
+
+check_gate = lambda do |job_name, step_name, expressions, cases|
+  gate_job = jobs.fetch(job_name)
+  gate_step = gate_job.fetch("steps").find do |step|
+    step["name"] == step_name
+  end
+  unless gate_job.fetch("continue-on-error", false) == false &&
+      gate_step&.fetch("if", nil).nil? &&
+      gate_step&.fetch("continue-on-error", false) == false
+    warn "#{job_name} gate enforcement must run and propagate failures"
```
0xgleb, line 166: should fix: `continue-on-error` is asserted false on the gate jobs and gate steps, but nothing asserts it on `backend-build`, `backend-test`, `dashboard-build`, or `hooks-run`. A job level `continue-on-error: true` on any of those makes its reported result `success` even when its steps fail, and the gates only compare `needs.<job>.result` against the literal `success`, so every required check goes green with no validation. That is the same class of silent bypass this harness was expanded to prevent, and it's the one structural attribute of the work jobs left unchecked. Asserting `jobs.fetch(name).fetch("continue-on-error", false) == false` for those four after the `expected_wiring` loop would close it.

.github/workflows/ci.yaml
```diff
@@ -349,43 +289,54 @@ jobs:
...
+      - name: Check dashboard validation
+        run: |
+          if [ "${{ needs.changes.result }}" = "success" ] && [ "${{ needs.changes.outputs.code }}" = "false" ]; then
+            echo "Docs-only change - dashboard validation not required."
+            exit 0
+          fi
+          if [ "${{ needs.dashboard-build.result }}" != "success" ]; then
+            echo "dashboard validation failed (build=${{ needs.dashboard-build.result }})"
+            exit 1
+          fi
```
0xgleb, line 334: minor: this gate and the `hooks` one are copies of the pre-existing `backend` gate, same `runs-on`, same `always()`, same single check step, same `!= "success" && exit 1` shape, and in two of three the same docs-only short circuit. One pattern expressed three times with no shared source. These jobs are the entire load bearing mechanism for branch protection, so drift here is silent: a future gate that forgets the `!= "success"` arm, or adds a work job to `needs` in one copy but not the others, publishes a green required check for unvalidated code. Extracting it into a composite action taking the dependency results and the docs-only flag as inputs would also let the harness verify one script instead of three, and a fourth required check would inherit the semantics by construction.

scripts/test-fork-pr-ci.sh
```diff
@@ -36,25 +45,210 @@ assert_count 1 '    branches: [master]'
...
+  cases.each do |values, expected_success|
+    rendered = expressions.zip(values).reduce(gate_script) do |script, (expression, value)|
+      script.gsub("${{ #{expression} }}", value)
+    end
+    if rendered.include?("$" + "{{")
+      warn "#{job_name} gate has unrendered expressions"
+      exit 1
+    end
+
+    _stdout, _stderr, status = Open3.capture3("bash", "-c", rendered)
```
0xgleb, line 187: minor: the runner executes `run:` steps with `bash --noprofile --norc -eo pipefail`, but this simulates them under plain `bash -c`. The current gate scripts are simple enough that the two agree, but the harness exists to keep that true as the scripts change, and a gate that grows a pipeline or a command failing before the final `if` would simulate as passing here and fail in CI, or the reverse. Passing the same flags through to `Open3.capture3` keeps the simulation honest.

scripts/test-fork-pr-ci.sh
```diff
@@ -36,25 +45,210 @@ assert_count 1 '    branches: [master]'
...
-  echo "CI permissions must remain read-only" >&2
-  exit 1
-fi
+# Every subscribed pull request event must leave behind a run that executes the
+# required checks. An optimizer-controlled job-level skip is unsafe because
+# GitHub reports a skipped required job as successful. A replacement event may
+# cancel an older run, but the replacement must use the same concurrency group
+# and perform validation itself.
+assert_absent 'withgraphite/graphite-ci-action@'
+assert_absent 'GRAPHITE_CI_OPTIMIZATION_TOKEN'
```
0xgleb, line 54: minor: this guard stops the optimizer coming back in `ci.yaml`, but `GRAPHITE_CI_OPTIMIZATION_TOKEN` itself is still configured in repository settings and is still a valid Graphite credential with no consumer. A stored secret nothing legitimate reads is standing exposure that no one would notice being abused. Worth revoking it in Graphite and deleting it from Actions secrets as part of landing this, and calling that out in the description so whoever has admin does it.

docs/ci.md
```diff
@@ -13,10 +13,24 @@ commit. A skipped or stale duplicate could then hide the result that branch
...
+GitHub emits the same event for title and body edits, which change no code. All
+subscribed events for a pull request share one concurrency group, so a later
+event cancels and replaces an older run. The replacement performs the same
+validation even when only PR metadata changed. This avoids a race where a head
+or base update is cancelled and the surviving run reports the required jobs as
+skipped, which GitHub treats as successful.
+
+The required checks are never gated by the Graphite CI optimizer. In particular,
+middle-of-stack PRs run their own backend validation rather than relying on a
+different stack branch. Only a confirmed docs-only diff short-circuits the
```
0xgleb, line 25: minor: this argues against the Graphite CI optimizer, which this same PR removes and which is now referenced nowhere in the repo. A future reader hits a dangling reference and a sentence rebutting something they can't see, and the second sentence only parses once you already know the optimizer used to skip middle-of-stack PRs. Stating the positive fact instead reads as a standing record: every PR runs its own backend, dashboard, and hooks validation, no external service can stand a required check down, and only a confirmed docs-only diff short circuits the toolchains.

README.md
```diff
@@ -289,7 +289,8 @@ nix run .#deployAll   # first deployment
 ### CI/CD
 
 - **CI** (`.github/workflows/ci.yaml`): Builds all packages, runs tests and
-  clippy inside nix derivations, builds dashboard. Runs on every push.
+  clippy inside nix derivations, and builds the dashboard. Runs for pull request
+  activity and pushes to `master`.
```
0xgleb, line 293: nit: the `on:` block also declares `workflow_dispatch`, and docs/ci.md discusses reserving it for manual runs, so this bullet under-describes the trigger set it was just corrected to describe. Adding manual `workflow_dispatch` keeps the two documents in agreement.

0xgleb (approved):

PR #1112, opened 2026-08-03
fix: attribute Bebop fills to the Bebop venue

findolor:
## What

Closes RAI-1601.

Attributes shared-inventory fills to their configured onchain venue throughout the counter-trades window. The known Bebop and Uniswap v4 adapters now display as their actual venues, direct orderbook fills remain Raindex, and unconfigured operators display as Unknown Onchain instead of being misattributed.

## Why

Inventory settlements previously discarded their operator before the onchain trade was persisted. The dashboard therefore hardcoded every onchain fill as Raindex, including the Bebop fill from transaction `0x28a364eaa0fc6d505ddd0702cd9949853193bb28c92cdfef68bcd5a803f1806b` at log index `194`.

## How

- Configure public inventory-adapter operator addresses as explicit venue/operator pairs under `[raindex]`.
- Persist the execution source on each onchain trade and derive its venue consistently for REST history, live WebSocket updates, venue filtering, sorting, explorer links, and trade details.
- Preserve legacy source-less events, and let `process-tx` safely attribute already-accounted historical fills without placing another hedge, including upgrading an unknown inventory source when its matching adapter is later configured.
- Introduce `terminal_outcomes_v3` for the expanded venue model while retaining v2/v1 downgrade behavior for older clients and servers.
- Re-probe v3 after a temporary fallback, skip future unknown dashboard venues without reconnect loops, and keep legacy-compatible venue filters correct after downgrade.
- Retry transient source-correction reloads through a supervised handoff monitor, serialize authoritative publication, and retain pending corrections across monitor restarts.
- Update the specification, domain terminology, deployment configuration, Rust DTOs, and dashboard models.

## Testing

- [x] Full local CI: workspace checks with and without all features, 4,241 tests passed with 6 configured skips, workspace Clippy, and rustfmt
- [x] DTO generation and freshness check
- [x] Dashboard install, lint, and Svelte diagnostics
- [x] Focused regression coverage for attribution, historical repair, protocol downgrade/re-probe, venue filtering, future venues, explorer behavior, source-correction retries, publication ordering, retry-queue saturation, and shared SQLite pool capacity
- [x] Simplification pass plus three-round staged read-only review

## Screenshots

Validated in the deterministic market simulator with a Bebop-only filter showing the expected AAPL fill.

Bebop-only counter-trades filter

## Anything else

No environment variable or secret is needed. Adapter operator addresses are public onchain deployment configuration: staging declares the known Bebop and Uniswap v4 adapters, while legacy production configuration keeps the adapter list empty.

## New Features
- Added support for Bebop, Uniswap V4, and unidentified on-chain venues in trade history.
- On-chain fills now display their attributed venue, including adapter-routed trades.
- Added venue-aware filtering, labels, colors, and transaction links.
- Added a newer trade-history protocol with automatic fallback for older clients.

## Bug Fixes
- Improved ordering for on-chain trades across supported venues.
- Preserved legacy fills and historical compatibility.
- Unsupported future venues are skipped without disrupting valid trade history.

## Documentation
- Updated guidance for venue attribution and adapter configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,8 +92,13 @@ The `[raindex]` section requires an explicit `inventory_mode` (`"legacy"` or
 `"managed"`) and a `vault_owner` address (the on-chain owner the vaults are
 keyed by; no fallback). `"managed"` additionally requires an `inventory` address
 (the shared `RaindexInventory` the bot operates via `OPERATOR_ROLE`) and is
-forbidden from being set under `"legacy"`. See the `[raindex]` block in
-`example.config.toml` for the full field documentation.
+forbidden from being set under `"legacy"`. Its required `inventory_adapters`
+table maps public operator addresses to venues such as Bebop and Uniswap v4;
+unknown operators remain visibly unattributed. This deployment metadata belongs
+in plaintext config, not secrets or environment variables. Trade protocol v3
+preserves configured and unknown onchain venues; older protocol versions
+collapse adapter and Unknown Onchain venues to Raindex for compatibility. See
+the `[raindex]` block in `example.config.toml` for the full field documentation.
 
 Current broker support is limited to `alpaca-broker-api` and `dry-run`.
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -120,6 +120,13 @@ explicitly per deployment via `inventory_mode` under `[raindex]`:
   clear error rather than reverting on the first rebalance; it also revokes any
   stale pre-migration allowance the bot granted the OrderBook directly.
 
+Each deployment explicitly configures its inventory adapters under `[raindex]`
+as venue/operator pairs. These public onchain addresses are deployment settings,
+not secrets or environment variables. The same venue may list more than one
+operator during a rotation, but one operator cannot name multiple venues. Legacy
+mode forbids adapter entries because it has no shared inventory through which
+they could settle.
+
 Distinct from the settlement target is `vault_owner`: the on-chain owner every
 `vaultBalance2` read, vault-registry entry, and ClearV3/TakeOrderV3 order-owner
 fill match is scoped by. It is required and explicit (no fallback). While the
@@ -4507,24 +4514,43 @@ multiple broker-specific contexts.
    reference, spread bps) and per-symbol price charts over time.
 
 4. **Trade History**: Recent trades filterable by venue (onchain/offchain/both).
-   Onchain entries are completed fills. Offchain counter-trade entries include
-   successful fills, terminal failures, and terminal cancellations; each entry
-   carries its terminal outcome timestamp. Failed entries expose the broker or
-   placement error. Failed and cancelled entries keep the requested quantity
-   distinct from the broker-accepted quantity. The accepted and actually-filled
-   quantities are nullable provenance: `null` means the persisted history does
-   not prove that fact, including legacy outcomes that predate explicit fill
-   evidence. An explicit broker-reported zero therefore distinguishes a wholly
-   unfilled cancellation from an outcome whose fill remains unknown. Remaining
-   and excess quantities are derived only when both accepted and filled
-   quantities are known. An overfill reports the complete actual fill plus its
-   excess rather than clamping the fill to the accepted quantity. Non-terminal
-   submitted, partially-filled, and cancelling orders remain excluded. Terminal
-   outcomes use the same provenance in initial history and live dashboard
-   updates. A broker-terminal failure or cancellation with a positive fill but
-   no execution price is still persisted and surfaced; the owning position
-   remains pending for manual reconciliation so the unpriced fill cannot be
-   silently applied or discarded.
+   Onchain entries are completed fills and carry their execution source:
+   `ClearV3`/`TakeOrderV3` fills are Raindex, while an `InventoryTrade` persists
+   the raw inventory `operator` and resolves it through the deployment's
+   configured adapter table. Recognized operators are reported as their venue
+   (currently Bebop or Uniswap v4). An operator absent from the table is
+   persisted and logged as unrecognized, and its trade is reported as the
+   distinct Unknown Onchain venue rather than being asserted to be Raindex.
+   Initial history and live WebSocket updates derive the venue from the same
+   persisted source, and venue filtering includes every exposed adapter venue.
+   Legacy `OnChainTradeEvent::Filled` events that predate source persistence
+   retain a legacy source marker and display as Raindex for compatibility. The
+   event store alone cannot reconstruct their discarded operator. Reprocessing
+   the transaction through the existing `process-tx` recovery path reads the
+   operator from chain and appends a source-attribution event through CQRS,
+   repairing history without repeating an already-acknowledged hedge. The same
+   repair upgrades an Unknown Onchain fill when its persisted operator is later
+   configured, but cannot replace it with an adapter carrying a different
+   operator.
+
+   Offchain counter-trade entries include successful fills, terminal failures,
+   and terminal cancellations; each entry carries its terminal outcome
+   timestamp. Failed entries expose the broker or placement error. Failed and
+   cancelled entries keep the requested quantity distinct from the
+   broker-accepted quantity. The accepted and actually-filled quantities are
+   nullable provenance: `null` means the persisted history does not prove that
+   fact, including legacy outcomes that predate explicit fill evidence. An
+   explicit broker-reported zero therefore distinguishes a wholly unfilled
+   cancellation from an outcome whose fill remains unknown. Remaining and excess
+   quantities are derived only when both accepted and filled quantities are
+   known. An overfill reports the complete actual fill plus its excess rather
+   than clamping the fill to the accepted quantity. Non-terminal submitted,
+   partially-filled, and cancelling orders remain excluded. Terminal outcomes
+   use the same provenance in initial history and live dashboard updates. A
+   broker-terminal failure or cancellation with a positive fill but no execution
+   price is still persisted and surfaced; the owning position remains pending
+   for manual reconciliation so the unpriced fill cannot be silently applied or
+   discarded.
 
    Dashboard trade messages remain compatible during rolling deploys. The
    canonical protocol sends every terminal outcome as `trade_update`. Filled
@@ -4544,13 +4570,18 @@ multiple broker-specific contexts.
    `trade_protocol=terminal_outcomes_v2`. Cancelled outcomes are also v2-only,
    because adding a new tagged outcome to v1 would break older exhaustive
    consumers; a v1 fallback therefore omits cancelled entries from history
-   rather than losing them. A newer dashboard requests v2 and retries both
-   WebSocket and HTTP history with v1 when an older backend rejects the unknown
-   protocol value, while continuing to accept v1 responses and reconstructing
-   the complete fill as the legacy filled plus excess quantity. A legacy zero
-   fill remains unknown because v1 synthesized zero when no fill evidence
-   existed. A successful v1 WebSocket fallback probes v2 again after its next
-   disconnect so a transient restart cannot pin the client to v1. Failed
+   rather than losing them. Adapter-specific venue values use
+   `trade_protocol=terminal_outcomes_v3`; legacy, v1, and v2 responses downgrade
+   every adapter or unknown-onchain venue to Raindex so an older exhaustive
+   dashboard cannot enter a parse/reconnect loop. A newer dashboard requests v3
+   and retries v2, then v1, when an older backend rejects an unknown protocol.
+   Legacy-protocol filter requests likewise collapse selected adapter venues to
+   Raindex so an old backend does not reject the venue query. The dashboard
+   continues to accept v1 responses and reconstructs the complete fill as the
+   legacy filled plus excess quantity. A legacy zero fill remains unknown
+   because v1 synthesized zero when no fill evidence existed. A successful
+   lower-version WebSocket fallback probes v3 again after its next disconnect so
+   a transient restart cannot pin the client to an older protocol. Failed
    outcomes are sent only through a terminal-outcome protocol. Trade ordering
    compares the complete RFC 3339 timestamp, including sub-millisecond
    precision, then uses the same stable trade-ID tie-breaker for initial history
@@ -4561,13 +4592,15 @@ multiple broker-specific contexts.
    the same validation rules for the canonical and legacy wire shapes:
    timestamps must be valid UTC RFC 3339 values, venues/directions/outcomes must
    be known variants, the total share quantity must be positive, and outcome
-   quantities must be non-negative decimal strings. An invalid payload is
-   reported visibly and leaves the last known-good trade state unchanged. The
-   remaining domains from an otherwise valid snapshot still refresh. An invalid
-   WebSocket payload closes the connection, shows the reconnect banner, and
-   retries with bounded exponential backoff. The Rust trade DTO encodes the
-   positive total-quantity invariant so an invalid trade cannot be constructed
-   or deserialized at the server boundary.
+   quantities must be non-negative decimal strings. A trade carrying a future,
+   unknown venue is skipped with a warning so a rolling dashboard deployment can
+   continue showing compatible trades without reconnecting. Any other invalid
+   payload is reported visibly and leaves the last known-good trade state
+   unchanged. The remaining domains from an otherwise valid snapshot still
+   refresh. Any other invalid WebSocket payload closes the connection, shows the
+   reconnect banner, and retries with bounded exponential backoff. The Rust
+   trade DTO encodes the positive total-quantity invariant so an invalid trade
+   cannot be constructed or deserialized at the server boundary.
 
    Terminal trade live updates are delivered through a persistent delivery
    ledger and job queue, not directly from the CQRS reactor. The ledger is keyed
@@ -4584,7 +4617,13 @@ multiple broker-specific contexts.
    permitted: the dashboard merges trade updates by trade ID, making a retry or
    crash replay idempotent from the operator's point of view. Publishing with no
    connected WebSocket receivers is successful because the next connection
-   obtains the same terminal outcome from its initial-state snapshot.
+   obtains the same terminal outcome from its initial-state snapshot. A later
+   source-attribution correction reloads the authoritative onchain-trade history
+   and broadcasts a replacement update. Transient reload failures receive
+   bounded retries; exhaustion restarts the handoff monitor, whose startup
+   reconciliation reloads every pending source correction retained by that
+   monitor. After a process restart, clients reconnect from the
+   already-corrected snapshot.
 
 5. **Live Events**: Real-time domain event stream (aggregate type, ID, sequence,
    event type, timestamp). Starts empty, populates via WebSocket.
--- a/crates/config/src/evm.rs
+++ b/crates/config/src/evm.rs
@@ -1,5 +1,7 @@
 //! EVM configuration types: chain config, RPC secrets, and runtime context.
 
+use std::collections::HashSet;
+
 use alloy::primitives::Address;
 use serde::Deserialize;
 use thiserror::Error;
@@ -71,6 +73,69 @@ pub enum InventoryModeTag {
     Managed,
 }
 
+/// Venue identified by an inventory adapter's `operator` address.
+#[derive(Debug, Clone, Copy, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum InventoryAdapterVenue {
+    Bebop,
+    UniswapV4,
+}
+
+/// One deployment-specific shared-inventory adapter.
+#[derive(Debug, Clone, Copy, Deserialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub struct InventoryAdapter {
+    pub venue: InventoryAdapterVenue,
+    pub operator: Address,
+}
+
+/// Explicit inventory-adapter table configured for one deployment.
+#[derive(Debug, Clone, Default, Deserialize, PartialEq, Eq)]
+#[serde(try_from = "Vec<InventoryAdapter>")]
+pub struct InventoryAdapters(Vec<InventoryAdapter>);
+
+impl TryFrom<Vec<InventoryAdapter>> for InventoryAdapters {
+    type Error = EvmConfigError;
+
+    fn try_from(adapters: Vec<InventoryAdapter>) -> Result<Self, Self::Error> {
+        Self::try_new(adapters)
+    }
+}
+
+impl InventoryAdapters {
+    /// Builds and validates an explicit adapter table.
+    pub fn try_new(adapters: Vec<InventoryAdapter>) -> Result<Self, EvmConfigError> {
+        let adapters = Self(adapters);
+        adapters.validate()?;
+        Ok(adapters)
+    }
+
+    /// Returns the venue configured for `operator`, if any.
+    pub fn venue_for(&self, operator: Address) -> Option<InventoryAdapterVenue> {
+        self.0
+            .iter()
+            .find(|adapter| adapter.operator == operator)
+            .map(|adapter| adapter.venue)
+    }
+
+    fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    fn validate(&self) -> Result<(), EvmConfigError> {
+        let mut operators = HashSet::new();
+        for adapter in &self.0 {
+            if !operators.insert(adapter.operator) {
+                return Err(EvmConfigError::DuplicateInventoryAdapterOperator {
+                    operator: adapter.operator,
+                });
+            }
+        }
+
+        Ok(())
+    }
+}
+
 /// Errors resolving an [`EvmConfig`] into a runtime [`EvmCtx`].
 #[derive(Debug, Error)]
 pub enum EvmConfigError {
@@ -85,6 +150,16 @@ pub enum EvmConfigError {
          enable a shared inventory"
     )]
     LegacyWithInventory { inventory: Address },
+    #[error(
+        "[raindex] inventory_mode = \"legacy\" forbids inventory adapters because \
+         no shared inventory exists"
+    )]
+    LegacyWithInventoryAdapters,
+    #[error(
+        "[raindex].inventory_adapters configures operator {operator} more than once; \
+         one operator cannot identify multiple venues"
+    )]
+    DuplicateInventoryAdapterOperator { operator: Address },
 }
 
 #[derive(Debug, Deserialize)]
@@ -104,6 +179,9 @@ pub struct EvmConfig {
     /// events on the pooled vaults are also surfaced here as
     /// `OperatorDeposit`/`OperatorWithdraw`.
     pub inventory: Option<Address>,
+    /// Public, deployment-specific adapter operator addresses used to attribute
+    /// shared-inventory settlements to their execution venue.
+    pub inventory_adapters: InventoryAdapters,
     /// Address that owns the Raindex orders and vaults on-chain -- the key every
     /// `vaultBalance2` read, vault-registry entry, and order-owner fill match is
     /// scoped by. Required and explicit (no fallback): this parameter determines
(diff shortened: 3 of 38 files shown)


findolor:
* **#1112** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1112?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1112?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1112).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/dto/src/trade.rs
```diff
@@ -21,10 +22,17 @@ impl TradingVenue {
...
             Self::Alpaca => "alpaca",
             Self::DryRun => "dry_run",
         }
     }
+
+    /// Whether trades from this venue settle onchain.
+    #[must_use]
+    pub const fn is_onchain(self) -> bool {
+        matches!(self, Self::Raindex | Self::Bebop)
+    }
```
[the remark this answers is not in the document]
findolor: Addressed in f48a29c0: `TradingVenue::is_onchain` now uses an exhaustive match with explicit onchain and offchain variants.

src/onchain_trade.rs
```diff
@@ -280,6 +290,27 @@ impl OnChainTrade {
...
+    pub(crate) fn trading_venue(self) -> TradingVenue {
+        match self {
+            Self::Inventory {
+                operator: BEBOP_OPERATOR,
+            } => TradingVenue::Bebop,
+            Self::Raindex | Self::Inventory { .. } => TradingVenue::Raindex,
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
findolor: Addressed in f48a29c0: first ingestion of an unknown inventory operator now emits a warning including the operator before retaining the Raindex fallback.

src/onchain/trade.rs
```diff
@@ -535,6 +537,9 @@ impl OnchainTrade {
             pyth_feed_ids,
             trade_details,
             TradeMetadata {
+                source: OnChainTradeSource::Inventory {
+                    operator: trade.deposit.operator,
+                },
```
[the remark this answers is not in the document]
findolor: Addressed in f48a29c0: the source assignment now documents that pairing/quarantine guarantees matching deposit and withdrawal operators.

src/onchain_trade.rs
```diff
@@ -280,6 +300,37 @@ impl OnChainTrade {
...
+    fn unrecognized_inventory_operator(self) -> Option<Address> {
+        match self {
+            Self::Raindex
+            | Self::Inventory {
+                operator: BEBOP_OPERATOR,
+            } => None,
+            Self::Inventory { operator } => Some(operator),
+        }
+    }
```
0xgleb, line 377: should fix: this maps only `BEBOP_OPERATOR` and folds every other inventory operator into `TradingVenue::Raindex`. The description calls those unknown operators, but a second one is already known and pinned in this repo: `try_from_inventory_trade_real_univ4_fill_is_buy` in src/onchain/trade.rs is a real prod settlement whose operator is `0x36ebb1e5149c60111dd035f0417a4b00d39caa88`, and the `RaindexTradeEvent` doc comment names the univ4 hook as an active OPERATOR_ROLE holder alongside Bebop. So univ4-routed fills ship labelled Raindex, which is the same mis-attribution this PR was opened to fix. The fallback also conflates two different facts, because a direct ClearV3/TakeOrderV3 fill and an unattributable adapter settlement become indistinguishable in history, filtering, and reconciliation. Separately, `trading_venue` and `unrecognized_inventory_operator` each hard-code the recognized-operator set independently, so adding an adapter means editing both, and editing only one leaves either a wrong venue or a permanent warning. Pulling the mapping into a single `adapter_venue(operator) -> Option<TradingVenue>` and deriving both from it makes a new adapter one line and stops the two disagreeing. Returning a distinct onchain venue for an unrecognized operator, rather than Raindex, would also stop asserting something demonstrably false.

dashboard/src/lib/trade-payload.ts
```diff
@@ -115,7 +116,8 @@ const parseTimestamp = (value: unknown, path: string): string => {
 }
 
 const parseVenue = (value: unknown, path: string): TradingVenue => {
-  if (value === 'raindex' || value === 'alpaca' || value === 'dry_run') return value
+  if (isTradingVenue(value)) return value
+
   return invalid(path, 'a known trading venue')
```
0xgleb: should fix: `parseVenue` throws for any venue string it doesn't recognise, and the batch parsers fan that out to the whole payload with no per-entry recovery. In the WebSocket path a throw inside `handleMessage` is caught by `tryCatch` and turned into `fsm.send('error')`, which tears the socket down and reconnects. Since `current_state` reseeds recent trades on every connect, a tab still running the pre-deploy bundle hits this on every reconnect for as long as a Bebop fill sits in the seed window, so it's a permanent error/reconnect loop rather than a one-off glitch. The REST history path fails the same way. The mirror case exists too: a new bundle sends `venue=raindex,bebop,alpaca` and an older backend's `TradingVenue::from_str` rejects `bebop` and 400s the whole request. SPEC.md around line 4541 states dashboard trade messages stay compatible during rolling deploys, and the `trade_protocol` v1/v2 negotiation only covers outcome-shape changes, not a new value in an existing field. Making an unrecognised venue non-fatal, either passing it through (venueLabel and venueColor already have default arms) or skipping the individual entry, would keep old tabs alive.

dashboard/src/lib/trade.ts
```diff
@@ -72,6 +72,19 @@ export type TradeHistoryFilter = {
...
+  'bebop',
+  'alpaca',
+  'dry_run'
+] as const satisfies readonly TradingVenue[]
+
+export const isTradingVenue = (value: unknown): value is TradingVenue =>
+  typeof value === 'string' && (TRADING_VENUES as readonly string[]).includes(value)
+
+export const isOnchainVenue = (venue: TradingVenue): boolean =>
+  venue === 'raindex' || venue === 'bebop'
```
0xgleb: should fix: `as const satisfies readonly TradingVenue[]` only checks that every listed element is a venue, not that every venue is listed, and `isOnchainVenue` is a hand-written pair of comparisons duplicating `TradingVenue::is_onchain`, which the Rust compiler does enforce exhaustively. Adding a fifth venue to the Rust enum type-checks cleanly here while `isTradingVenue` rejects it, the filter dropdown omits it, `buildParams` computes the all-selected threshold from the wrong length, and `matchesHistoryFilter` drops live updates for it. A `type _AllVenuesListed = AssertNever<Exclude<TradingVenue, (typeof TRADING_VENUES)[number]>>` next to the list, plus rewriting `isOnchainVenue` as a total `Record<TradingVenue, boolean>` lookup, turns the next venue into a build error instead of silently hidden fills.

dashboard/src/lib/trade.test.ts
```diff
@@ -211,6 +211,21 @@ describe('live trade history', () => {
...
+
+    expect(
+      mergeTradeHistory([raindexLog2], [bebopLog10], {
+        venues: new Set(['raindex', 'bebop']),
+        symbols: new Set(),
+        since: null,
+        until: null
+      })
+    ).toEqual([bebopLog10, raindexLog2])
+  })
```
0xgleb, line 253: should fix: this passes with `isOnchainVenue` reverted, so it doesn't test the change it was added for. Log indices 2 and 10 are the problem: the numeric branch sorts descending (10 first), and the `compareStrings` fallback sorts ascending, where `:10` < `:2` also puts 10 first. Both paths produce the asserted `[bebopLog10, raindexLog2]`. Under the old `left.venue === 'raindex' && right.venue === 'raindex'` predicate the mixed pair skips the numeric branch entirely and still lands on the same answer. Using 11 and 20 like the Rust test does discriminate, because ascending string order puts `:11` first while descending numeric puts `:20` first, so asserting `[bebopLog20, raindexLog11]` fails if the predicate regresses.

src/api.rs
```diff
@@ -653,6 +653,18 @@ impl TradeFilter {
...
+        self.venues
+            .as_ref()
+            .is_none_or(|venues| venues.iter().any(|venue| venue.is_onchain()))
+    }
+
+    fn includes_offchain(&self) -> bool {
+        self.venues
+            .as_ref()
+            .is_none_or(|venues| venues.iter().any(|venue| !venue.is_onchain()))
+    }
```
0xgleb, line 686: should fix: these replace the old `matches_venue(Raindex)` and `matches_venue(Alpaca) || matches_venue(DryRun)` checks in `load_trade_rows`, and nothing exercises them now that two venues map to onchain. The one new Bebop filtering test calls `parse_onchain_trade_row` directly and never reaches `includes_onchain`/`includes_offchain`. A test seeding a Bebop-sourced OnChainTrade and calling `load_trade_rows` with a Bebop-only filter, plus an Alpaca-only filter asserting the onchain trade is excluded, would cover the gate itself.

dashboard/src/lib/components/trade-history-panel.svelte
```diff
@@ -642,7 +651,7 @@
         <div class="mb-4 space-y-1.5 text-xs">
           <div class="flex items-center gap-2 font-mono">
             <span class="shrink-0 text-muted-foreground">ID</span>
-            {#if trade.venue === 'raindex'}
+            {#if isOnchainVenue(trade.venue)}
```
0xgleb, line 670: minor: switching this from `trade.venue === 'raindex'` to `isOnchainVenue(trade.venue)` means Bebop trades should now render as a clickable explorer link rather than a plain ID, but nothing in the diff opens a detail dialog to check it. A test that opens a Bebop trade and asserts the anchor with the expected `getExplorerTxUrl` href, plus one for a non-onchain trade asserting no link, would cover both arms.

dashboard/src/lib/components/trade-history-panel.test.ts
```diff
@@ -141,6 +141,32 @@ describe('TradeHistoryPanel', () => {
...
+    const { target } = mountPanel()
+
+    await vi.waitFor(() => {
+      expect(target.textContent).toContain('COIN')
+      expect(target.textContent).toContain('Bebop')
+      expect(target.textContent).toContain('1 of 1')
+    })
+
+    target.remove()
+  })
```
0xgleb: minor: the name says it offers Bebop venue filtering, but the body only mounts the panel with one Bebop trade and asserts the row renders. It never opens the filter dropdown, toggles the Bebop option, or checks the fetch params. Either rename it to describe what it asserts, or extend it to toggle the filter and assert `bebop` appears in or drops out of the request.

src/onchain_trade.rs
```diff
@@ -8,18 +8,21 @@ use std::num::ParseIntError;
...
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+use tracing::warn;
 
 use st0x_dto::{Direction, Trade, TradeOutcome, TradingVenue};
 use st0x_event_sorcery::{DomainEvent, EventSourced, Nil};
 use st0x_execution::Symbol;
 use st0x_finance::{FractionalShares, NotPositive, Positive};
 
+const BEBOP_OPERATOR: Address = address!("0x8b8b6e0507c125934c6129563f48e48c66f86475");
```
0xgleb: minor: this is a third-party, deployment-specific address baked into the binary with no citation. The only in-repo proof it is Bebop's OPERATOR_ROLE holder is a test in another module (`try_from_inventory_trade_real_bebop_fill_is_sell`, documented against a captured Base tx), and nothing here points at it. Every other address in this integration is configuration: `crates/config/src/evm.rs` takes `inventory` from `[raindex]`, and docs/shared-inventory-cutover.md treats the Bebop hook as a per-deployment placeholder granted OPERATOR_ROLE during cutover, with a redeploy tracked as RAI-1198. There's no operator field in config at all, so the mapping can't follow a rotation, a redeploy, or a second granted address, and staging and prod can't differ. Citing the provenance at the constant (matching the comment style above `REAL_USDC_BASE`) and sourcing adapter operators from `[raindex]` next to `inventory` would make a hook redeploy a config change rather than a rebuild.

SPEC.md
```diff
@@ -4507,24 +4507,36 @@ multiple broker-specific contexts.
...
+   `ClearV3`/`TakeOrderV3` fills are Raindex, while an `InventoryTrade` persists
+   the raw inventory `operator` and maps a recognized adapter operator to its
+   venue (currently Bebop). Persisting the raw operator keeps new adapters
+   attributable without changing hedge or position accounting. Initial history
+   and live WebSocket updates derive the venue from the same persisted source,
+   and venue filtering includes every exposed adapter venue. Legacy
+   `OnChainTradeEvent::Filled` events that predate source persistence
+   deserialize as Raindex. Their inventory operator was discarded before
+   persistence and cannot be reconstructed from the event store, so they remain
+   attributed to Raindex and are not retroactively corrected.
```
0xgleb: minor: this says an InventoryTrade maps a recognized adapter operator to its venue and that legacy source-less events deserialize as Raindex, but never states what happens to a recognized-shape settlement whose operator isn't in the table. A reader can't tell whether it is dropped, errored, shown as unknown, or reported as Raindex, and it's the last of those, which is the one with reconciliation consequences. Worth one sentence stating that and noting it's logged. The other gap is the motivating fill: `0x28a364...:194` is already persisted without a source, so it still displays as Raindex after this ships. The justification that it cannot be reconstructed is true of the event store but not of chain data, since the operator is still readable and `pair_inventory_settlements` already does it for the `process-tx` recovery path. Either narrow that sentence to say the event store alone can't recover it and point at a tracked backfill, or add the repair path.

docs/domain.md
```diff
@@ -94,11 +94,12 @@ and confusing.
...
-  equities and USDC are traded via limit orders placed and priced externally.
-  The bot monitors fills on those orders and manages vault balances
-  (deposits/withdrawals) for rebalancing; it does not add, remove, or reprice
-  the orders themselves.
+- **Onchain venue**: Tokenized-equity execution on Base. Direct `ClearV3` and
+  `TakeOrderV3` fills execute on Raindex. Settlements routed through the shared
+  `RaindexInventory` retain their adapter operator so recognized adapters such
+  as Bebop are reported as their own venue. The bot monitors these fills and
+  manages vault balances (deposits/withdrawals) for rebalancing; it does not
+  add, remove, or reprice the orders themselves.
```
0xgleb: minor: this is the terminology source of truth, but the rewritten bullet introduces `ClearV3`, `TakeOrderV3`, `RaindexInventory`, adapter operator, and recognized adapters without defining any of them, and none are defined elsewhere in the file. It also drops the sentence that actually defined the venue, so a reader can no longer learn what Raindex is from the glossary, only that fills execute on it. Restoring that definition and stating the adapter case in domain terms, with the contract event names left in SPEC.md where the code-level shapes belong, would keep this readable standalone.

src/performance.rs
```diff
@@ -238,6 +238,7 @@ pub async fn seed_simulated_hedge_latency_history(
                 .send(
                     &OnChainTradeId { tx_hash, log_index },
                     OnChainTradeCommand::WitnessAt {
+                        source: crate::onchain_trade::OnChainTradeSource::Raindex,
```
0xgleb: nit: `crate::onchain_trade::OnChainTradeSource::Raindex` is written fully qualified at every new call site here and across trade_loader.rs, dashboard/event.rs, conductor.rs, test_utils.rs, and the api.rs tests, even though those modules already import their siblings from the same module at the top. Nothing is ambiguous, so adding `OnChainTradeSource` to the existing `use crate::onchain_trade::{...}` group reads consistently with the surrounding code.

0xgleb (changes requested):

src/onchain_trade.rs
findolor: Fixed. Adapter attribution now comes from explicit `[raindex].inventory_adapters` venue/operator configuration. Bebop and Uniswap v4 have distinct venues, unconfigured operators persist as Unknown Onchain, and direct orderbook fills remain Raindex. One configuration-backed lookup drives attribution and warnings, with regression coverage.

dashboard/src/lib/trade-payload.ts
findolor: Fixed. `terminal_outcomes_v3` carries the expanded venue model; older protocols receive legacy-compatible venue downgrades. The dashboard negotiates v3 to v2 to v1, periodically re-probes v3 after REST fallback, re-probes after WebSocket fallback, and skips only a future unknown venue instead of entering a reconnect loop.

dashboard/src/lib/trade.ts
findolor: Fixed. The dashboard venue list now has a compile-time exhaustiveness assertion, and venue classification uses a total `Record<TradingVenue, boolean>`. Adding a venue without updating either mapping is now a type error.

dashboard/src/lib/trade.test.ts
findolor: Fixed. The mixed onchain sorting regression now uses log indices 11 and 20 and asserts 20 before 11, so reverting onchain classification makes the test fail.

src/api.rs
findolor: Fixed. API tests now exercise `load_trade_rows` with a Bebop-only filter and with an Alpaca-only filter, covering both the onchain inclusion gate and onchain exclusion for offchain-only queries.

dashboard/src/lib/components/trade-history-panel.svelte
findolor: Fixed. Component tests open a Bebop trade detail and assert its explorer transaction link, and separately verify that an offchain trade ID is not rendered as an explorer link.

dashboard/src/lib/components/trade-history-panel.test.ts
findolor: Fixed. The test now exercises the venue filter controls and verifies the resulting Bebop request behavior instead of only asserting that a seeded row renders.

src/onchain_trade.rs
findolor: Fixed. Operator addresses are no longer compiled constants. They are public, deployment-specific `[raindex].inventory_adapters` entries, validated for duplicate operators and forbidden in legacy inventory mode. Staging declares the known Bebop and Uniswap v4 operators; production legacy mode keeps the list empty.

SPEC.md
findolor: Fixed. SPEC now defines configured and unrecognized inventory sources explicitly. Historical repair is implemented through `process-tx`: it reconstructs the receipt source and emits `SourceAttributed` for an already-accounted fill without accounting or hedging it a second time. API and live updates fold and broadcast the corrected attribution.

docs/domain.md
findolor: Fixed. The glossary again defines Raindex in domain terms and describes shared-inventory venue adapters without relying on contract event names; the code-level ClearV3, TakeOrderV3, and inventory event details remain in SPEC.

src/performance.rs
findolor: Fixed. `OnChainTradeSource` is imported through each module's existing crate-internal import group and referenced directly at the call sites.

crates/config/src/evm.rs
```diff
@@ -71,6 +73,61 @@ pub enum InventoryModeTag {
...
+            if !operators.insert(adapter.operator) {
+                return Err(EvmConfigError::DuplicateInventoryAdapterOperator {
+                    operator: adapter.operator,
+                });
+            }
+        }
+
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
findolor: Fixed. `InventoryAdapters` now deserializes through `TryFrom<Vec<InventoryAdapter>>`, so the duplicate-operator invariant is enforced during TOML deserialization instead of only at a later validation call. The regression test now asserts that deserializing duplicate operators fails immediately.

dashboard/src/lib/trade-payload.test.ts
```diff
@@ -39,6 +39,25 @@ describe('trade payload validation', () => {
...
+      expect(parseTrade(onchain)).toEqual(onchain)
+    }
+  )
+
+  it('skips only an entry whose venue is newer than this dashboard', () => {
+    const supported = validTrade({ id: 'supported' })
+    const unsupported = validTrade({ id: 'unsupported', venue: 'future_venue' })
+
+    expect(parseTradeEntries([unsupported, supported])).toEqual([supported])
+  })
```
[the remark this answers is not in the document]
findolor: Fixed. Added a dashboard regression test proving `parseTradeEntries` still rethrows non-venue parsing errors; only entries with an unsupported future venue are skipped.

README.md
```diff
@@ -92,8 +92,11 @@ The `[raindex]` section requires an explicit `inventory_mode` (`"legacy"` or
 `"managed"`) and a `vault_owner` address (the on-chain owner the vaults are
 keyed by; no fallback). `"managed"` additionally requires an `inventory` address
 (the shared `RaindexInventory` the bot operates via `OPERATOR_ROLE`) and is
-forbidden from being set under `"legacy"`. See the `[raindex]` block in
-`example.config.toml` for the full field documentation.
+forbidden from being set under `"legacy"`. Its required `inventory_adapters`
+table maps public operator addresses to venues such as Bebop and Uniswap v4;
+unknown operators remain visibly unattributed. This deployment metadata belongs
+in plaintext config, not secrets or environment variables. See the `[raindex]`
+block in `example.config.toml` for the full field documentation.
```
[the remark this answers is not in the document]
findolor: Fixed. README now documents the protocol-version exception: v3 preserves native and unknown onchain venues, while v2 and v1 intentionally collapse them to Raindex for backward compatibility.

src/cli/alpaca_wallet.rs
```diff
@@ -727,7 +727,7 @@ mod tests {
     use st0x_config::RebalancingCtx;
     use st0x_config::create_test_issuance_ctx;
     use st0x_config::{AssetsConfig, EquitiesConfig, LogLevel, TradingMode};
-    use st0x_config::{EvmCtx, IngestionCutoff, InventoryMode};
+    use st0x_config::{EvmCtx, IngestionCutoff, InventoryAdapters, InventoryMode};
```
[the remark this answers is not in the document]
findolor: Fixed. The `st0x_config` import is now in the workspace-import group, separated from crate-internal imports as required by the repository convention.

src/onchain_trade.rs
```diff
@@ -162,24 +172,35 @@ impl EventSourced for OnChainTrade {
...
+                    source,
+                    symbol,
+                    amount,
+                    direction,
+                    price_usdc,
+                    block_number,
+                    block_timestamp,
+                    filled_at: Utc::now(),
+                }])
+            }
```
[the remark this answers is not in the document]
findolor: Fixed. Source-attribution decisions are now modeled by an exhaustive `SourceAttributionDecision` ADT, and the conductor handles every decision variant explicitly instead of relying on non-exhaustive `matches!` checks.

src/onchain_trade.rs
```diff
@@ -280,6 +320,64 @@ impl OnChainTrade {
...
+    Legacy,
+    Raindex,
+    Inventory {
+        operator: Address,
+        venue: InventoryVenue,
+    },
+    UnrecognizedInventory {
+        operator: Address,
+    },
+}
```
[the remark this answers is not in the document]
findolor: Fixed. `process-tx` can now upgrade an already-accounted `UnrecognizedInventory` source to a configured `Inventory` source when the operator matches, without accounting or hedging again. A different operator remains rejected. SPEC and regression tests cover both paths.

src/onchain_trade.rs
```diff
@@ -884,6 +1048,136 @@ mod tests {
...
+                },
+            })
+            .await
+            .then_expect_error();
+
+        assert!(matches!(
+            error,
+            LifecycleError::Apply(OnChainTradeError::SourceAlreadyAttributed)
+        ));
+    }
```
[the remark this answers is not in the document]
findolor: Fixed. Added regression coverage proving `AttributeSource` rejects `Legacy` as the replacement source and emits no event.

crates/config/src/evm.rs
```diff
@@ -261,6 +350,113 @@ mod tests {
...
+        )
+        .unwrap_err();
+
+        assert!(
+            error.to_string().contains(
+                "configures operator 0x4444444444444444444444444444444444444444 more than once"
+            ),
+            "duplicate operators must fail before an EvmConfig can be constructed: {error}"
+        );
+    }
```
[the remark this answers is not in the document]
findolor: Fixed in 08da30f7. Added regression coverage that configures two distinct Bebop operators, constructs EvmCtx successfully, and verifies the second operator resolves to Bebop, preserving operator-only deduplication for rotations.

src/api.rs
```diff
@@ -748,40 +761,53 @@ async fn load_onchain_trade_rows(
...
+             WHERE attributed.aggregate_type = filled.aggregate_type \
+               AND attributed.aggregate_id = filled.aggregate_id \
+               AND attributed.event_type = 'OnChainTradeEvent::SourceAttributed' \
+             ORDER BY attributed.sequence DESC LIMIT 1) \
+         FROM events AS filled \
+         WHERE filled.event_type = 'OnChainTradeEvent::Filled' \
+         ORDER BY filled.rowid DESC",
     )
     .fetch_all(pool)
     .await?;
```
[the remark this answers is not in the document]
findolor: Resolved in 08da30f7 by removing the raw events-table query entirely. Project CQRS rules prohibit ad hoc event-store reads, so trade history now uses the typed event-sorcery aggregate loader with bounded concurrent replay. The correlated subquery and its proposed index are no longer present.

src/dashboard/event.rs
```diff
@@ -727,6 +727,38 @@ impl Broadcaster {
...
+            ),
+            Err(error) => warn!(
+                target: "dashboard",
+                %id,
+                ?error,
+                "Failed to replay source-attributed onchain trade for broadcast"
+            ),
+        }
+
+        Ok(())
```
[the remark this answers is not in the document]
findolor: Fixed in 08da30f7. Failed source-attribution reloads are tracked and routed through the supervised bounded handoff retry monitor, with authoritative reload/publication ordering and restart reconciliation for retained pending corrections. Regression coverage verifies initial failure followed by corrected delivery, retry exhaustion/restart recovery, stale-delivery ordering, and saturated-queue lock safety.

src/dashboard/mod.rs
```diff
@@ -67,15 +68,34 @@ impl TradeProtocol {
...
+        } else {
+            serde_json::to_value(trade)?
+        };
+
+        if self != Self::TerminalOutcomesV3 {
+            wire["venue"] = serde_json::to_value(trade.venue.legacy_compatible())?;
         }
+
+        Ok(wire)
     }
```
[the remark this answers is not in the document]
findolor: Fixed in 08da30f7. Rust serialization now exhaustively matches every TradeProtocol variant, and TypeScript fallback uses a total Record<TradeProtocol, TradeProtocol> mapping, so adding a protocol version requires both sides to be updated.

src/api.rs
```diff
@@ -666,9 +661,27 @@ impl TradeFilter {
...
+        self.venues.as_ref().is_none_or(|venues| {
+            venues.iter().any(|filter| {
+                if self.trade_protocol == TradeProtocol::TerminalOutcomesV3 {
+                    *filter == venue
+                } else {
+                    filter.legacy_compatible() == venue.legacy_compatible()
+                }
+            })
+        })
+    }
```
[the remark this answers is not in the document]
findolor: Fixed in 559a33b0. `matches_venue` now uses an exhaustive `TradeProtocol` match: v3 compares exact venues, while `LegacyFills`, v1, and v2 explicitly use legacy-compatible comparison.

0xgleb (approved):

PR #1137, opened 2026-08-06
fix: release the broker-idempotency anchor when the order is terminal

JuaniRios:
## What

- Releases the broker-idempotency anchor (`Position.last_failed_offchain_order_id`) when the Alpaca order under it is finished, so a symbol whose order expired can hedge again with no operator action. Closes [RAI-1559](https://linear.app/makeitrain/issue/RAI-1559/release-the-broker-idempotency-anchor-when-the-referenced-order-is).
- Adds `AnchorDisposition { Preserve, Release }` on the `Position` aggregate, threaded through `PositionCommand::FailOffChainOrder` and `PositionEvent::OffChainOrderFailed`.
- Adds `OrderFailureTerminality { Terminal, NotTerminal }` in `st0x-execution`, so the bot can tell an Alpaca status that is final from one that can still resume.
- Logs the real Alpaca reply when a placement is rejected for a duplicate `client_order_id`.

## Why

The anchor is never released when the Alpaca order it points at is already terminal. Every later placement for that symbol reuses the dead order's `client_order_id`, Alpaca answers `422 client_order_id must be unique`, the duplicate-recovery path adopts the stale terminal order as the result, and the hedge never reaches the market. Only a fill clears the anchor, and a fill cannot happen while the anchor holds, so the loop feeds itself and needs an operator to break it.

Production hit this more than once. DRAM on 2026-07-28 produced 402 failed orders and left about USD 830 short unhedged. DRAM and NVDA on 2026-08-04 and 2026-08-05 produced 843 failed placements over about 16 hours. Each one ended only when an operator ran a manual position adjustment.

## How

- **Release rule.** `OffChainOrderFailed` carries an `anchor` field. `Release` clears the anchor, `Preserve` keeps the first failed order id as before. The event field is `#[serde(default)]`, so events written before this change replay exactly as they did.
- **Evidence, not guesswork.** `Release` needs positive evidence that the broker order under the key is finished. A direct broker classification always wins. When there is no classification, the presence of an `executor_order_id` is the fallback. Every other site keeps `Preserve`, which is the fail-safe.
- **Terminality classification.** `map_broker_status_to_order_status` folds six Alpaca statuses into `OrderStatus::Failed`. `broker_failure_terminality` now splits them. `expired` and `rejected` are final. `done_for_day` is final only because every order this bot places is Day time-in-force, so it cannot resume in a later session. `replaced`, `suspended`, and `calculated` are not final, because those orders can still fill. The match is exhaustive, so a new Alpaca status cannot default into "final".
- **CLI repair keeps `Preserve`.** That path force-fails an order that is usually still live at the broker. A release there would re-arm the double hedge the anchor prevents.
- **Duplicate 422.** Both duplicate-recovery branches in `alpaca_broker_api/order.rs` log at `warn!` with the Alpaca code and message, instead of hiding them in a `debug!`.

## Testing

- `legacy_off_chain_order_failed_event_defaults_to_preserve` deserializes a hand-written JSON literal in the shape of a real production event, so a change to the wire format breaks the test instead of moving both sides together.
- `poll_with_expired_broker_order_enqueues_rejection_with_broker_time` and `expired_order_rejection_releases_anchor_and_pending` mirror the existing cancellation test across the poll job and the rejection job.
- `poll_with_suspended_broker_order_enqueues_rejection_as_not_terminal` plus an httpmock fixture with a real Alpaca `"status": "suspended"` body pin the classification at the executor boundary.
- `retry_of_suspended_order_rejection_preserves_the_anchor` covers the at-least-once retry window. It failed against the code before the fix, and passes after.
- `fail_pending_preserves_anchor_despite_executor_order_id_evidence` guards the deliberate CLI repair override, so a later refactor cannot turn it into a release without a red test.
- Preserve and Release pairs cover `dispatch_post_place_state`, `route_placement_outcome`, orphan recovery, the trade retry loop, and the CLI reconcile path.
- `cargo nextest run` passes 3021 tests in `st0x-hedge` and 350 in `st0x-execution`. `cargo clippy --workspace --all-targets --all-features` and `cargo fmt` are clean.

## Anything else

`Position::SCHEMA_VERSION` goes from 6 to 7, so snapshots rebuild on boot and every persisted aggregate replays under the new code. Run `nix run .#prodVerifyMigrations` before deploy. The unit test above covers the legacy event shape, but only a replay against real production data proves it.

Two follow-ups are filed under [RAI-1678](https://linear.app/makeitrain/issue/RAI-1678/address-follow-ups-from-review-of-rai-1559-broker-idempotency-anchor):

- [RAI-1679](https://linear.app/makeitrain/issue/RAI-1679/carry-broker-terminality-on-the-offchainorder-aggregate-so-every): five release sites still infer terminality from the presence of an `executor_order_id`, because the `OffchainOrder` aggregate drops the classification when it records `Failed`. The path that observes a suspended order is correct and tested here. These sites are the narrower retry window, and the fix needs a second event-schema change.
- [RAI-1680](https://linear.app/makeitrain/issue/RAI-1680/release-the-anchor-in-the-cli-force-fail-repair-path-when-the-order-is): the CLI repair path preserves the anchor even when the order is already known finished. It depends on RAI-1679.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of failed broker orders by distinguishing terminal outcomes from retryable failures.
  - Prevented duplicate hedges by preserving client-order safeguards when execution cannot be confirmed.
  - Released safeguards after confirmed terminal outcomes, intentional cancellation, expiry, rejection, or completion.
  - Improved recovery and reconciliation for missing, suspended, and failed orders.
  - Added compatibility handling for previously stored order and position data.

### DIFF
--- a/crates/execution/src/alpaca_broker_api/order.rs
+++ b/crates/execution/src/alpaca_broker_api/order.rs
@@ -9,9 +9,10 @@ use uuid::Uuid;
 use super::client::AlpacaBrokerApiClient;
 use super::{AlpacaBrokerApiError, CryptoOrderFailureReason, MissingOrderField, TimeInForce};
 use crate::{
-    ClientOrderId, Direction, ExecutorOrderId, FractionalShares, MarketOrder, OrderPlacement,
-    OrderStatus, OrderUpdate, Positive, Symbol, Usd, deserialize_float_from_number_or_string,
-    deserialize_option_float_from_number_or_string, serialize_float_as_string,
+    ClientOrderId, Direction, ExecutorOrderId, FractionalShares, MarketOrder,
+    OrderFailureTerminality, OrderPlacement, OrderStatus, OrderUpdate, Positive, Symbol, Usd,
+    deserialize_float_from_number_or_string, deserialize_option_float_from_number_or_string,
+    serialize_float_as_string,
 };
 
 const ALPACA_CRYPTO_MAX_DECIMAL_PLACES: u8 = 6;
@@ -297,7 +298,7 @@ impl CryptoOrderResponse {
     }
 
     /// Classifies the order's current status into a fill/pending/failed outcome,
-    /// consistent with the terminal mapping in `map_broker_status_to_order_status`.
+    /// consistent with the terminal mapping in `classify_broker_status`.
     ///
     /// The match is exhaustive (no wildcard) so a newly added Alpaca status forces
     /// a compile error here rather than silently mapping to `Pending` and retrying
@@ -401,7 +402,8 @@ pub(super) async fn place_market_order(
     let (order_id, shares, extended_hours, limit_price) = match client.place_order(&request).await {
         Ok(response) => (response.id, placed_shares, false, None),
         Err(error) if is_duplicate_client_order_id(&error) => {
-            debug!(
+            warn!(
+                %error,
                 client_order_id = %market_order.client_order_id,
                 "Broker rejected duplicate client_order_id; reconciling the order it already accepted"
             );
@@ -541,7 +543,8 @@ pub(super) async fn place_limit_order(
             Some(*limit_order.limit_price.as_price()),
         ),
         Err(error) if is_duplicate_client_order_id(&error) => {
-            debug!(
+            warn!(
+                %error,
                 client_order_id = %limit_order.client_order_id,
                 "Broker rejected duplicate client_order_id; reconciling the order it already accepted"
             );
@@ -597,7 +600,7 @@ pub(super) async fn get_order_status(
         OrderSide::Sell => Direction::Sell,
     };
 
-    let status = map_broker_status_to_order_status(response.status);
+    let (status, failure_terminality) = classify_broker_status(response.status);
     let price = response.filled_average_price;
     let shares_filled = response.filled_quantity.map(FractionalShares::new);
 
@@ -654,6 +657,7 @@ pub(super) async fn get_order_status(
         updated_at,
         price,
         shares_filled,
+        failure_terminality,
     })
 }
 
@@ -699,34 +703,48 @@ fn broker_time_or_observation(broker_time: Option<DateTime<Utc>>, order_id: &str
     })
 }
 
-fn map_broker_status_to_order_status(status: BrokerOrderStatus) -> OrderStatus {
+/// Classifies a broker status into the `OrderStatus` it maps to, and --
+/// whenever that status is `OrderStatus::Failed` -- the failure's
+/// terminality per Alpaca's order lifecycle
+/// (https://docs.alpaca.markets/docs/orders-at-alpaca). `DoneForDay` is
+/// terminal only because every order this bot places is Day time-in-force,
+/// so it cannot resume in a later session.
+///
+/// A single exhaustive match producing both classifications together, rather
+/// than two independent matches over `BrokerOrderStatus`: the previous shape
+/// let a status be added to one match's `Failed`/terminal arm without the
+/// other match being updated to agree, which only surfaced as a runtime
+/// `MissingOrderField::FailureTerminality` error. Here, a newly added status
+/// must be classified for both dimensions in the same arm or the match is
+/// non-exhaustive and the crate fails to compile.
+fn classify_broker_status(
+    status: BrokerOrderStatus,
+) -> (OrderStatus, Option<OrderFailureTerminality>) {
+    use BrokerOrderStatus::*;
+    use OrderFailureTerminality::{NotTerminal, Terminal};
+
     match status {
-        // Submitted to broker and in progress
-        BrokerOrderStatus::New
-        | BrokerOrderStatus::Accepted
-        | BrokerOrderStatus::PendingNew
-        | BrokerOrderStatus::AcceptedForBidding
-        | BrokerOrderStatus::PendingCancel
-        | BrokerOrderStatus::PendingReplace
-        | BrokerOrderStatus::Stopped => OrderStatus::Submitted,
+        // Submitted to broker and in progress.
+        New | Accepted | PendingNew | AcceptedForBidding | PendingCancel | PendingReplace
+        | Stopped => (OrderStatus::Submitted, None),
 
         // Partially filled -- distinct from Submitted so the poll loop can
         // drive `UpdatePartialFill` on the aggregate before any cancel.
-        BrokerOrderStatus::PartiallyFilled => OrderStatus::PartiallyFilled,
+        PartiallyFilled => (OrderStatus::PartiallyFilled, None),
 
-        // Successfully filled
-        BrokerOrderStatus::Filled => OrderStatus::Filled,
+        // Successfully filled.
+        Filled => (OrderStatus::Filled, None),
 
         // Cancelled by the broker after a cancel request was accepted.
-        BrokerOrderStatus::Canceled => OrderStatus::Cancelled,
+        Canceled => (OrderStatus::Cancelled, None),
 
-        // Failed/terminal statuses
-        BrokerOrderStatus::Expired
-        | BrokerOrderStatus::DoneForDay
-        | BrokerOrderStatus::Rejected
-        | BrokerOrderStatus::Replaced
-        | BrokerOrderStatus::Suspended
-        | BrokerOrderStatus::Calculated => OrderStatus::Failed,
+        // Failed/terminal statuses: the order will never resume or fill
+        // further, so a caller may release its idempotency key.
+        Expired | Rejected | DoneForDay => (OrderStatus::Failed, Some(Terminal)),
+
+        // Failed statuses the order may still resume or fill from: a caller
+        // must NOT release its idempotency key on these.
+        Replaced | Suspended | Calculated => (OrderStatus::Failed, Some(NotTerminal)),
     }
 }
 
@@ -860,6 +878,7 @@ mod tests {
     use crate::alpaca_broker_api::auth::{
         AlpacaAccountId, AlpacaBrokerApiCtx, AlpacaBrokerApiMode,
     };
+    use crate::alpaca_broker_api::duplicate_client_order_id_body;
     use st0x_float_macro::float;
 
     const TEST_ACCOUNT_ID: AlpacaAccountId =
@@ -1053,7 +1072,7 @@ mod tests {
                 .path("/v1/trading/accounts/904837e3-3b76-47ec-b432-046db621571b/orders");
             then.status(422)
                 .header("content-type", "application/json")
-                .json_body(json!({"message": "client_order_id must be unique"}));
+                .json_body(duplicate_client_order_id_body());
         });
 
         // We reconcile by adopting the order the broker actually accepted.
@@ -1102,6 +1121,52 @@ mod tests {
         assert_eq!(placement.limit_price, None);
     }
 
+    /// The duplicate-422 warn log includes the Alpaca-reported numeric code
+    /// alongside the message. `place_market_order`'s reconciliation swallows
+    /// the intermediate `ApiError` on success, so this drives `place_order`
+    /// directly against a realistic duplicate-client_order_id body to prove
+    /// `code` is actually parsed into `alpaca_code`, not just `message`.
+    #[tokio::test]
+    async fn place_order_parses_alpaca_code_from_duplicate_client_order_id_body() {
+        let server = MockServer::start();
+        let ctx = create_test_ctx(AlpacaBrokerApiMode::Mock(server.base_url()));
+
+        let mock = server.mock(|when, then| {
+            when.method(POST)
+                .path("/v1/trading/accounts/904837e3-3b76-47ec-b432-046db621571b/orders");
+            then.status(422)
+                .header("content-type", "application/json")
+                .json_body(duplicate_client_order_id_body());
+        });
+
+        let client = AlpacaBrokerApiClient::new(&ctx).unwrap();
+        let request = OrderRequest {
+            symbol: Symbol::new("AAPL").unwrap(),
+            quantity: Positive::new(FractionalShares::new(float!(1))).unwrap(),
+            side: OrderSide::Buy,
+            order_type: "market",
+            time_in_force: TimeInForce::Day.as_api_str(),
+            extended_hours: false,
+            client_order_id: ClientOrderId::from_uuid(uuid!(
+                "77777777-7777-4777-8777-777777777777"
+            )),
+        };
+
+        let error = client.place_order(&request).await.unwrap_err();
+
+        mock.assert();
+        let AlpacaBrokerApiError::ApiError {
+            alpaca_code,
+            message,
+            ..
+        } = error
+        else {
+            panic!("expected ApiError, got {error:?}");
+        };
+        assert_eq!(alpaca_code, Some(40_010_001));
+        assert_eq!(message, "client_order_id must be unique");
+    }
+
     #[tokio::test]
     async fn place_market_order_adoption_reports_adopted_extended_hours_terms() {
         // Lost-response scenario the convergence sweep depends on: an
@@ -1122,7 +1187,7 @@ mod tests {
                 .path("/v1/trading/accounts/904837e3-3b76-47ec-b432-046db621571b/orders");
             then.status(422)
                 .header("content-type", "application/json")
-                .json_body(json!({"message": "client_order_id must be unique"}));
+                .json_body(duplicate_client_order_id_body());
         });
 
         let lookup_mock = server.mock(|when, then| {
@@ -1184,7 +1249,7 @@ mod tests {
                 .path("/v1/trading/accounts/904837e3-3b76-47ec-b432-046db621571b/orders");
             then.status(422)
                 .header("content-type", "application/json")
-                .json_body(json!({"message": "client_order_id must be unique"}));
+                .json_body(duplicate_client_order_id_body());
         });
 
         let lookup_mock = server.mock(|when, then| {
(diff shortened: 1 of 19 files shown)


JuaniRios:
* **#1137** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1137?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1137?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1137).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/offchain/order.rs
```diff
@@ -2079,13 +2082,15 @@ pub(crate) fn terminal_position_finalization(
             executor_order_id.clone(),
             NoFillOutcome::Failed {
                 error: error.clone(),
+                anchor: AnchorDisposition::Release,
```
0xgleb: should fix: this arm always sets anchor to Release whenever Failed carries an executor_order_id, with no OrderFailureTerminality. Orphan recovery, position-check finalize, and poll finalization all go through terminal_position_finalization, so a locally MarkFailed suspended order can clear last_failed_offchain_order_id while Alpaca may still fill. I'd default to Preserve here until RAI-1679 stores terminality on the aggregate, or thread the poll classification into this helper.

src/offchain/order/handle_rejection.rs
```diff
@@ -227,21 +236,45 @@ impl Job<HandleOrderRejectionCtx> for HandleOrderRejection {
...
             }
-            | Failed { .. } => PositionCommand::FailOffChainOrder {
-                offchain_order_id: self.offchain_order_id,
-                error: self.error.clone(),
-            },
+
+            Failed {
+                executor_order_id, ..
+            } => {
+                // Two independent sources can each supply the positive
```
0xgleb: should fix: step 1 persists MarkFailed before step 2 updates the position, and step 2 still matches the pre-MarkFailed snapshot. Other workers can see Failed plus executor_order_id while pending_offchain_order_id is still set, and recovery that uses terminal_position_finalization can Release before this job applies NotTerminal. Worth aligning recovery rules with this job or persisting terminality on the order before exposing Failed.

src/offchain/order/handle_rejection.rs
```diff
@@ -227,21 +236,45 @@ impl Job<HandleOrderRejectionCtx> for HandleOrderRejection {
...
+                // evidence that releases the anchor: this job's own
+                // `broker_terminality` (the enqueuing poll observed the
+                // broker-documented terminal state directly), or the order's
+                // own `executor_order_id` (an earlier attempt's step 1
+                // committed before step 2 crashed, so the order is already
+                // `Failed` with the broker key recorded on it even though
+                // this job instance carries no terminality evidence).
```
0xgleb: minor: on apalis retry when the order is already Failed, from_broker_evidence(self.broker_terminality, executor_order_id) treats None terminality plus a broker id as Release. Pre-deploy queued jobs missing broker_terminality can release on retry even when the poll saw a non-terminal status. Consider Preserve when terminality is unknown on this arm, or drain/requeue on deploy.

src/offchain/order/handle_rejection.rs
```diff
@@ -260,10 +293,17 @@ fn position_command_for_retained_fill(
     broker_timestamp: chrono::DateTime<chrono::Utc>,
     fallback_error: String,
 ) -> PositionCommand {
+    // Every caller of this function (PartiallyFilled, Cancelling, and Failed
+    // arms alike) always supplies a broker-assigned `executor_order_id`:
+    // positive evidence the order reached a terminal state at the broker, so
+    // the fallback below must release the anchor rather than preserve it.
+    let anchor = AnchorDisposition::Release;
```
0xgleb: should fix: position_command_for_retained_fill always uses Release and never reads self.broker_terminality. The main match arms honor NotTerminal, but priced retained-fill paths can still clear the anchor if the broker failure was suspended or otherwise non-terminal. I'd derive anchor with from_broker_evidence here too.

src/integration_tests/arbitrage.rs
```diff
@@ -290,6 +292,7 @@ async fn poll_submitted_orders<E: st0x_execution::Executor + Clone>(
                             PositionCommand::FailOffChainOrder {
                                 offchain_order_id: order_id,
                                 error,
+                                anchor: AnchorDisposition::Preserve,
```
0xgleb, line 299: minor: poll_submitted_orders still sends FailOffChainOrder with Preserve and bypasses PollOrderStatus plus HandleOrderRejection, even though mocks now attach Terminal terminality. Tests like position_checker_recovers_failed_execution will not catch regressions on anchor release for terminal broker failures. Consider driving the production job path or asserting last_failed_offchain_order_id clears when simulating terminal Failed.

0xgleb (changes requested):

src/offchain/order.rs
JuaniRios: Fixed in `b2e68a98`. That arm is `Preserve` now. Went further than suggested: `executor_order_id` presence no longer releases anywhere. `AnchorDisposition::from_broker_terminality` releases only on `Some(Terminal)`, and `from_broker_evidence`/`from_executor_order_id` are deleted. RAI-1679 still tracks storing terminality on the aggregate so these paths can release precisely instead of waiting a cycle for the poll path.

src/offchain/order/handle_rejection.rs
JuaniRios: Fixed in `b2e68a98`. The recovery paths that could observe `Failed` plus `executor_order_id` now preserve, so they cannot release before this job applies `NotTerminal`. Persisting terminality on the order before exposing `Failed` is RAI-1679.

src/offchain/order/handle_rejection.rs
JuaniRios: Fixed in `b2e68a98`. Good catch on the deploy window. The already-`Failed` arm derives from `broker_terminality` alone, so a job queued before this change deserializes `None` and preserves. No drain or requeue needed.

src/offchain/order/handle_rejection.rs
JuaniRios: Fixed in `b2e68a98`. `position_command_for_retained_fill` takes `broker_terminality` and derives the anchor from it. The comment claiming `executor_order_id` was positive terminal evidence is gone, it was the same wrong inference.

src/integration_tests/arbitrage.rs
JuaniRios: Fixed in `b2e68a98`. Added `poll_and_reject_via_real_jobs`, which runs the real `PollOrderStatus::perform` then the enqueued `HandleOrderRejection`. `position_checker_recovers_failed_execution` uses it for the terminal-failure step and asserts `last_failed_offchain_order_id` clears. Left the shared `poll_submitted_orders` helper alone, 17 other call sites depend on its partial-fill and cancellation behaviour.

src/offchain/order/handle_rejection.rs
```diff
@@ -259,11 +267,15 @@ fn position_command_for_retained_fill(
...
 ) -> PositionCommand {
+    let anchor = AnchorDisposition::from_broker_terminality(broker_terminality);
+
     Positive::new(shares_filled).map_or_else(
         |_| PositionCommand::FailOffChainOrder {
             offchain_order_id,
             error: fallback_error,
+            anchor,
         },
         |positive_filled| PositionCommand::CompleteOffChainOrder {
```
agryaznov, line 294: should_fix: The nonzero branch still finalizes the position regardless of `broker_terminality` -- only the zero-share arm uses the new `anchor`. Reachable: order partially fills, broker suspends it, poll enqueues rejection with `Some(NotTerminal)` -> position completed with the partial qty, pending cleared, and the suspended order drops out of the poller. If it resumes and fills the remainder, nothing tracks that fill. Suggest: for `Some(NotTerminal)`, don't finalize -- keep the order/position pending (or re-enqueue a poll) instead of completing off a partial.

src/offchain/order.rs
```diff
@@ -2279,6 +2287,7 @@ async fn reconcile_pre_cancel(
             failed_at,
             shares_filled,
             avg_price,
+            ..
```
agryaznov: should_fix: This `..` discards the `terminality` the probe just fetched -- the freshest evidence in the system -- and the comment above ("terminally failed at the broker") is now wrong for `Suspended`/`Replaced`/`Calculated`. A suspended order gets marked locally `Failed`, stops being polled, and can still resume+fill at the broker untracked. Suggest: match on `terminality` here -- only `Terminal` should take this mark-failed path; `NotTerminal` should stay live (proceed with the cancel, or leave `Submitted` for the poller).

crates/execution/src/alpaca_broker_api/order.rs
```diff
@@ -699,6 +704,23 @@ fn broker_time_or_observation(broker_time: Option<DateTime<Utc>>, order_id: &str
     })
 }
 
+/// Classifies a broker-reported Failed order's finality, per Alpaca's order
+/// lifecycle (https://docs.alpaca.markets/docs/orders-at-alpaca). `DoneForDay` is
+/// final only because every order this bot places is Day time-in-force, so it
+/// cannot resume in a later session. Exhaustive: a newly added Alpaca status must
+/// not default into `Terminal` and release a caller's idempotency key early.
+fn broker_failure_terminality(status: BrokerOrderStatus) -> Option<OrderFailureTerminality> {
```
agryaznov: non-blocking: This match and `map_broker_status_to_order_status` must agree on the `Failed` status set by hand -- the executor's fail-fast (`MissingOrderField::FailureTerminality`) even documents the drift risk. A status added to one match but not the other compiles clean and then stalls the rejection pipeline at runtime. Suggest: one classifier returning `(OrderStatus, Option<OrderFailureTerminality>)` from a single match, so drift is a compile error and the error variant disappears.

crates/execution/src/alpaca_broker_api/order.rs
```diff
@@ -1048,12 +1070,17 @@ mod tests {
...
+        // The body shape (including the numeric `code`) matches a real Alpaca
+        // duplicate-client_order_id rejection.
         let place_mock = server.mock(|when, then| {
             when.method(POST)
                 .path("/v1/trading/accounts/904837e3-3b76-47ec-b432-046db621571b/orders");
             then.status(422)
                 .header("content-type", "application/json")
-                .json_body(json!({"message": "client_order_id must be unique"}));
+                .json_body(json!({
+                    "code": 40_010_001,
```
agryaznov: non-blocking: This test (and its sibling at L1148) mock the duplicate-`client_order_id` 422 with `"code": 40_010_001`, while the shared `MockApi` (`mock_api.rs`) and the older tests serve a code-less `{"message": ...}` body. Two canonical bodies for the same broker error: code keying on `alpaca_code` would pass here and break against every `MockApi`-driven path. Suggest: one shared fixture for this response, reused by both.

agryaznov (changes requested):

agryaznov:
Anchor rule looks solid now — `Release` only on `Some(Terminal)` is the right standard. My `should_fix` inlines are a separate bug class: locally finalizing an order the broker can still fill drops it from the poller, so a later resume+fill goes untracked (delta exposure).

Follow-up for RAI-1678: once terminality is persisted on the aggregate (RAI-1679), consider deriving `client_order_id` deterministically (e.g. `uuidv5(symbol, execution_seq)`) — the stored anchor and its release-policy edge cases disappear, broker-side dedupe stays, and DB-loss recovery (list Alpaca orders → recompute keys → match) comes free.

src/offchain/order/handle_rejection.rs
JuaniRios: Fixed in `69e0c637`. `position_command_for_retained_fill` returns early with `FailOffChainOrder` unless `broker_terminality == Some(Terminal)`, so a partial is never locked in as final for an order the broker may still fill.

I went conservative rather than keeping the order in the poll loop. `reschedule_self` has no attempt cap or timeout (only `backpressure_streak` is bounded, and it resets on every normal reschedule), so holding a suspended order there risks the ~15k-reschedule runaway we hit before. The anchor is preserved instead, so the retry dedupes against the same broker order and the fill reconciles once it reaches a terminal state.

New test: `suspended_partial_fill_rejection_does_not_complete_the_position`. Also flipped `partial_fill_rejection_retains_executed_quantity_on_position`, it was asserting the bug.

src/offchain/order.rs
JuaniRios: Fixed in `69e0c637`. `reconcile_pre_cancel`'s `Failed` arm matches on `terminality` now. Only `Terminal` short-circuits the DELETE. `NotTerminal` reconciles any newer broker-reported fill and then proceeds with the real cancel, same as the `Pending`/`Submitted` arm. The "terminally failed" comment is corrected.

Tests: `cancel_order_with_not_terminal_broker_failure_proceeds_with_cancel` and `cancel_order_with_not_terminal_broker_failure_reconciles_newer_partial_fill`.

crates/execution/src/alpaca_broker_api/order.rs
JuaniRios: Fixed in `69e0c637`. The two matches are now one exhaustive `classify_broker_status` returning both values, so a status cannot be added to one dimension only.

One caveat: I kept the executor fail-fast and `MissingOrderField::FailureTerminality`. `OrderUpdate` carries a flat `OrderStatus` plus `Option<OrderFailureTerminality>`, so the compiler still cannot prove `Failed => Some(..)` at that boundary without pairing them in the type. Pairing `OrderUpdate` is the real end state, happy to do it here if you'd prefer.

crates/execution/src/alpaca_broker_api/order.rs
JuaniRios: Fixed in `69e0c637`. Added a shared `duplicate_client_order_id_body()` fixture in `alpaca_broker_api/mod.rs`, now used by `mock_api.rs` and all seven duplicate-422 mocks, with the code included.

crates/execution/src/order/mod.rs
```diff
@@ -124,6 +124,12 @@ pub struct OrderUpdate<OrderId> {
     /// `OrderStatus::PartiallyFilled` so the caller can reconcile the local
     /// aggregate via `UpdatePartialFill`.
     pub shares_filled: Option<FractionalShares>,
+    /// Broker-terminality classification of the failure, populated only when
+    /// `status` is `OrderStatus::Failed`. Carries the evidence executor
+    /// implementations need to build `OrderState::Failed`'s `terminality`
+    /// field without leaking their broker-specific status enum across the
+    /// `Executor` boundary.
+    pub failure_terminality: Option<OrderFailureTerminality>,
```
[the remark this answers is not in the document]
JuaniRios: Skipping this one. `failure_terminality` does have a consumer: `executor.rs:210` reads it to build `OrderState::Failed { terminality }`, so it is not debug-only transport.

Encoding terminality into the status enum is the right end state, but it means pairing them in `OrderUpdate`'s type, which is wider than this PR. Same caveat I noted on the classifier thread.

JuaniRios:
Answering the two out-of-diff findings from the CodeRabbit review, since they have no thread to reply on.

**`src/offchain/order.rs` L2282-2288, persist broker terminality on the aggregate.** Agreed, and it is filed as [RAI-1679](https://linear.app/makeitrain/issue/RAI-1679/carry-broker-terminality-on-the-offchainorder-aggregate-so-every) under [RAI-1678](https://linear.app/makeitrain/issue/RAI-1678/address-follow-ups-from-review-of-rai-1559-broker-idempotency-anchor). It needs a second event-schema change and a `SCHEMA_VERSION` bump on `OffchainOrder`, so it stays out of this PR. Note the hazard it describes is already closed here: no release path infers terminality from `executor_order_id` presence any more, so those sites preserve rather than release wrongly. What remains is precision, releasing at placement time instead of waiting a cycle for the poll path.

**`src/offchain/order.rs` L2073-2095, route both arms through `AnchorDisposition::from_broker_evidence`.** Stale. `from_broker_evidence` and `from_executor_order_id` were both deleted in `b2e68a98`. The only constructor now is `from_broker_terminality`, which releases solely on `Some(Terminal)`, and both of those arms already go through the fail-safe.

crates/execution/src/alpaca_broker_api/order.rs
```diff
@@ -2274,45 +2339,130 @@ mod tests {
...
+            let (order_status, terminality) = classify_broker_status(status);
+            assert_eq!(
+                order_status == OrderStatus::Failed,
+                terminality.is_some(),
+                "status {status:?} classified as {order_status:?} with \
+                 terminality {terminality:?} -- Failed must always pair with \
+                 Some terminality, and nothing else may carry one"
+            );
+        }
+    }
```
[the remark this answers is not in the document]
JuaniRios: Fixed in `3dab87e2`. Replaced the invariant loop with a table asserting the exact `(OrderStatus, Option<OrderFailureTerminality>)` pair for all 16 `BrokerOrderStatus` variants, so a status routed to the wrong non-failure status is caught too.

Mutation-checked your exact example: mis-mapping `PendingCancel` to `Cancelled` now fails the test, where the old assertion passed. Kept the named per-status tests for readable failures.

agryaznov (approved):
Residuals (wedge-until-operator, runtime check at the OrderUpdate boundary) are fine as follow-ups under RAI-1678/1679. LGTM.

0xgleb (approved):

PR #1141, opened 2026-08-07
Add mint authorization signing and delivery for orchestrator-mode mints

rouzwelt:
## What

Closes RAI-1243. Adds `mint_authorization` — the module that produces and delivers EIP-712 recipient authorizations for orchestrator-mode mints.

## Why

Assets being cut over to the ST0xOrchestrator can no longer be minted directly; the orchestrator requires a `MintAuth` authorization signed by the recipient wallet (this bot's wallet) before it will execute a mint. Without this, any asset migrated to orchestrator mode would stall at the on-chain step.

## How

**Signing (`MintAuthorizationService`):** Rather than reconstructing the EIP-712 domain and typehash locally, the service reads the digest directly from the orchestrator contract's `mintAuthDigest` view. This guarantees the signed hash is byte-identical to what the orchestrator verifies at mint time, and eliminates any risk of local domain/typehash drift. A minimal hand-written `IST0xOrchestratorV1` ABI binding exposes only the two views the bot needs (`mintAuthDigest` and `nonceUsed`).

**Delivery (`MintAuthorizationDeliverer`** **on** **`IssuanceClient`):** Delivers `{nonce, signature}` to issuance keyed by tokenization request ID. HTTP responses are classified into typed variants (`Recorded`, `MintNotFoundYet`, `Conflict`, `Rejected`, `RetryableFailure`) so the delivery job can branch on retry vs. park without inspecting raw status codes. The issuance endpoint URL is never allowed to reach logs — `ClientError` is reduced to the endpoint-free `IssuanceClientFailure` at the boundary.

**Vault-mode discovery (`VaultModeReader`** **on** **`IssuanceClient`):** Reads the `vault_mode` field from issuance's per-asset status endpoint so the mint flow can determine whether a recipient authorization is needed. Issuance's config is the single source of truth; the bot keeps no asset-mode list. Unknown assets fail closed — guessing vault-direct would skip a required authorization and stall the mint.

**Shared** **`IssuanceClientFailure`:** The `FreezeCheckError` in the rebalancing trigger previously defined its own identical `FreezeClientFailure` enum. That is replaced with the shared `IssuanceClientFailure` from `mint_authorization`, and the `From<ClientError>` conversion is delegated to the shared impl.

[...]

PR #1139, opened 2026-08-07
Add `sign_digest` to `Wallet` trait with local and Turnkey backends

rouzwelt:
## What

Adds a `sign_digest` method to the `Wallet` trait that signs a precomputed 32-byte digest (e.g. an EIP-712 signing hash) without broadcasting a transaction or consuming a nonce.

## Why

Certain on-chain verification flows (e.g. MintAuthV1) require a detached signature over a precomputed hash rather than a signed transaction. Neither the local nor Turnkey wallet backends previously exposed this capability through the `Wallet` trait.

## How

A new `sign_digest(digest: B256) -> Result<Signature, EvmError>` method is added to the `Wallet` trait and implemented across all backends:

- **Local signer**: delegates directly to `PrivateKeySigner::sign_hash`. The signer is now stored alongside the `signing_provider` so it is available for detached signing without going through the transaction pipeline.
- **Turnkey**: uses the `ACTIVITY_TYPE_SIGN_RAW_PAYLOAD_V2` activity with `HASH_FUNCTION_NO_OP`, transmitting the digest as bare lowercase hex so Turnkey signs it as-is rather than hashing again. `TracingTurnkeySigner` is made `Clone` and `Arc`\-wrapped so `TurnkeyWallet` can hold a dedicated `digest_signer` handle while the original lives inside the `EthereumWallet` for transaction signing — both share the same HTTP client. The assembled `r`/`s`/`v` components are validated by recovering the signer address from the digest, applying the same defensive trust boundary used for transaction signatures.
- **Stub/mock wallets**: panic with a descriptive message, making accidental calls in unrelated tests immediately visible.

Recovery id parsing accepts both the canonical `00`/`01` encoding Turnkey documents and the Ethereum-style `1b`/`1c` (27/28) encoding, so a provider-side representation change cannot silently flip signature parity. Short, oversized, or non-hex scalar components are rejected with precise errors rather than being silently zero-extended.

## Testing

- **`sign_digest_recovers_to_wallet_address`** (local): signs a digest with the local wallet and verifies the signature recovers to the wallet's address.
- **`sign_digest_returns_signature_from_completed_activity`** (Turnkey): mocks the `/sign_raw_payload` endpoint, asserts the request carries the correct `HASH_FUNCTION_NO_OP` and hex-encoded digest, and verifies the assembled signature matches and recovers correctly.
- **`sign_digest_rejects_signature_from_wrong_signer`**: confirms components recovering to a different address are rejected.
[...]

PR #1138, opened 2026-08-06
Add optional `[orchestrator]` config section for ST0xOrchestrator address

rouzwelt:
## What

Adds an optional `[orchestrator]` config section that carries the `ST0xOrchestrator` contract address, exposing it as `OrchestratorConfig` through `Ctx`.

## Why

Orchestrator-mode mints require a signed `MintAuthV1` recipient authorization, which needs the orchestrator's contract address as the EIP-712 verifying contract. Without a place to configure this address, the bot cannot support orchestrator-mode assets. The section is optional so the bot continues to run unchanged while all assets remain vault-direct ("deploys dark"), but a mint that encounters an orchestrator-mode asset with no configured address must fail loudly rather than guess.

## How

A new `orchestrator` module defines `OrchestratorConfig`, which wraps a single `Address` field. Deserialization goes through an intermediate `OrchestratorConfigToml` type whose `TryFrom` conversion rejects the zero address at parse time — meaning a misconfigured or placeholder address fails `validate-config` before any asset ever cuts over to orchestrator mode. The `[orchestrator]` section is absent from the config by default (`None`), preserving existing vault-direct behavior with no required migration.

`OrchestratorConfig` is threaded through `Config`, `ValidatedParts`, and `Ctx`, including the builder used in tests.

## Testing

Unit tests in `orchestrator.rs` cover valid address parsing, zero-address rejection, malformed addresses, missing `address` field, and unknown fields.

Integration-level tests in `loader.rs` verify that a well-formed `[orchestrator]` section flows through `parse_and_validate` into `ValidatedParts`, that omitting the section produces `None`, and that a zero address causes a `ConfigToml` error with the expected message.

## Anything else

Which assets are orchestrator-mode is determined by the issuance bot's per-asset `vault_mode` status field — this config section only supplies the contract address and never an asset list. The `example.config.toml` entry for `[orchestrator]` is commented out to reflect that it is optional.  
  
Closes [RAI-1674](https://linear.app/makeitrain/issue/RAI-1674/orchestrator-config)

---

[...]

PR #1130, opened 2026-08-06
Restart-tainted balance guard for open-order straddling reboots

rouzwelt:
## What

Closes [RAI-1501](https://linear.app/makeitrain/issue/RAI-1501/restart-across-an-open-hedge-order-can-transiently-double-count-the). Introduces restart taint to eliminate the double-count drift window that arises when a hedge order straddles a bot restart.

## Why

The aggregate records every broker poll, including mid-order readings that the live view's fill guards refused to apply. Those guards rely on a local-clock `last_offchain_fill_applied_at` stamp that does not survive a restart. After boot, the hydrated Hedging balance for a symbol whose hedge order was open at shutdown may or may not already contain the order's fill — there is no way to know. Without this fix, applying the fill delta on top of a hydrated number that already absorbed it produces an unbounded double-count drift.

## How

At boot, `set_pending_offchain_order_symbols` now stamps a restart taint alongside the order gate for every symbol with an open hedge order, and sets a venue-level cash taint because any open order also moves the Hedging USDC balance. The taint is seeded in one place (the boot seam) so the gate and the taint cannot desynchronize.

While a symbol is tainted:

- A discovered fill skips both the equity delta and the mirrored USDC delta, clears the order gate without recording an applied-fill timestamp, and leaves the taint in place. A failed or cancelled order also leaves the taint, since the pre-restart portion may have partially filled.
- Equity dispatch for the symbol and USDC dispatch for the venue are suppressed, exactly as under an engaged divergence gate.

The poller resolves the taint against broker truth:

- A matching quiet poll proves the hydrated value and lifts the taint immediately, without escalating.
- A diverging quiet poll escalates a forced reconcile on the first poll rather than waiting out the N-poll confirmation threshold — the ambiguity was created by the boot, not a transient race — and a verified heal lifts the taint.

The accepted cost is that a fill during the tainted window is reflected only after the re-base lands, at most one poll interval later, for at most the orders that straddled the restart.

## Testing

Five new integration tests cover the full resolution matrix:

- `restart_tainted_divergence_escalates_on_first_quiet_poll` — a tainted divergence escalates at `consecutive_polls == 1` regardless of the configured threshold, and an unhealed escalation keeps the taint.
- `restart_tainted_divergence_heals_and_clears_taint_within_one_poll` — the full post-restart path: aggregate dedupes the ordinary snapshot, the taint-driven escalation re-bases the view from broker truth, and the taint and gate are both cleared within a single poll.
[...]

PR #1145, opened 2026-08-07
Wire orchestrator-mode mint authorization signing and delivery

rouzwelt:
## What

Implements orchestrator-mode mint authorization (RAI-1243). When an asset's `vault_mode` on the issuance service reads `"orchestrator"`, the mint saga now signs a `MintAuthV1` recipient authorization and delivers it to issuance before polling for completion. Vault-direct assets are unaffected.

## Why

Issuance will not execute an orchestrator-mode mint until it receives a valid recipient authorization signed by the configured orchestrator wallet. Without this, orchestrator-mode mints would be accepted by the bot but stall indefinitely at the issuance polling step.

## How

**Aggregate changes (`tokenized_equity_mint.rs`):**

- Added `MintAuthorizationProgress` enum (`NotSigned`, `Signed`, `Delivered`) embedded in the `MintAccepted` state with `#[serde(default)]` so existing event streams replay unchanged.
- Added `MintAuthorizationSigned` and `MintAuthorizationDelivered` events, persisted to the event log with hex-encoded nonce and signature.
- Added `SignMintAuthorization` and `RecordMintAuthorizationDelivered` commands. Signing is idempotent: a re-drive finding an existing authorization emits nothing rather than minting a fresh nonce (the nonce is the mint's on-chain idempotency key).
- Bumped `SCHEMA_VERSION` to 5 to clear stale snapshots.
- Extended `interrupted_mint_ids` to include mints whose latest event is `MintAuthorizationSigned` or `MintAuthorizationDelivered`, so a crash between signing and delivery is recovered on restart.

**`ConfiguredMintAuthorizer`:**

- New `Enabled`/`Disabled` enum added to `mint_authorization.rs`, mirroring `BotGasReceiptCostEnqueuer`. `Disabled` fails loudly if signing is ever reached without an `[orchestrator]` config section. Added to `EquityTransferServices`.

**Saga wiring (`rebalancing/equity/mod.rs`):**

- `CrossVenueEquityTransfer` gains an optional `MintAuthorizationWiring` (vault-mode reader, token address map, delivery queue), opted in via `with_mint_authorization`.
- `ensure_mint_authorization` is called both on the initial mint acceptance path and on every resume from `MintAccepted`. It reads `vault_mode` from issuance, signs if orchestrator-mode, and enqueues the delivery job. A live-delivery guard prevents a second delivery chain from starting during a delayed-redrive window.

**Delivery job (`rebalancing/equity/authorization_job.rs`):**

- New `DeliverMintAuthorization` apalis job. Loads the persisted authorization from the aggregate and delivers it to issuance. Uses bounded delayed redrives (30 s delay, 20-attempt budget, ~10 minutes total) for retryable outcomes (`MintNotFoundYet`, transport failures) rather than consuming the three-attempt apalis budget. Non-retryable rejections (409/422) park the mint with an operator alert and return `Ok` so the best-effort worker keeps running.
- Registered on a best-effort worker so a stuck authorization never halts hedging or fill detection.

**Conductor wiring:**

- `build_mint_authorization_infra` constructs the `ConfiguredMintAuthorizer` (enabled only when `[orchestrator]` is configured), the delivery queue, the `MintAuthorizationWiring`, and a shared `IssuanceClient`. Orphaned delivery rows from a prior crash are requeued at startup.
- `DeliverMintAuthorizationJobQueue` and `DeliverMintAuthorizationCtx` are threaded through `PositionAndRebalancing`, `RebalancingInfrastructure`, and the conductor builder.
- CLI and non-rebalancing paths set `mint_authorizer: ConfiguredMintAuthorizer::Disabled`; the comment explains why orchestrator-mode mints must run through the server.

**Stall clock and performance tracking:**

- `MintAuthorizationSigned` and `MintAuthorizationDelivered` refresh the stall clock in `MintTracking` without advancing the stage (genuine forward progress for a mint waiting in `Accepted`).
- Both events are treated as no-ops in inventory, reliability, and timing projections where they carry no share movement.

## Testing

- Unit tests on the aggregate cover: signing persists `Signed` progress; re-signing keeps the original nonce; distinct mints generate independent nonces; signing with `Disabled` authorizer fails loudly; delivery transitions to `Delivered` and is idempotent; delivery without a signed authorization errors.
- Wire-format pin tests assert the exact hex JSON shape of `MintAuthorizationSigned` and `MintAuthorizationDelivered` events.
- Replay tests confirm pre-authorization streams default to `NotSigned` and that a pinned hex payload materializes with the correct nonce and signature bytes.
- `interrupted_mint_ids` test confirms mints awaiting delivery are included in startup recovery.
- Delivery job tests cover: `Recorded` marks the aggregate delivered; `Rejected`/`Conflict` parks with one operator alert; `MintNotFoundYet` schedules a delayed redrive; budget exhaustion dead-letters with an alert; an already-delivered mint is a no-op.
- Integration tests on `CrossVenueEquityTransfer` cover: orchestrator-mode resume signs once and enqueues exactly one delivery job; a second resume is idempotent; the live-delivery guard prevents a second chain over a pending redrive successor; vault-direct resume signs nothing and enqueues nothing.

## Anything else

The `[orchestrator]` config section remains optional. While every asset is vault-direct the bot deploys without it, and an orchestrator-mode mint reaching the signing step fails loudly rather than guessing an address. The mint stays in `MintAccepted` with its authorization persisted and is resumable once the section is added.

Closes [RAI-1677](https://linear.app/makeitrain/issue/RAI-1677/mint-authorization-signing-delivery-and-mode-discovery)  
Part of [RAI-1243](https://linear.app/makeitrain/issue/RAI-1243/liquidity-bot-sign-and-deliver-mintauthv1-for-orchestrator-mode-mints)

---

<picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture> <picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture>
<sup>Need help on this PR? Tag `@codesmith-bot` with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added orchestrator-mode mint authorization with signing, persistence, and delivery acknowledgements.
  * Added stable per-mint nonces, signature reuse, idempotent delivery, and duplicate protection.
  * Added automatic retries for temporary delivery failures and operator alerts for unresolved authorizations.
  * Vault-direct mints continue without authorization delivery.
* **Bug Fixes**
  * Authorization events now update mint progress and timing without advancing lifecycle stages or reporting failures.
  * Improved recovery handling for interrupted authorization deliveries.

### DIFF
--- a/src/cli/rebalancing.rs
+++ b/src/cli/rebalancing.rs
@@ -22,6 +22,8 @@ use st0x_execution::{
     FractionalShares, Network, Symbol, TimeInForce,
 };
 use st0x_finance::Usdc;
+use st0x_issuance_client::IssuanceClient;
+use st0x_issuance_dto::VaultModeTag;
 use st0x_raindex::{RaindexService, RaindexVaultId};
 use st0x_tokenization::{
     AlpacaTokenizationService, IssuerRequestId, TokenizationRequest, TokenizationRequestStatus,
@@ -36,6 +38,7 @@ use crate::bot_gas::BotGasReceiptCostEnqueuer;
 use crate::equity_redemption::{
     DetectionFailure, EquityRedemption, EquityRedemptionCommand, RedemptionAggregateId,
 };
+use crate::mint_authorization::{ConfiguredMintAuthorizer, VaultModeReader};
 use crate::rebalancing::equity::{CrossVenueEquityTransfer, EquityTransferServices};
 use crate::rebalancing::to_wrapped_equities;
 use crate::rebalancing::usdc::{CrossVenueCashTransfer, UsdcSettlementParams, UsdcTransferError};
@@ -136,6 +139,14 @@ async fn build_equity_transfer_services(
         tokenizer: tokenization_service.clone(),
         wrapper: wrapper.clone(),
         bot_gas_enqueuer: BotGasReceiptCostEnqueuer::Disabled,
+        // The CLI transfer path never signs mint authorizations:
+        // `transfer_equity_command` rejects orchestrator-mode mints up
+        // front (`ensure_vault_direct_mint`), so only vault-direct assets
+        // reach this saga, which also skips `with_mint_authorization` and
+        // therefore never consults this value. Orchestrator-mode mints run
+        // through the server, where the authorizer and vault-mode wiring
+        // come from `[orchestrator]` config.
+        mint_authorizer: ConfiguredMintAuthorizer::Disabled,
     };
 
     let mint_store = StoreBuilder::<TokenizedEquityMint>::new(pool.clone())
@@ -157,6 +168,32 @@ async fn build_equity_transfer_services(
     );
 
     Ok(EquityTransferCliServices { transfer, wallet })
+}
+
+/// Refuses an orchestrator-mode CLI mint before any aggregate is created.
+///
+/// The CLI wires no vault-mode reader, mint authorizer, or delivery
+/// worker: an orchestrator-mode mint initiated here would sit unauthorized
+/// (stalled at polling) until a server restart resumed it. Fails closed --
+/// an indeterminate mode is never treated as vault-direct, mirroring the
+/// server saga's own rule.
+async fn ensure_vault_direct_mint(
+    vault_mode_reader: &dyn VaultModeReader,
+    symbol: &Symbol,
+) -> anyhow::Result<()> {
+    let mode = vault_mode_reader
+        .vault_mode(symbol)
+        .await
+        .context("could not determine the asset's vault mode; refusing to mint blind")?;
+
+    match mode {
+        VaultModeTag::VaultDirect => Ok(()),
+        VaultModeTag::Orchestrator => anyhow::bail!(
+            "{symbol} is orchestrator-mode: its mint needs a signed recipient \
+             authorization, which only the server wires (vault-mode reads, \
+             signing, delivery). Run this transfer through the server instead."
+        ),
+    }
 }
 
 pub(super) async fn transfer_equity_command<Writer: Write>(
@@ -183,6 +220,13 @@ pub(super) async fn transfer_equity_command<Writer: Write>(
 
     match direction {
         TransferDirection::ToRaindex => {
+            let issuance = IssuanceClient::new(
+                ctx.issuance.base_url.clone(),
+                ctx.issuance.api_key.header_value(),
+            )
+            .context("failed to build the issuance client for the vault-mode check")?;
+            ensure_vault_direct_mint(&issuance, symbol).await?;
+
             writeln!(stdout, "   Creating mint request...")?;
             writeln!(stdout, "   Receiving Wallet: {}", cli_services.wallet)?;
 
@@ -1784,6 +1828,7 @@ mod tests {
     use super::*;
     use crate::equity_redemption::{EquityRedemptionError, redemption_aggregate_id};
     use crate::inventory::ImbalanceThreshold;
+    use crate::mint_authorization::StubVaultModeReader;
     use crate::onchain::mock::MockRaindex;
     use crate::test_utils::setup_test_db;
     use crate::usdc_rebalance::{ReconcileReason, TransferRef, UsdcRebalanceCommand};
@@ -4068,6 +4113,7 @@ mod tests {
             tokenizer: Arc::new(MockTokenizer::new()),
             wrapper: Arc::new(MockWrapper::new()),
             bot_gas_enqueuer: BotGasReceiptCostEnqueuer::Disabled,
+            mint_authorizer: ConfiguredMintAuthorizer::Disabled,
         }
     }
 
@@ -4811,5 +4857,32 @@ mod tests {
             err_msg.contains("already failed"),
             "failing an already-failed redemption must refuse; got: {err_msg}"
         );
+    }
+
+    /// The CLI wires no authorization infrastructure, so an
+    /// orchestrator-mode mint must be refused before any aggregate exists
+    /// -- accepting it would strand the mint unauthorized until a server
+    /// restart resumed it.
+    #[tokio::test]
+    async fn cli_mint_refuses_orchestrator_mode_asset() {
+        let reader = StubVaultModeReader(VaultModeTag::Orchestrator);
+
+        let error = ensure_vault_direct_mint(&reader, &Symbol::new("RKLB").unwrap())
+            .await
+            .unwrap_err();
+
+        assert!(
+            error.to_string().contains("orchestrator-mode"),
+            "the refusal must say why and point at the server: {error}"
+        );
+    }
+
+    #[tokio::test]
+    async fn cli_mint_allows_vault_direct_asset() {
+        let reader = StubVaultModeReader(VaultModeTag::VaultDirect);
+
+        ensure_vault_direct_mint(&reader, &Symbol::new("RKLB").unwrap())
+            .await
+            .unwrap();
     }
 }
--- a/src/conductor.rs
+++ b/src/conductor.rs
@@ -69,6 +69,7 @@ use crate::equity_redemption::{
 use crate::inventory::{
     BroadcastingInventory, Inventory, InventoryProjection, InventorySnapshot, Venue,
 };
+use crate::mint_authorization::{ConfiguredMintAuthorizer, MintAuthorizationService};
 use crate::offchain::order::{
     ExecutorOrderPlacer, OffchainOrder, OffchainOrderId, OffchainOrderPlacement, OrderPlacer,
     PollOrderStatus, PollOrderStatusJobQueue, TerminalPositionFinalization,
@@ -99,7 +100,8 @@ use crate::position::{
 };
 use crate::position_check::{HedgeScanSkipReason, record_scan_skip};
 use crate::rebalancing::equity::{
-    CrossVenueEquityTransfer, EquityTransferServices, ResumeTokenizationAggregate,
+    CrossVenueEquityTransfer, DeliverMintAuthorizationCtx, DeliverMintAuthorizationJobQueue,
+    EquityTransferServices, MintAuthorizationWiring, ResumeTokenizationAggregate,
     ResumeTokenizationCtx, ResumeTokenizationJobQueue, ResumeTokenizationTarget,
     TransferEquityToHedging, TransferEquityToHedgingCtx, TransferEquityToMarketMaking,
     TransferEquityToMarketMakingCtx,
@@ -832,22 +834,7 @@ impl Conductor {
             seed_vault_registry_ctx,
         ) = setup_vault_registry(&pool, &apalis_pool, &ctx).await?;
 
-        // Grant one-time idempotent MAX approvals to the trusted spenders (our
-        // ERC-4626 wrapper vaults and the Raindex orderbook) before any worker
-        // or rebalancer runs, so wrap/deposit never reverts with
-        // ERC20InsufficientAllowance. Fails fast -- the bot must not come up
-        // healthy with these missing. Only runs when a wallet is configured;
-        // without one the bot cannot submit on-chain transactions at all.
-        grant_startup_token_approvals(&ctx).await?;
-
-        // Catch the lifecycle-failure read model up to the event log before
-        // its OffchainOrder reactor (registered below) goes live, in both
-        // modes. Must run BEFORE `PositionAndRebalancing::setup`: setup spawns
-        // resume workers that write the event store concurrently, and this
-        // catch-up's deferred read-then-write transaction surfaces
-        // SQLITE_BUSY_SNAPSHOT immediately (busy_timeout does not apply to
-        // deferred-upgrade conflicts), failing the whole boot.
-        catch_up_lifecycle_failures(&pool).await?;
+        run_startup_maintenance(&ctx, &pool).await?;
 
         let rebalancing = optional_rebalancing_ctx(&ctx)?;
         let notifier = build_alert_notifier(ctx.alerts.as_ref(), "Operational alerting")?;
@@ -871,6 +858,8 @@ impl Conductor {
             transfer_equity_to_market_making_ctx,
             transfer_equity_to_hedging_ctx,
             resume_tokenization_queue,
+            deliver_mint_authorization_queue,
+            deliver_mint_authorization_ctx,
         } = PositionAndRebalancing::setup(
             rebalancing,
             RebalancingDeps {
@@ -1024,6 +1013,8 @@ impl Conductor {
             .seed_vault_registry_ctx(seed_vault_registry_ctx)
             .resume_tokenization_queue(resume_tokenization_queue)
             .maybe_resume_tokenization_ctx(resume_tokenization_ctx)
+            .deliver_mint_authorization_queue(deliver_mint_authorization_queue)
+            .maybe_deliver_mint_authorization_ctx(deliver_mint_authorization_ctx)
             .record_bot_gas_receipt_cost_queue(record_bot_gas_receipt_cost_queue)
             .maybe_record_bot_gas_receipt_cost_ctx(record_bot_gas_receipt_cost_ctx)
             .job_cleanup(job_cleanup)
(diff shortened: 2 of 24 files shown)


rouzwelt:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1145?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1209** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1209?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1202** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1202?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1151** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1151?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1150** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1150?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1145** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1145?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1145?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1141** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1141?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1139** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1139?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1138** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1138?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1130** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1130?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1129** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1129?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1128** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1128?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1115** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1115?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1114** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1114?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1113** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1113?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1145).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

rouzwelt:
@coderabbit review

src/conductor.rs
```diff
@@ -1769,6 +1770,128 @@ fn build_alert_notifier(
...
+        wiring: MintAuthorizationWiring {
+            vault_mode_reader: issuance_client.clone(),
+            token_addresses,
+            delivery_queue: queue.clone(),
+        },
+        queue,
+        issuance_client,
+    })
+}
+
```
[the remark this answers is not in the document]
rouzwelt: both branches are already covered end-to-end — the orchestrator e2e configures [orchestrator] and exercises the Enabled path through a real signed delivery, and every other e2e boots without the section through Disabled. A unit test of this function would require constructing a full Ctx to assert DI plumbing (queue clones reaching three fields), which the compiler already guarantees structurally. Not worth the fixture weight.

src/conductor.rs
```diff
@@ -2057,12 +2180,16 @@ fn spawn_rebalancing_infrastructure<Chain: Wallet + Clone>(
...
+        let mint_authorization =
+            build_mint_authorization_infra(&deps.ctx, &deps.apalis_pool, &base_wallet).await?;
+
         let equity_transfer_services = EquityTransferServices {
             raindex: raindex_service.clone(),
             vault_lookup: vault_lookup.clone(),
             tokenizer: tokenizer.clone(),
             wrapper: wrapper.clone(),
             bot_gas_enqueuer: bot_gas_enqueuer.clone(),
+            mint_authorizer: mint_authorization.authorizer,
```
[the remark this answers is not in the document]
rouzwelt: The cited learning doesn't match the current codebase: with_bot_gas_enqueuer is itself still a builder method on this exact struct today, so the live precedent is builder-with-explicit-absence-default. This PR now strengthens that shape — absence became an explicit ConfiguredMintAuthorization::VaultDirectOnly variant (mirroring BotGasReceiptCostEnqueuer::Disabled) rather than None. Switching to a constructor parameter would touch ~28 construction sites for consistency with a convention the code doesn't follow; happy to do that as its own sweep covering bot-gas too if we want to adopt it for real.

src/tokenized_equity_mint.rs
```diff
@@ -1577,6 +1780,9 @@ impl EventSourced for TokenizedEquityMint {
                     tokenization_request_id,
                     requested_at,
                     accepted_at,
+                    // The authorization served its purpose once tokens
+                    // arrive; later states deliberately do not carry it.
+                    authorization: _,
```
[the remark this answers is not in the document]
rouzwelt: The analysis is against an older revision. transition_record_authorization_delivered no longer requires a Signed state — a post-advance acknowledgement records unconditionally as an audit fact (see recording_delivery_after_mint_advanced_still_records). And the job's post-advance no-op is correct, not a gap: the mint can only reach TokensReceived after issuance minted, which it does only once the authorization is already in its hands — so there is nothing left to deliver, and dropping the payload with the state is intentional.

src/cli/rebalancing.rs
```diff
@@ -136,6 +137,17 @@ async fn build_equity_transfer_services(
...
+        // so the saga cannot read `vault_mode` and treats every asset as
+        // vault-direct (with a warning) -- a CLI mint of an
+        // orchestrator-mode asset therefore proceeds unauthorized and
+        // stalls at polling until the authorization is delivered by the
+        // server's resume path. This `Disabled` value is only consulted if
+        // signing is ever reached, which the missing wiring prevents.
+        mint_authorizer: ConfiguredMintAuthorizer::Disabled,
     };
 
     let mint_store = StoreBuilder::<TokenizedEquityMint>::new(pool.clone())
```
0xgleb, line 152: should fix: This CLI path can initiate an orchestrator-mode mint with no vault-mode reader, signer, or delivery queue. Reject the command before acceptance or wire it through the same authorization infrastructure as the server.

src/rebalancing/equity/mod.rs
```diff
@@ -681,6 +753,117 @@ impl CrossVenueEquityTransfer {
...
+                    )
+                    .await?;
+
+                // Bounds delivery to ONE live chain per mint: redrive
+                // successors carry no idempotency key, so without this
+                // check a resume during a delayed-redrive window would
+                // start a second chain. Fails open on a guard read error
+                // (warn + push): a duplicate chain delivers byte-identical
+                // payloads and is benign, while a skipped push would stall
+                // the delivery.
```
0xgleb, line 818: minor: Propagate failures from the active-delivery lookup instead of treating them as `false`. The check and enqueue sequence can still race, so a per-mint uniqueness or idempotency guard would make the intended one-live-chain invariant durable, even though duplicate chains deliver the same payload.

src/conductor.rs
```diff
@@ -1769,6 +1770,128 @@ fn build_alert_notifier(
...
+    /// One issuance client serves both mint-authorization consumers (the
+    /// saga's vault-mode read and the delivery job), built from the same
+    /// `[issuance]` credentials as the freeze guard's own client instance.
+    issuance_client: Arc<IssuanceClient>,
+}
+
+/// Builds [`MintAuthorizationInfra`]. The authorizer is `Enabled` only when
+/// `[orchestrator]` is configured: while every asset is vault-direct the bot
+/// deploys dark without the section, and an orchestrator-mode mint reaching
+/// the signing step then fails loudly rather than guessing an address.
```
0xgleb, line 1889: should fix: Continuing after orphan repair fails is unsafe because a stale `Running` row is then considered live and suppresses resume enqueueing. Fail startup or recover stale rows by lease age before allowing that suppression.

src/rebalancing/equity/authorization_job.rs
```diff
@@ -0,0 +1,648 @@
...
+/// alone cannot see them -- without this bound, a restart during a
+/// delayed-redrive window would start a second delivery chain. Chains
+/// deliver byte-identically (issuance-idempotent), so a duplicate is
+/// benign rather than dangerous; this mirrors the one-poll-job-per-order
+/// bound on `PollOrderStatus`. "Live" matches `requeue_orphaned`'s
+/// in-flight definition: `Pending`, `Queued` (claimed, not yet running),
+/// or `Running`.
+pub(crate) async fn has_live_delivery_job(
+    queue: &DeliverMintAuthorizationJobQueue,
+    issuer_request_id: &IssuerRequestId,
```
0xgleb, line 309: should fix: `Conflict` and `Rejected` currently leave only logs and an alert, so startup recovery can enqueue the same non-retryable delivery again. Persist a parked outcome before completing the job, exclude it from automatic recovery, and cover both branches.

src/rebalancing/equity/authorization_job.rs
```diff
@@ -0,0 +1,648 @@
...
+
+/// Alert delivery failures are logged, never allowed to mask the job
+/// outcome (mirrors the USDC transfer jobs' alert handling).
+async fn notify_swallowing_failure(ctx: &DeliverMintAuthorizationCtx, message: &str) {
+    if let Err(alert_error) = ctx.notifier.notify(message).await {
+        warn!(
+            target: "tokenization",
+            ?alert_error,
+            "Failed to deliver mint-authorization operator alert"
+        );
```
0xgleb, line 360: should fix: Returning this error after exhausting redrives lets the generic worker retry the same exhausted payload and repeat the alert. Persist exhaustion and complete successfully, or make this failure non-retryable with durable alert deduplication.

src/tokenized_equity_mint.rs
```diff
@@ -2137,6 +2351,175 @@ impl EventSourced for TokenizedEquityMint {
...
+            }
+            | Self::WrapSubmitted {
+                issuer_request_id, ..
+            }
+            | Self::TokensWrapped {
+                issuer_request_id, ..
+            }
+            | Self::VaultDepositSubmitted {
+                issuer_request_id, ..
+            }
```
0xgleb, line 2497: minor: Retain the authorization identity, or include it in `RecordMintAuthorizationDelivered`, so a late acknowledgement can be checked against the mint that was signed. The current later-state arms accept the acknowledgement after authorization progress is dropped, which weakens audit provenance but does not itself alter issuance or custody.

src/tokenized_equity_mint.rs
```diff
@@ -2137,6 +2351,175 @@ impl EventSourced for TokenizedEquityMint {
...
+                let nonce = B256::random();
+                let signed = services
+                    .mint_authorizer
+                    .sign(token, *quantity, nonce)
+                    .await
+                    .map_err(
+                        |error| TokenizedEquityMintError::AuthorizationSigningFailed {
+                            error_message: error.to_string(),
+                        },
+                    )?;
```
0xgleb, line 2397: minor: Preserve `MintAuthorizationError` as a typed source instead of converting it to a string. This keeps configuration, scaling, and EVM failures distinguishable and preserves the source chain.

src/conductor/builder.rs
```diff
@@ -1195,6 +1214,37 @@ fn register_resume_tokenization_worker(
     })
```
0xgleb, line 1240: minor: Add a builder-level regression test covering a delivery context with a queued authorization job, plus the intentional `None` context case. The registration change is straightforward, so this is coverage hardening rather than evidence that the worker is currently omitted.

0xgleb (changes requested):

src/rebalancing/equity/mod.rs
rouzwelt: Fail-open is deliberate and the asymmetry is the reason: a duplicate chain delivers byte-identical payloads that issuance treats as an idempotent 200, so the worst case of proceeding is benign, while propagating a transient DB read error would fail the transfer step and stall the mint. Durable uniqueness already exists where it matters — the initial enqueue is push_idempotent keyed on the issuer request id, so the check-then-enqueue race can only ever duplicate a redrive successor, which is the benign case above.

src/rebalancing/equity/authorization_job.rs
rouzwelt: Excluding parked mints from automatic recovery would break the resolution path: resolution happens on the issuance side, and the resume-driven flow is how this bot observes it (polling into TokensReceived). Redelivery after a park is byte-identical and issuance-idempotent — a still-unresolved 409 just re-parks with one alert per restart, which is deliberate: an unresolved park should resurface rather than go quiet. Persisting a parked flag in the aggregate buys alert dedup across restarts at the cost of an event-schema addition plus a mechanism to clear it once issuance resolves — happy to file it as a follow-up if restart-frequency alert noise shows up in practice.

src/tokenized_equity_mint.rs
rouzwelt: Provenance lives in the event stream, not the state: MintAuthorizationSigned (nonce + signature) is permanently in the same per-mint stream the late MintAuthorizationDelivered lands in, so the correlation is the stream identity itself. There's also no cross-mint path to guard against — the ack is dispatched by the delivery job keyed on this aggregate's id, delivering the authorization it just loaded from this aggregate. Adding the nonce to the ack event would record a value the handler has nothing to check it against once state has advanced (the authorization is dropped precisely because issuance holds it).

src/tokenized_equity_mint.rs
rouzwelt: Agreed in principle — this is the one place the repo's no-string-errors rule bends, and it's a framework constraint, not a choice: TokenizedEquityMintError must satisfy the DomainError serde bound, and MintAuthorizationError wraps EvmError (non-serde transport internals), so the typed source can't be embedded. Same documented constraint as the neighboring RequestFailed (see the variant's doc comment). The rendered message preserves the full Display chain (config/scaling/EVM failures stay distinguishable in logs), and no caller branches on the variant.

src/conductor/builder.rs
rouzwelt: None of the six conditionally-registered sibling workers has a registration-level test, for a structural reason: apalis Monitor doesn't expose its registrations for assertion, so such a test could only re-assert the guard clause's own code. The Some path is exercised end-to-end in the stacked e2e PR (the delivery worker actually consumes the queue and the mint completes), and the None arm is a debug-log guard identical to its five siblings. Happy to add it if we introduce a registration-introspection seam, but that's a builder-wide change, not specific to this worker.

0xgleb (approved):

JuaniRios (approved):

PR #1153, opened 2026-08-09
config(prod): cut the liquidity bot over to the prod RaindexInventory

hardyjosh:
## What

Cuts the **prod** liquidity bot over to its own `RaindexInventory`. RAI-1610 / RAI-1694.

Blocked on RAI-1694 (the `OPERATOR_ROLE` grants) — **draft until those land**, because the bot's startup preflight fails closed without them.

## Why

Prod has been `inventory_mode = "legacy"` the whole time. Only **staging** was ever cut over — the 31 Jul Bebop pilot fill ran through staging's bot against staging's inventory (`0x5FE36e6b…`). The prod bot has never settled through a shared inventory at all, so this is a real cutover, not a flag flip.

## How

Points prod at a freshly deployed inventory + hook (Base, 2026-08-09):

| | |
|---|---|
| `RaindexInventory` | `0x10e4db39275C3b128C01bA1194D45D19aE1520d9` |
| `RaindexBebopHook` | `0x3EE87Dcabfb324B2231044928080Dd4E6AC19baE` |

Deliberately **not** staging's `0x5FE36e6b…`. One inventory across both environments would put staging's whitelisted hook and prod funds in a single blast radius.

`vault_owner` moves from the bot wallet to the inventory because Raindex forbids delegated order management — orders placed *through* the inventory are owned *by* it, so `vaultBalance2` reads, the vault registry and order-owner fill matching all scope to the inventory.

`inventory_adapters` carries **bebop only** — no prod univ4 hook exists yet (M4). Also drops a stale comment block referencing `0x6b7B523f…`, a third and older inventory that was never prod's.

## Testing

- `cargo nextest run -p st0x-config -E 'test(all_repo_config_tomls_are_valid)'` → **passes**
- 4 unrelated failures in `st0x-config` (`unknown variant `private-key`, there are no variants` in `WalletConfig`) reproduce on clean master with this change stashed — pre-existing, not from this PR

## Anything else

**@JuaniRios99 — two things to check, both outside what I could settle from the code:**

1. **`deployment_block = 43259507` is unchanged.** The new inventory was deployed today (~block 49747000), so it has no order events before then. Leaving the old value means backfilling ~6.5M blocks of history for an owner that didn't exist. Harmless or wasteful? Your call — I didn't want to guess at the semantics.

2. **Vault migration + funding isn't in this PR.** This only repoints the bot. Moving/creating the vaults under the new inventory and funding both legs is RAI-1416, and needs to happen before there's anything to trade against.

⚠️ The bot's `verify_operator_role()` preflight checks `hasRole(OPERATOR_ROLE, wallet)` **literally** — it does not accept `DEFAULT_ADMIN_ROLE` as a substitute, even though `deposit4`/`withdraw4` are `onlyAdminOrOperator` at the contract level. So the Turnkey wallet needs an explicit `OPERATOR_ROLE` grant despite already being admin. That's grant 2 in RAI-1694, and it's the one most likely to get skipped as "redundant".

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013PKAcAsfiPcBTwQ3pMGQUC

## Summary by CodeRabbit

* **Configuration**
  * Updated production inventory management and deployment settings.
  * Added support for dedicated inventory handling and Bebop integration.
  * Updated vault ownership and deployment alignment.
  * Corrected the USDC cash vault identifier to support accurate rebalancing.

### DIFF
--- a/config/prod/st0x-hedge.toml
+++ b/config/prod/st0x-hedge.toml
@@ -43,25 +43,20 @@ orderbook = "0xe522cB4a5fCb2eb31a52Ff41a4653d85A4fd7C9D"
 # with bot-EOA-owned vaults) or "managed" (settle through the shared
 # RaindexInventory below, which the bot must operate via OPERATOR_ROLE).
 #
-# Kept "legacy" until the RAI-1198 cutover: the shared inventory
-# (0x6b7B523fADd1677413AD92c9404C8F0796bACF6F, admin 0x679dF30b...) exists, but
-# the bot's EOA does not hold OPERATOR_ROLE on it yet, so enabling "managed"
-# now would fail the startup preflight and block the whole bot. At cutover,
-# switch this to "managed", set `inventory` to the migrated inventory, and flip
-# `vault_owner` below to that same address in the change that grants the bot
-# OPERATOR_ROLE and migrates the vaults.
-inventory_mode = "legacy"
-# Legacy mode has no shared inventory, so adapter settlements are impossible.
-inventory_adapters = []
-# REQUIRED. Owner of the Raindex orders/vaults on-chain -- the key every
-# vaultBalance2 read, vault-registry entry, and order-owner fill match is
-# scoped by. Currently the bot's Turnkey wallet (= [wallet].address below),
-# because the vaults are bot-EOA-owned pre-migration. Once RAI-1198 moves the
-# vaults under the inventory contract, flip this to the `inventory` address
-# above.
-vault_owner = "0xA9C16673F65AE808688cB18952AFE3d9658C808f"
-# Block number to start backfilling order events from.
-deployment_block = 43259507
+# Dedicated production inventory. The wallet must hold OPERATOR_ROLE because
+# startup verifies that role explicitly, even when the wallet is also admin.
+inventory_mode = "managed"
+# Operator addresses that attribute inventory settlements to venues.
+inventory_adapters = [
+  { venue = "bebop", operator = "0x3EE87Dcabfb324B2231044928080Dd4E6AC19baE" },
+]
+inventory = "0x10e4db39275C3b128C01bA1194D45D19aE1520d9"
+# REQUIRED. Owner used for vault balances, registry entries, and order-fill
+# matching. Managed mode requires this to equal `inventory`.
+vault_owner = "0x10e4db39275C3b128C01bA1194D45D19aE1520d9"
+# Backfill floor. No inventory-owned event can predate its deployment block; an
+# existing later checkpoint still resumes at checkpoint + 1.
+deployment_block = 49746968
 # Block confirmations required before a submitted transaction is treated as
 # settled. Does not govern fill ingestion (controlled by ingestion_cutoff).
 required_confirmations = 3
@@ -139,8 +134,9 @@ realert_interval = 43200
 
 # USDC vault and rebalancing settings.
 [assets.cash]
-# Raindex vault ID for USDC deposits/withdrawals.
-vault_id = "0xfab2"
+# USDC vault used by Bebop and cash rebalancing. The inventory keys vaults by
+# (token, vaultId), so both paths must use the same vault.
+vault_id = "0xfab"
 # Whether automatic USDC rebalancing between venues is active.
 rebalancing = "enabled"
 # Maximum USDC moved per rebalance operation.


hardyjosh:
* **#1153** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1153?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1153?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1153).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

config/prod/st0x-hedge.toml
```diff
@@ -43,25 +43,39 @@ orderbook = "0xe522cB4a5fCb2eb31a52Ff41a4653d85A4fd7C9D"
...
+# Prod settles through its own dedicated inventory (deployed 2026-08-09,
+# RAI-1610) -- deliberately NOT the one staging uses. Sharing an inventory
+# across environments would put staging's whitelisted hook and prod funds in
+# one blast radius.
+#
+# The Turnkey wallet (= [wallet].address below) holds DEFAULT_ADMIN_ROLE on
+# this contract AND is granted OPERATOR_ROLE (RAI-1694). Admin alone is enough
+# at the contract level -- deposit4/withdraw4 are onlyAdminOrOperator -- but
+# verify_operator_role() checks hasRole(OPERATOR_ROLE) literally and fails
+# closed, so the explicit grant is required for the bot to start at all.
```
JuaniRios: [low] This is too much permanent context for the config. Can we reduce it to: `Dedicated production inventory. The wallet must hold OPERATOR_ROLE because startup verifies that role explicitly, even when the wallet is also admin.`

config/prod/st0x-hedge.toml
```diff
@@ -43,25 +43,39 @@ orderbook = "0xe522cB4a5fCb2eb31a52Ff41a4653d85A4fd7C9D"
...
+inventory_mode = "managed"
+# Public OPERATOR_ROLE holders for venue attribution. Bebop only: there is no
+# prod univ4 hook yet (that lands with M4).
```
JuaniRios: [low] `no prod univ4 hook yet` and `lands with M4` are mutable roadmap state, so this comment will go stale. I'd keep only the stable meaning: these operator addresses attribute inventory settlements to venues.

config/prod/st0x-hedge.toml
```diff
@@ -43,25 +43,39 @@ orderbook = "0xe522cB4a5fCb2eb31a52Ff41a4653d85A4fd7C9D"
...
-# because the vaults are bot-EOA-owned pre-migration. Once RAI-1198 moves the
-# vaults under the inventory contract, flip this to the `inventory` address
-# above.
-vault_owner = "0xA9C16673F65AE808688cB18952AFE3d9658C808f"
-# Block number to start backfilling order events from.
-deployment_block = 43259507
+# scoped by. Managed mode scopes all three to the inventory contract, not the
+# bot's Turnkey wallet: Raindex forbids delegated order management, so orders
+# placed through the inventory are owned by the inventory itself.
+vault_owner = "0x10e4db39275C3b128C01bA1194D45D19aE1520d9"
```
JuaniRios, line 56: [low] This can be much shorter. The useful invariant is: `Owner used for vault balances, registry entries, and order-fill matching. Managed mode requires this to equal inventory.`

config/prod/st0x-hedge.toml
```diff
@@ -43,25 +43,39 @@ orderbook = "0xe522cB4a5fCb2eb31a52Ff41a4653d85A4fd7C9D"
...
+# Block number to start backfilling order events from. Acts as a floor:
+# backfill resumes at max(checkpoint + 1, deployment_block), and a cold start
+# with no checkpoint begins here.
+#
+# Moved to the block the inventory was deployed in. `vault_owner` above is now
+# the inventory, and an inventory-owned order cannot exist before the contract
+# did -- so nothing is skipped, and a cold start no longer scans ~6.5M blocks
+# of Base history that cannot contain a single relevant event. No effect on the
+# running bot, whose checkpoint is already well ahead of this.
```
JuaniRios: [low] `nothing is skipped` and `No effect on the running bot` are too categorical. A later OrderBook checkpoint still skips earlier inventory settlements. Please keep the stable behavior only: `Backfill floor. No inventory-owned event can predate its deployment block; an existing later checkpoint still resumes at checkpoint + 1.`

config/prod/st0x-hedge.toml
```diff
@@ -140,7 +154,20 @@ realert_interval = 43200
...
+# inventory keys vaults by (token, vaultId) -- so USDC under 0xfab and USDC
+# under 0xfab2 are two different vaults. Leaving cash on 0xfab2 would have the
+# bot rebalancing a vault the hook never touches, while the vault the hook
+# actually draws on and deposits into goes unmanaged.
+#
+# 0xfab is the proven managed configuration: staging runs `[assets.cash]
+# vault_id = "0xfab"` and its inventory holds USDC on 0xfab with 0xfab2 empty.
+# Under the legacy prod bot EOA, USDC is currently funded on BOTH ids -- 0xfab
+# as the orders' counter-vault, 0xfab2 as the separate rebalancing pot. The
+# migration consolidates onto 0xfab.
```
JuaniRios: [low] This cites `config/t0-bebop.toml` as if it were local, but it lives in `st0x.bebop`. It also embeds live balance state that becomes stale after migration. Can we reduce this to: `USDC vault used by Bebop and cash rebalancing. Inventory keys vaults by (token, vaultId), so both paths must use the same vault.`

JuaniRios (changes requested):

hardyjosh:
Thanks @JuaniRios — all five applied, and one of them caught something real beyond the wording.

**`inventory_mode`** — taken verbatim:
> Dedicated production inventory. The wallet must hold OPERATOR_ROLE because startup verifies that role explicitly, even when the wallet is also admin.

**`inventory_adapters`** — dropped the roadmap state, kept the stable meaning: *"Operator addresses that attribute inventory settlements to venues."*

**`vault_owner`** — taken verbatim.

**`deployment_block`** — you were right that `nothing is skipped` was wrong, not just over-stated. A later OrderBook checkpoint does skip earlier inventory settlements, because the checkpoint is keyed by OrderBook and survives the inventory change. Now reads exactly as you suggested:
> Backfill floor. No inventory-owned event can predate its deployment block; an existing later checkpoint still resumes at checkpoint + 1.

That leaves the Path A/B checkpoint decision genuinely open at cutover time — runbook §6. Staging took Path A.

**`[assets.cash]`** — you're right that it cited `config/t0-bebop.toml` as if local (it's in `st0x.bebop`) and embedded balance state that goes stale on migration. Now: *"USDC vault used by Bebop and cash rebalancing. The inventory keys vaults by (token, vaultId), so both paths must use the same vault."*

---

**Separately — the branch was 3 commits behind and would have reverted [#1154](https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1154).** `reserved = 250` (the USDC conversion collar holdback) was being dropped by the stale base. Restacked onto master, so the diff is now only the intended change. Worth a second look at that specifically.

`all_repo_config_tomls_are_valid` and `repo_config_vault_owner_matches_settlement_mode` both pass.

CodeRabbit's point about not claiming OPERATOR_ROLE is granted is resolved by your wording — it now states the requirement rather than the status. For the record both grants are live on-chain as of today (RAI-1694), hook and wallet, but the comment no longer depends on that being true.

JuaniRios (approved):

PR #1169, opened 2026-08-11
fix: renew Tailscale certificates automatically

JuaniRios:
## What

- Fix [RAI-1708](https://linear.app/makeitrain/issue/RAI-1708/make-tailscale-tls-certificates-renew-automatically).
- Let the persistent daily timer start `tailscale-cert` more than once.
- Reload nginx after each successful renewal so it serves the new certificate.

## Why

- `RemainAfterExit = true` left the oneshot service active after it finished.
- Later timer starts did nothing, so certificate renewal required a deployment or manual restart.

## How

- Set `RemainAfterExit = false` in the shared Tailscale module for prod and staging.
- Add `validate-tailscale-cert-renewal.sh` to check the service lifecycle, timer schedule, timer activation, persistence, and nginx reload.
- Run the new check in CI and document the renewal contract in `SPEC.md`.

## Testing

- `./scripts/validate-tailscale-cert-renewal.sh`
- `nix develop .#ci-hooks -c pre-commit run --files .github/workflows/ci.yaml SPEC.md nix/tailscale.nix scripts/validate-tailscale-cert-renewal.sh`
- `nix develop .#ci-backend -c cargo check --workspace --all-features`
- `nix develop .#ci-backend -c cargo clippy --workspace --all-targets --all-features`
- `nix develop .#ci-backend -c cargo fmt -- --check`
- GitHub CI passed, including all four backend test shards.

## Screenshots

N/A

## Anything else

One deployment is required to install the corrected systemd unit. After that, renewals need no deployment or manual restart.

---
<a href="https://app.blacksmith.sh/ST0x-Technology/codesmith/st0x.liquidity/pr/1169"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789052282&installation_model_id=19370&pr_number=1169&repository=ST0x-Technology%2Fst0x.liquidity&return_to=https%3A%2F%2Fgithub.com%2FST0x-Technology%2Fst0x.liquidity%2Fpull%2F1169&signature=02d9305c5125802b736ca30aecc5a2212bb1661e6b7177b2077efc26d6c5cf43"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/nix/tailscale.nix
+++ b/nix/tailscale.nix
@@ -69,13 +69,13 @@ in
         ];
         serviceConfig = {
           Type = "oneshot";
-          RemainAfterExit = true;
+          # The daily timer can only restart a oneshot that returns to inactive.
+          RemainAfterExit = false;
           User = "nginx";
           Group = "nginx";
-          # Reload nginx so it picks up renewed certs on subsequent runs
-          # triggered by the timer. || true keeps the unit green on first
-          # boot when nginx hasn't started yet.
-          ExecStartPost = "+${pkgs.bash}/bin/bash -c 'systemctl is-active --quiet nginx.service && systemctl reload nginx.service || true'";
+          # Reload nginx after timer-triggered renewals. An inactive nginx is
+          # expected on first boot; an attempted reload failure is not.
+          ExecStartPost = "+${pkgs.bash}/bin/bash -c 'if systemctl is-active --quiet nginx.service; then systemctl reload nginx.service; fi'";
         };
         script = ''
           set -euo pipefail
@@ -96,7 +96,7 @@ in
       };
 
       nginx.after = [ "tailscale-cert.service" ];
-      nginx.requires = [ "tailscale-cert.service" ];
+      nginx.wants = [ "tailscale-cert.service" ];
     };
 
     timers.tailscale-cert = {
--- a/scripts/validate-tailscale-cert-renewal.nu
+++ b/scripts/validate-tailscale-cert-renewal.nu
@@ -1,0 +1,82 @@
+#!/usr/bin/env nu
+
+const CONFIGURATIONS = [
+  "st0x-liquidity"
+  "st0x-liquidity-staging"
+]
+
+const SELECT_RENEWAL_CONFIG = '{ systemd, ... }: {
+  type = systemd.services.tailscale-cert.serviceConfig.Type;
+  remainAfterExit = systemd.services.tailscale-cert.serviceConfig.RemainAfterExit;
+  onCalendar = systemd.timers.tailscale-cert.timerConfig.OnCalendar;
+  persistent = systemd.timers.tailscale-cert.timerConfig.Persistent;
+  wantedBy = systemd.timers.tailscale-cert.wantedBy;
+  execStartPost = systemd.services.tailscale-cert.serviceConfig.ExecStartPost;
+  nginxWants = systemd.services.nginx.wants;
+  nginxRequires = systemd.services.nginx.requires;
+}'
+
+def evaluate-config [configuration: string]: nothing -> record {
+  let attribute = $".#nixosConfigurations.($configuration).config"
+  let result = (^nix eval --json --apply $SELECT_RENEWAL_CONFIG $attribute | complete)
+
+  if $result.exit_code != 0 {
+    error make {
+      msg: $"failed to evaluate ($attribute): ($result.stderr | str trim)"
+    }
+  }
+
+  $result.stdout | from json
+}
+
+def assert-config [configuration: string, option: string, actual: any, expected: any] {
+  if $actual != $expected {
+    error make {
+      msg: $"($configuration) expected ($option)=($expected | to nuon), got ($actual | to nuon)"
+    }
+  }
+}
+
+def assert-nginx-reload [configuration: string, exec_start_post: any] {
+  let command_text = ($exec_start_post | to text)
+
+  if not ($command_text | str contains "systemctl reload nginx.service") {
+    error make {
+      msg: $"($configuration) certificate renewal does not reload nginx"
+    }
+  }
+
+  if ($command_text | str contains "|| true") {
+    error make {
+      msg: $"($configuration) certificate renewal swallows nginx reload failures"
+    }
+  }
+}
+
+def assert-nginx-dependency [configuration: string, wants: list<string>, requires: list<string>] {
+  if "tailscale-cert.service" not-in $wants {
+    error make {
+      msg: $"($configuration) nginx does not want tailscale-cert.service"
+    }
+  }
+
+  if "tailscale-cert.service" in $requires {
+    error make {
+      msg: $"($configuration) nginx hard-requires tailscale-cert.service"
+    }
+  }
+}
+
+def main [] {
+  for configuration in $CONFIGURATIONS {
+    let config = (evaluate-config $configuration)
+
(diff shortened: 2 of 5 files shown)


JuaniRios:
* **#1169** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1169?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1169?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1169).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

nix/tailscale.nix
agryaznov: *_low_*: The comment says the `|| true` exists so first boot stays green "when nginx hasn't started yet", but the `systemctl is-active --quiet nginx.service &&` guard already covers that case — when nginx is down the `&&` short-circuits and the `|| true` rescues the exit code. What the `|| true` additionally swallows is a reload that was attempted and *failed*. That is the one case worth knowing about: the cert was renewed on disk but nginx is still serving the old one, and the unit reports success.

**Fix:** make the guard structural so only the intended case is tolerated — `if systemctl is-active --quiet nginx.service; then systemctl reload nginx.service; fi` — which stays green when nginx is down and fails loudly when a real reload fails.

agryaznov: *_medium_*: `nginx.requires = [ "tailscale-cert.service" ]` combined with `nginx.after` was previously satisfied by an already-active unit: once the oneshot had run, `RemainAfterExit=true` kept it `active (exited)`, so restarting nginx did not re-run it. With `RemainAfterExit=false` the unit sits `inactive` between timer runs, so `Requires=` now pulls it in on every nginx start — including deploys that restart nginx without rebooting.

That matters because the unit's script blocks on `until BackendState = Running` for up to 60 seconds and then `exit 1`s. So an nginx restart is now gated on tailscaled being healthy, and a tailscaled hiccup blocks nginx from starting at all — where previously, post-first-boot, the two were independent. The failure mode is inverted from what you want: the existing cert on disk is still valid for weeks, so a renewal problem should never prevent serving it.

**Fix:** change nginx's dependency from `requires` to `wants` (keeping `after`), so ordering is preserved but a cert-unit failure degrades to serving the still-valid existing cert rather than taking the dashboard down. If a hard dependency is wanted for the genuine first-boot case, split provisioning from renewal into two units so only provisioning is required.

agryaznov, line 102: *_medium_*: The original bug was silent: the timer fired daily, did nothing, and nobody noticed until the certificate expired. This PR restores renewal but not observability — if `tailscale cert` fails on the daily run the unit just enters `failed`, and nothing in the repo watches host unit state. The same 90-day silent countdown to an expired cert is still reachable, only via a different path.

The added validation script guards the *configuration*, which is valuable, but it runs in CI and cannot observe the deployed host.

**Fix:** add an `OnFailure=` unit that routes to the existing alerting (the repo already has a Telegram notifier under `[alerts]`), or export cert expiry so it can be alerted on. Without one of the two, the next failure is found the same way as the last one.

scripts/validate-tailscale-cert-renewal.nu
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env nu
+
+const CONFIGURATIONS = [
+  "st0x-liquidity"
+  "st0x-liquidity-staging"
+]
+
+def evaluate-json [configuration: string, option: string]: nothing -> any {
+  let attribute = $".#nixosConfigurations.($configuration).config.($option)"
+  let result = (^nix eval --json $attribute | complete)
```
agryaznov: *_low_*: `evaluate-json` shells out to `nix eval --json .#nixosConfigurations.<cfg>.config.<option>` once per option, and `main` (lines 43-49) calls it six times per configuration. That is twelve evaluations of the full NixOS module system per CI run to read six facts.

**Fix:** evaluate one attrset per configuration — a single `nix eval --json` returning `{ type, remainAfterExit, onCalendar, persistent, wantedBy, execStartPost }` — and assert against its fields.

scripts/validate-tailscale-cert-renewal.nu
```diff
@@ -0,0 +1,51 @@
...
+      msg: $"($configuration) expected ($option)=($expected | to nuon), got ($actual | to nuon)"
+    }
+  }
+}
+
+def assert-nginx-reload [configuration: string] {
+  let option = "systemd.services.tailscale-cert.serviceConfig.ExecStartPost"
+  let exec_start_post = (evaluate-json $configuration $option)
+
+  if not ($exec_start_post | str contains "systemctl reload nginx.service") {
```
agryaznov: *_low_*: `$exec_start_post | str contains "systemctl reload nginx.service"` works here because `ExecStartPost` is set as a plain string in `nix/tailscale.nix:79`, but NixOS accepts a list for that option equally, and `nix eval --json` would then return a JSON array — at which point `str contains` no longer means what the assertion assumes. The check would change behaviour on a refactor that is otherwise legitimate.

**Fix:** normalise before matching (`$exec_start_post | to text`, or join a list into a single string), so the assertion holds for both accepted shapes.

agryaznov (approved):
Left some suggestions inline, none of them being a blocker

nix/tailscale.nix
JuaniRios: Fixed in `8288a5f1`. The post-start hook now skips reload only when nginx is inactive. A failed attempted reload fails the unit.

JuaniRios: Fixed in `8288a5f1`. nginx now uses `wants` plus `after`, so renewal stays ordered without blocking nginx when renewal fails.

JuaniRios: Good catch. Host-level systemd failures are not wired into the app notifier atm, so I kept that out of this focused renewal fix. Tracked in [RAI-1728](https://linear.app/makeitrain/issue/RAI-1728/alert-when-tailscale-certificate-renewal-fails).

scripts/validate-tailscale-cert-renewal.nu
JuaniRios: Fixed in `8288a5f1`. The validator now evaluates one attrset per configuration, reducing full NixOS evaluations from twelve to two.

scripts/validate-tailscale-cert-renewal.nu
JuaniRios: Fixed in `8288a5f1`. `ExecStartPost` is normalized with `to text` before matching, so both accepted shapes are covered.

PR #1165, opened 2026-08-11
docs: explain close-flatten operations

JuaniRios:
## What

- Track [RAI-1690](https://linear.app/makeitrain/issue/RAI-1690).
- Document the required close-flatten config and ordered price source chain.
- Document cross widening, failure classification, dead letters, alerts, claim recovery, and outcome metrics.

## Why

- Operators need the runtime behavior and failure signals in one place.
- The old docs still described the unentitled SIP path and 300-second repricing.

## How

- Update `README.md` with config and price source behavior.
- Update `docs/conductor.md` with the retry, dead-letter, recovery, and alert flow.
- Document the 60-second reprice cycle and its interaction with the position scan.

## Testing

No new tests. This PR only changes `README.md` and `docs/conductor.md`.

## Screenshots

N/A

## Anything else

- This is the final PR in the stack.
- Runtime changes are in [#1180](https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1180).

## Summary by CodeRabbit

* **New Features**
  * Added extended-hours hedge pricing with validated quotes, broker marks, and delayed market-data fallback.
  * Added progressive cross-limit widening as the close window approaches.
  * Added automatic repricing every 60 seconds using fresh pricing references.

* **Reliability**
  * Added bounded retries, dead-letter handling, claim recovery, and deduplicated notifications for failed hedge attempts.
  * Added safe handling for unavailable symbols and permanent versus temporary quote failures.

* **Documentation**
  * Documented close-flatten settings, pricing fallbacks, retry behavior, and cross-limit requirements.

PR #1179, opened 2026-08-12
docs: specify close-flatten pricing and isolation

JuaniRios:
## What

- Track [RAI-1690](https://linear.app/makeitrain/issue/RAI-1690).
- Specify the ordered price source chain for extended-hours hedges.
- Specify the widening close-flatten cross and per-symbol failure isolation.
- Record the decision and rejected alternatives in ADR 0019.

## Why

- The hardcoded SIP quote returned an entitlement 403 during close flattening.
- One symbol could exhaust the shared retry budget and stop every hedge worker.
- The latest-trade source also froze after the regular session.

## How

- Prefer an optional primary quote, then the broker position mark, then `delayed_sip`.
- Ramp the cross from `counter_trade_slippage_bps` to `close_flatten_cross_max_bps`.
- Bound transient retries, then dead-letter and alert by symbol.
- Keep process-scoped recovery failures on the shared retry path.

## Testing

No new tests. This PR only changes `SPEC.md` and ADR 0019.

## Screenshots

N/A

## Anything else

- This is the base PR in the stack.
- Runtime changes follow in [#1180](https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1180).

---
<a href="https://app.blacksmith.sh/ST0x-Technology/codesmith/st0x.liquidity/pr/1179"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789133950&installation_model_id=19370&pr_number=1179&repository=ST0x-Technology%2Fst0x.liquidity&return_to=https%3A%2F%2Fgithub.com%2FST0x-Technology%2Fst0x.liquidity%2Fpull%2F1179&signature=cc0d56264f46265dd53f064e2da9bd878c797925d948fa5c25819b68dcd89910"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Extended-hours hedge pricing now supports optional primary quotes, position marks, and delayed market-data fallbacks.
  * Close-flatten protection progressively widens within the configured flatten window to improve execution reliability.
* **Reliability Improvements**
  * Pricing failures are categorized for faster handling, with automatic retries for temporary issues.
  * Permanent failures are isolated for operator review.
* **Monitoring**
  * Dead-letter events and skipped pricing scans now generate metrics and deduplicated alerts.
  * Repricing uses the same widening behavior and a 60-second timeout.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -395,10 +395,15 @@ the broker calendar:
 - **Regular** — standard hours; counter-trades place market orders.
 - **Extended** — pre-market or after-hours; the broker accepts only limit orders
   flagged `extended_hours = true`. Outside close-flatten mode, counter-trades
-  place a limit order priced from the latest trade price plus a configured
-  slippage buffer (`counter_trade_slippage_bps`). Rounding favors fill
-  probability (buys round up, sells round down) and honors the broker's minimum
-  price variance (penny increments at or above $1.00, sub-penny below).
+  use the same optional-primary-quote, mark, delayed-quote chain as close
+  flattening, plus a configured slippage buffer (`counter_trade_slippage_bps`).
+  With no primary provider wired, the mark replaces the market-data latest trade
+  because the latter resolves to the plan's default feed, which stops updating
+  between roughly 16:00 and 16:40 ET; pricing off it meant every hedge from then
+  until the 20:00 ET close used a frozen reference and silently failed to fill
+  whenever the market moved past the buffer. Rounding favors fill probability
+  (buys round up, sells round down) and honors the broker's minimum price
+  variance (penny increments at or above $1.00, sub-penny below).
 - **Closed** — outside all sessions (overnight, weekends, holidays); no order is
   placed. The exposure is left for the next scan to hedge once the venue
   reopens, rather than failing a durable job against a multi-hour closure.
@@ -447,20 +452,136 @@ placement:
   promptly. Partial fills are reconciled before cancellation; after the broker
   confirms the cancellation, every remaining broker-executable amount is
   released back to the position for retry.
-- Every hedge placed while the mode is active crosses the current NBBO and
-  applies the configured protection bound: buys use the current ask plus
-  `counter_trade_slippage_bps`, while sells use the current bid minus that
-  buffer. Buy prices round up and sell prices round down to Alpaca's allowed
-  tick. Quote requests explicitly select Alpaca's SIP feed so a
-  subscription-dependent default cannot silently substitute an IEX-only quote.
-  Missing SIP access and missing, non-positive, or crossed quotes are retryable
-  errors; the bot never falls back to a latest-trade or stale position price.
-- The existing `extended_hours_reprice_timeout_secs` remains active. With the
-  configured 300-second timeout and 900-second flatten window, an unfilled
-  aggressive hedge is cancelled, reconciled, and repriced from a fresh quote
-  approximately three times before close.
+- Every extended-hours hedge resolves its reference through one ordered chain:
+  an optional current bid/ask market-data source, then the **position mark**,
+  then the emergency delayed quote described below. Buys use the ask when the
+  primary quote is available; sells use the bid. The current deployment has no
+  entitled primary market-data source, so the optional source reports
+  unavailable and the effective reference remains the broker's `current_price`.
+  The source boundary is explicit so a future provider such as Alpaca SIP can be
+  added without changing the fallback contract or hedge orchestration. The mark
+  remains a fallback even after a primary source is added: a missing or failed
+  market-data lookup never prevents an otherwise priceable hedge. The selected
+  reference is crossed by a widening protection bound. Buys add the bound and
+  sells subtract it. Buy prices round up and sell prices round down to Alpaca's
+  allowed tick. See
+  [ADR 0019](adrs/0019-mark-priced-close-flatten-with-widening-cross.md).
+- The cross widens with elapsed time inside the window rather than applying one
+  flat band. It ramps linearly from `counter_trade_slippage_bps` at the window's
+  start to `close_flatten_cross_max_bps` at the extended-session close. The ramp
+  is anchored to the window, not to a per-order attempt count, so the cross is a
+  pure function of the window and the current time: restart-safe, identical
+  across retries, and independent of how many reprice cycles actually land. A
+  hedge that first becomes ready mid-window opens partway up the ramp, because
+  it has less time left to flatten. This converges toward a fill without needing
+  a real-time spread, which the bot cannot observe. An attempt that does not
+  fill is corrected only by a later, wider one, so the ceiling bounds how far
+  the bot will chase and a flatten can still end the session unfilled.
+- The `[broker]` close-flatten settings are required whenever extended-hours
+  broker windows are active: `extended_hours_close_flatten_window_secs`,
+  `close_flatten_cross_max_bps`, and `close_flatten_reprice_timeout_secs`. There
+  are no implicit defaults. Startup fails if any is missing, if the window or
+  timeout is zero or outside its supported range, or if
+  `close_flatten_cross_max_bps` is below `counter_trade_slippage_bps` or above
+  the global counter-trade slippage ceiling. The ordering constraint prevents a
+  ramp that narrows toward close.
+- When neither the optional primary source nor the mark supplies a reference,
+  the bot falls back to a quote on the `delayed_sip` feed. The emergency feed is
+  not configurable. It is the only value currently available that both answers
+  and returns a real consolidated book: `sip` is rejected without the
+  entitlement, and `iex` returns single-venue stub quotes once it stops trading.
+  Its quote is a genuine NBBO fifteen minutes old, and a wide but genuine quote
+  is crossed rather than refused, because refusing it would mean not flattening.
+- A missing, non-positive, crossed, or rejected result fails only its current
+  reference source. The resolver continues from the optional primary quote to
+  the position mark, then to the hardcoded `delayed_sip` quote. Only the error
+  selected after all three sources fail is scoped to the one symbol and raised
+  before the position is claimed. A permanent failure (any non-429 4xx) then
+  dead-letters the hedge attempt rather than retrying it in place: retrying a
+  rejection that an immediate retry cannot clear only consumes the shared retry
+  budget and can stop the process for every other symbol. A 429 follows the
+  bounded job reschedule policy under
+  [Broker rate-limit (429) backpressure](#broker-rate-limit-429-backpressure)
+  and does not consume the transient budget below. Transient failures --
+  transport errors, timeouts, and 5xx -- are re-driven on the hedge's own
+  durable budget instead: three re-drives at one, two, and four seconds, the
+  same cadence the shared worker budget would have given them. Abandoning a
+  transient failure at once would trade a full scan interval of unhedged
+  exposure for a saved second, but re-driving one forever would eventually stop
+  the process for every other symbol, so a symbol whose transient failure
+  outlives that budget dead-letters like a permanent one. A permanent cause
+  dead-letters at once and never spends the immediate transient budget, since
+  another request seconds later cannot change the answer. That dead-letter is an
+  observation and page, not suppression: the position scan deliberately attempts
+  the standing delta again on its next pass so recovery from an operator or
+  entitlement fix needs no restart. Alert deduplication prevents that periodic
+  retry from paging repeatedly for the same cause. Each dead-letter increments
+  `hedge_dead_lettered_total{symbol,reason}` and pages the operator once per
+  `(symbol, reason)`, since a dead-lettered hedge leaves a standing per-symbol
+  delta while the bot stays up; the position scan re-enqueues the hedge on its
+  next pass. The page for that exact `(symbol, reason)` is released once one of
+  the symbol's hedges reaches the broker again; other active reasons for the
+  symbol remain deduplicated independently. A feed regression that recurs a
+  later session therefore pages afresh rather than being silenced for the life
+  of the process, and a restart re-pages a failure that survived it. Initial-job
+  abandonment first reconciles any position claim that an earlier attempt of the
+  same job left outstanding. It records the dead-letter only after recovery
+  establishes that no broker order or claim remains. A pricing failure while
+  recovering an existing claimed order uses the same bounded
+  backpressure/transient budgets. Once that budget is exhausted, it dead-letters
+  the symbol but retains the claim under the pending offchain order. The
+  periodic recovery sweep owns the next recovery attempt, so the failure does
+  not stop every hedge worker. Dead-lettering is reached only after the optional
+  primary market-data source, the mark, and the hardcoded `delayed_sip` quote
+  feed have all failed, never on the first failed lookup: flattening before a
+  multi-day gap is mandatory, so a worse fill price is always preferred to no
+  fill.
+- An extended-hours buy the position scan drops before it can become a hedge job
+  -- because no reference price could be resolved for it, or because the cross
+  could not be applied -- is counted on
+  `hedge_scan_skipped_total{symbol,reason}`. The dead-letter counter cannot
+  report a job that never existed, so this is the per-symbol signal for a
+  standing delta the scan keeps skipping. A permanent or unclassified failed
+  reference-price lookup also pages the operator through the same
+  `(symbol, reason)` dedup the hedge dead-letter uses. A transient or
+  rate-limited failure remains counted but waits for the next scan rather than
+  paging on its first observation. Any page is released by the same successful
+  placement as a job-side dead-letter. Inside a close-flatten window, the same
+  skip also increments `close_flatten_blocked_total{symbol,reason}`. The former
+  is authoritative for all scan/immediate-path skips; the latter is the
+  close-flatten-only dashboard subset. They intentionally describe overlapping
+  populations and must not be summed.
+- Repricing uses two explicit cadences. The ordinary extended-hours
+  `extended_hours_reprice_timeout_secs` remains 300 seconds, preserving broker
+  time priority and avoiding repeated cancel/place calls at an unchanged flat
+  cross from 16:00 until the close-flatten window. Inside the window,
+  `close_flatten_reprice_timeout_secs` is configured to 60 seconds. An unfilled
+  aggressive hedge is repeatedly cancelled, reconciled, and repriced from a
+  fresh reference before close. Replacement occurs on a later position scan,
+  whose cadence is the separately configured `position_check_interval` (60
+  seconds in the shipped deployments), so the timeout alone does not determine
+  the number of attempts inside a window. Each later reprice samples the time
+  ramp further along, so it crosses wider than the one before it. Both timeout
+  fields are required whenever extended-hours broker windows are active; neither
+  cadence is an implicit fallback.
+- `close_flatten_outcomes_total{symbol,direction,outcome}` records terminal
+  broker-state dispatches for close-flatten placements, with `outcome` equal to
+  `filled`, `cancelled`, or `failed`. Evaluate attempt fill rate per
+  close-flatten window as `filled / (filled + cancelled + failed)`, aggregating
+  symbols and directions only when the operational question permits it. A
+  partial fill that later completes is `filled`; a partially filled order that
+  terminates cancelled or failed remains that terminal outcome and is not
+  share-weighted. Unfilled terminal outcomes therefore remain in the
+  denominator. Compare this rate with
+  `close_flatten_placements_total{symbol,direction,cross_bucket}` over the same
+  windows: concentration in the ceiling bucket together with a low fill rate is
+  the signal to evaluate raising `close_flatten_cross_max_bps`, while fills at
+  lower buckets argue against widening it. These are attempt-level operational
+  counters, not a unique-order ledger; delayed follow-up commits can cause a
+  terminal state to be observed more than once, so dashboards must use the same
+  observation semantics for numerator and denominator.
 - New onchain fills received while close-flatten mode is active use the same
-  quote-crossing placement path automatically.
+  mark-crossing placement path automatically.
 - If the extended session closes before a queued attempt runs, the session
   re-check prevents submitting an invalid extended-hours order.
 
--- a/adrs/0019-mark-priced-close-flatten-with-widening-cross.md
+++ b/adrs/0019-mark-priced-close-flatten-with-widening-cross.md
@@ -1,0 +1,237 @@
+# ADR 0019: Resolve close-flatten prices through a fallback chain
+
+- Status: Accepted
+- Date: 2026-08-10
+- Updated: 2026-08-11 after Alpaca recommended current bid/ask market data as
+  the primary limit-order reference and the position mark as fallback
+
+## Context
+
+Close-flatten mode (ADR-less, introduced by PR #1045 for RAI-1404) prices its
+hedges by crossing the SIP quote: buys take the ask plus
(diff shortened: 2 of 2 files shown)


JuaniRios:
* **#1165** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1165?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1180** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1180?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1179** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1179?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1179?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1179).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

JuaniRios:
@coderabbitai full review

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
```diff
@@ -0,0 +1,207 @@
...
+- We lose the marketability guarantee that crossing a live NBBO provided. A mark
+  plus a band is an estimate of where the book is, not a price derived from it.
+  The widening cross recovers this probabilistically rather than absolutely: an
+  attempt can fail to fill, and only a later, wider one corrects it.
+- A flatten can still end the session unfilled if the ramp's ceiling is below
+  the true spread. This is a real residual risk and is the reason the ceiling is
+  configurable rather than fixed.
+- Fill rates inside the close-flatten window become the metric that tells us
+  whether the ramp parameters are right. Without them this decision cannot be
+  evaluated, so they are required, not optional.
```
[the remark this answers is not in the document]
JuaniRios: Good catch. I’m not defining this contract in the spec yet because the current counters cannot produce it honestly. Outcomes do not retain the cross bucket, partial fills are not distinct, and terminal observations need a stable event identity. Adding that would require runtime code outside this stack.

SPEC.md
```diff
@@ -447,20 +452,87 @@ placement:
...
+  flat band. It ramps linearly from `counter_trade_slippage_bps` at the window's
+  start to `close_flatten_cross_max_bps` at the extended-session close. The ramp
+  is anchored to the window, not to a per-order attempt count, so the cross is a
+  pure function of the window and the current time: restart-safe, identical
+  across retries, and independent of how many reprice cycles actually land. A
+  hedge that first becomes ready mid-window opens partway up the ramp, because
+  it has less time left to flatten. This converges toward a fill without needing
+  a real-time spread, which the bot cannot observe. An attempt that does not
+  fill is corrected only by a later, wider one, so the ceiling bounds how far
+  the bot will chase and a flatten can still end the session unfilled.
```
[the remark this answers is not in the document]
JuaniRios: Already covered in #1180. `CloseFlattenCrossRamp` uses the current window time, validates the ceiling, and is applied in scan-time and placement pricing. The tests cover the start, midpoint, end, and both clamped boundaries.

SPEC.md
```diff
@@ -447,20 +452,87 @@ placement:
...
+  would eventually stop the process for every other symbol, so a symbol whose
+  transient failure outlives that budget dead-letters like a permanent one. A
+  permanent cause dead-letters at once and never spends the budget, since
+  re-asking cannot change the answer. Each dead-letter increments
+  `hedge_dead_lettered_total{symbol,reason}` and pages the operator once per
+  `(symbol, reason)`, since a dead-lettered hedge leaves a standing per-symbol
+  delta while the bot stays up; the position scan re-enqueues the hedge on its
+  next pass. The page is released once one of that symbol's hedges reaches the
+  broker again, so a feed regression that recurs a later session pages afresh
+  rather than being silenced for the life of the process, and a restart re-pages
```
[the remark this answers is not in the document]
JuaniRios: The symbol-wide release is intentional. `(symbol, reason)` deduplicates diagnostic pages, but each page represents the same standing symbol delta. Once one hedge reaches the broker, that delta is being resolved, so every cause slot for the symbol is released for a later recurrence.

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
```diff
@@ -0,0 +1,207 @@
...
+- The mark is only available for symbols where a broker position exists. The
+  `delayed_sip` fallback covers the gap, at the cost of a 15-minute-old price.
```
agryaznov, line 181: *_high_*: `delayed_sip` is 15 minutes delayed (probe table, line 41) and the close-flatten window is 900 seconds — exactly 15 minutes (`extended_hours_close_flatten_window_secs = 900`, `config/prod/st0x-hedge.toml`). So a `delayed_sip` quote taken at the window's start reflects 19:30 ET, and one taken at the close reflects 19:45 ET: inside the flatten window the fallback shows only pre-window prices, at every point in the window. The ADR treats the delay as a cost ("at the cost of a 15-minute-old price") without noting that the delay equals the window, so on this path the bot crosses a band around a price from before the flatten began — during precisely the period when late-session price discovery is most likely to move it. This is the path taken whenever the mark is missing, which decision 5 makes materially more common (see [line 181](https://github.com/ST0x-Technology/st0x.liquidity/blob/8ba6934c/adrs/0019-mark-priced-close-flatten-with-widening-cross.md#L181)).

**Fix:** state the arithmetic explicitly and say what it implies for the ramp on the `delayed_sip` path — either the ramp should start wider when pricing off a quote older than the window, or the ADR should record that this residual is accepted and why.

SPEC.md
```diff
@@ -447,20 +452,87 @@ placement:
...
+  the bot falls back to a quote on the `delayed_sip` feed. The emergency feed is
+  not configurable. It is the only value currently available that both answers
+  and returns a real consolidated book: `sip` is rejected without the
+  entitlement, and `iex` returns single-venue stub quotes once it stops trading.
+  Its quote is a genuine NBBO fifteen minutes old, and a wide but genuine quote
+  is crossed rather than refused, because refusing it would mean not flattening.
+- Missing, non-positive, or crossed quotes, and quote requests the feed rejects
+  permanently (any non-429 4xx), are scoped to the one symbol and are raised
+  before the position is claimed, so they dead-letter the hedge attempt rather
+  than retrying it in place: retrying a rejection an immediate retry cannot
```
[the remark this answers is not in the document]
JuaniRios: Fixed in 8ba6934c. The sentence now reads “retrying a rejection that an immediate retry cannot clear.”

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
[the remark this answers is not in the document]
JuaniRios: Good catch. I tracked the runtime telemetry work in [RAI-1727](https://linear.app/makeitrain/issue/RAI-1727/define-accurate-close-flatten-fill-rate-telemetry). It needs cross-bucket attribution, distinct partial-fill outcomes, and a stable terminal identity, so it stays outside this stack.

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
```diff
@@ -0,0 +1,207 @@
...
+   the hardcoded `delayed_sip` quote last. A buy takes the primary quote's ask
+   and a sell takes its bid. Alpaca recommends market data as the primary basis
+   for limit orders and the mark as fallback. The primary capability is
+   deliberately optional because this account does not currently have a usable
+   real-time market-data entitlement and may never use Alpaca SIP. With no
+   provider installed, the chain starts effectively at the mark, preserving the
+   production behaviour this ADR introduced. A missing or failed primary lookup
+   always falls through to the mark.
+
+2. **The cross widens with elapsed time inside the window** instead of applying
```
agryaznov, line 108: *_medium_*: Decision 2 makes the cross "a pure function of the window and the current time", which requires knowing the extended-session close. Decision 3 establishes that a missing price never skips the flatten — but the ADR never says what happens when the session lookup fails, and the ramp cannot be computed without it. That path exists in the implementation and currently drops the attempt.


**Fix:** extend decision 3's principle to the session boundary, or state the fallback cross used when the window cannot be resolved.

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
```diff
@@ -0,0 +1,207 @@
...
+   a single flat band. It ramps linearly from `counter_trade_slippage_bps` at
```
agryaznov, line 109: *_low_*: The ramp starts at `counter_trade_slippage_bps`, which per `README.md` also "controls the buy-side preflight buffer". Retuning the ramp's floor therefore silently changes buy-side preflight sizing, and vice versa. That coupling may be intentional, but it is not acknowledged.

**Fix:** note the shared knob and why reusing it is preferable to a dedicated ramp floor.

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
```diff
@@ -0,0 +1,207 @@
...
+   attempt sample the ramp further along, so the cross converges toward a fill
+   without needing a spread we cannot observe in real time.
+
+3. **A missing price never skips the flatten.** Dead-lettering remains the
+   terminal behaviour, but only once every reference source has failed, not on
+   the first failed lookup. The order is optional primary quote, position mark,
+   `delayed_sip` quote, then dead-letter. Failure of a preferred source never
+   suppresses a usable fallback.
+
+4. **There is no market-data feed configuration.** The
```
agryaznov: *_medium_*: Decision 4 says the `[broker] market_data_feed` key "is removed, along with the `AlpacaMarketDataFeed` enum it selected". Neither exists on `master`; the code hardcodes `feed=sip` at `crates/execution/src/alpaca_market_data.rs:200`. The ADR describes an unmerged branch as the current state, which will mislead anyone reading it as the record of what changed.

**Fix:** restate as what the decision establishes — the feed stays hardcoded and no feed configuration is exposed — rather than as the removal of something that was never on `master`.

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
```diff
@@ -0,0 +1,207 @@
...
+   quotes once it stops trading around 16:00 ET, and `delayed_sip` is the only
+   value that both answers and returns a real book. A required key with exactly
+   one correct value is not a choice, it is a way to misconfigure the bot into
+   losing money, which is how `iex` reached the production config in the first
+   place. The executor instead exposes an optional primary-quote capability
+   whose default is unavailable. A future provider is selected by wiring an
+   implementation, not by exposing a setting that is invalid in the current
+   deployment.
+
+5. **Ordinary extended-hours hedges use the same source chain**, replacing
```
agryaznov, line 152: *_medium_*: Decision 5 replaces `fetch_latest_trade_price` for hedge pricing, but two other consumers on `master` keep it: the buy cash preflight at `crates/execution/src/alpaca_broker_api/executor.rs:264` (inside the `Direction::Buy` arm, feeding `preflight_buy_cash`), and the `Executor` trait method of the same name at `executor.rs:288-292`, which exposes the source as public API. Both keep the frozen-at-16:00-ET reference in every session. The decision reads as though the stale source is retired outright, and the Context's "same data gap affects 16:00 to 20:00 ET every weekday" argument applies to both.

**Fix:** scope the decision to the hedge-pricing path and state explicitly whether the preflight caller is in scope, with the reason either way.

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
```diff
@@ -0,0 +1,207 @@
...
+## Consequences
+
+- We lose the marketability guarantee that crossing a live NBBO provided. A mark
+  plus a band is an estimate of where the book is, not a price derived from it.
+  The widening cross recovers this probabilistically rather than absolutely: an
+  attempt can fail to fill, and only a later, wider one corrects it.
+- A flatten can still end the session unfilled if the ramp's ceiling is below
+  the true spread. This is a real residual risk and is the reason the ceiling is
+  configurable rather than fixed.
+- Fill rates inside the close-flatten window become the metric that tells us
```
agryaznov: *_medium_*: The Consequences section states "Fill rates inside the close-flatten window become the metric that tells us whether the ramp parameters are right. Without them this decision cannot be evaluated, so they are required, not optional." No fill-rate metric is defined anywhere — `SPEC.md` introduces `hedge_dead_lettered_total{symbol,reason}` and `hedge_scan_skipped_total{symbol,reason}`, both failure counters, and "fill rate" appears nowhere outside this ADR line. The document states its own evaluation criterion and then does not create it, so the ramp ceiling has no feedback signal to tune against.

**Fix:** either specify the fill-rate metric (name, labels, where it is recorded) in `SPEC.md` as part of this change, or move it to a Follow-ups section with an owner and stop calling it required.

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
```diff
@@ -0,0 +1,207 @@
...
+immediate dependency because it is not self-serve for a Broker API partner,
+pricing is bespoke and currently unknown, and the exposure recurs every Friday
+while procurement runs. The optional primary-quote boundary preserves this path
+without making the entitlement a prerequisite for the present fix.
+
+**Configure `feed=iex`.** What the branch under review currently does. Rejected:
+the table above shows it prices 5% to 16% through the real market inside the
+exact window it is needed, which is worse than the crash it replaces.
+
+**Keep the feed configurable, restricted to `Sip | DelayedSip`.** Keeps the
```
agryaznov, line 208: *_medium_*: The alternative "Keep the feed configurable, restricted to `Sip | DelayedSip`" is rejected partly because the `delayed_sip` path is "reached only when a symbol has no broker position" — rare on the close-flatten path, since flattening implies a position. But decision 5 extends the same chain to ordinary extended-hours hedges, which routinely open exposure in symbols where no broker position exists yet. On that path the mark is absent by default and `delayed_sip` becomes the ordinary case, not the edge case. The consequence "The mark is only available for symbols where a broker position exists" is stated but never reconciled with decision 5, so the alternatives section reasons about a pre-decision-5 world.

**Fix:** reconcile the two — either qualify the rejection rationale (the `delayed_sip` path is common for ordinary hedges, and the reasons to remove the key stand anyway), or state how often decision 5 is expected to land on `delayed_sip`.

SPEC.md
```diff
@@ -447,20 +452,96 @@ placement:
...
+  added without changing the fallback contract or hedge orchestration. The mark
+  remains a fallback even after a primary source is added: a missing or failed
+  market-data lookup never prevents an otherwise priceable hedge. The selected
+  reference is crossed by a widening protection bound. Buys add the bound and
+  sells subtract it. Buy prices round up and sell prices round down to Alpaca's
+  allowed tick. See
+  [ADR 0019](adrs/0019-mark-priced-close-flatten-with-widening-cross.md).
+- The cross widens with elapsed time inside the window rather than applying one
+  flat band. It ramps linearly from `counter_trade_slippage_bps` at the window's
+  start to `close_flatten_cross_max_bps` at the extended-session close. The ramp
```
agryaznov, line 471: *_medium_*: The key appears once as the ramp's upper bound, with no config block, no statement that it is required with no default, and no ordering constraint. Startup rejects `configured < ramp_base_bps` or `> MAX_COUNTER_TRADE_SLIPPAGE_BPS` (`crates/config/src/loader.rs:513-526`, `CloseFlattenCrossMaxBpsOutOfRange`) — a value below `counter_trade_slippage_bps` inverts the ramp so the cross narrows toward the close, and fails the deploy.

**Fix:** document it with the other `[broker]` close-flatten keys, state that startup fails when it is missing, and state the ordering constraint.

SPEC.md
```diff
@@ -447,20 +452,96 @@ placement:
...
+  transient failure at once would trade a full scan interval of unhedged
+  exposure for a saved second, but re-driving one forever would eventually stop
+  the process for every other symbol, so a symbol whose transient failure
+  outlives that budget dead-letters like a permanent one. A permanent cause
+  dead-letters at once and never spends the budget, since re-asking cannot
+  change the answer. Each dead-letter increments
+  `hedge_dead_lettered_total{symbol,reason}` and pages the operator once per
+  `(symbol, reason)`, since a dead-lettered hedge leaves a standing per-symbol
+  delta while the bot stays up; the position scan re-enqueues the hedge on its
+  next pass. The page is released once one of that symbol's hedges reaches the
```
agryaznov: *_medium_*: The text says a permanent failure dead-letters at once because "re-asking cannot change the answer", then says "the position scan re-enqueues the hedge on its next pass". So a permanent 403 is re-asked roughly every 60 seconds with no backoff; the dead-letter is a page, not a stop.

**Fix:** say the dead-letter records and pages while the scan keeps retrying, or describe the suppression that actually stops the re-ask.

SPEC.md
```diff
@@ -447,20 +452,96 @@ placement:
...
+  offchain order. The periodic recovery sweep owns the next recovery attempt, so
+  the failure does not stop every hedge worker. Dead-lettering is reached only
+  after the optional primary market-data source, the mark, and the hardcoded
+  `delayed_sip` quote feed have all failed, never on the first failed lookup:
+  flattening before a multi-day gap is mandatory, so a worse fill price is
+  always preferred to no fill.
+- An extended-hours buy the position scan drops before it can become a hedge job
+  -- because no reference price could be resolved for it, or because the cross
+  could not be applied -- is counted on
+  `hedge_scan_skipped_total{symbol,reason}`. The dead-letter counter cannot
```
agryaznov, line 542: *_medium_*: The spec introduces `hedge_scan_skipped_total{symbol,reason}` for scan-side skips, but `record_close_flatten_block` already counts the same event inside the window (`src/trading/offchain/hedge.rs:252,263,377`). Two counters for one condition means neither is the number to alert on.

**Fix:** fold the new reasons into the existing counter, or state which is authoritative and why both exist.

SPEC.md
```diff
@@ -447,20 +452,96 @@ placement:
...
+  reference-price lookup also pages the operator through the same
+  `(symbol, reason)` dedup the hedge dead-letter uses. A transient or
+  rate-limited failure remains counted but waits for the next scan rather than
+  paging on its first observation. Any page is released by the same successful
+  placement as a job-side dead-letter.
 - The existing `extended_hours_reprice_timeout_secs` remains active. With the
-  configured 300-second timeout and 900-second flatten window, an unfilled
-  aggressive hedge is cancelled, reconciled, and repriced from a fresh quote
-  approximately three times before close.
+  configured 60-second timeout and 900-second flatten window, an unfilled
```
agryaznov: *_low_*: The spec asserts "the configured 60-second timeout and 900-second flatten window" and derives "roughly seven or eight attempts". All four shipped configs set `extended_hours_reprice_timeout_secs = 300` at this head, giving two or three attempts. Stacked PR #1180 sets them to 60, so this resolves on merge — same root cause and same reasoning as [the ADR at line 116](https://github.com/ST0x-Technology/st0x.liquidity/blob/8ba6934c/adrs/0019-mark-priced-close-flatten-with-widening-cross.md#L116), including the `AGENTS.md:101` independence rule.

**Fix:** state the timeout as the cadence the design requires and note it takes effect with the implementation PR, rather than as the currently configured value.

agryaznov (approved):
not a blocker, but to good to clarify some points

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
JuaniRios: Good catch. Fixed in 81b19e01. The ADR now states that the 15-minute delay equals the 900-second window, so every delayed quote is pre-window. It also records why we accept the same bounded ramp instead of starting wider.

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
JuaniRios: Fixed in 81b19e01. The ADR now separates price-source fallback from calendar failure. A queued hedge treats the latter as process-scoped, while scan and immediate paths count the skip and retry on a later scan.

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
JuaniRios: Fixed in 81b19e01. The ADR now calls out the shared counter_trade_slippage_bps knob and explains why the first close-flatten attempt should match ordinary extended-hours policy.

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
JuaniRios: Fixed in 81b19e01. Decision 4 now says the feed stays hardcoded and no feed config or enum is introduced.

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
JuaniRios: Fixed in 81b19e01. Decision 5 now scopes the replacement to extended-hours hedge pricing and its exact-limit preflight. Regular-hours market-order preflight keeps the latest-trade lookup because IEX is live then.

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
JuaniRios: Fixed in 81b19e01. SPEC now defines close_flatten_outcomes_total, its labels, denominator, partial-fill handling, and how to compare it with close_flatten_placements_total when tuning close_flatten_cross_max_bps.

adrs/0019-mark-priced-close-flatten-with-widening-cross.md
JuaniRios: Fixed in 81b19e01. The alternatives section now says delayed_sip is uncommon for close-flatten sells but expected to be common for ordinary extended-hours buys without a broker position.

SPEC.md
JuaniRios: Fixed in 81b19e01. SPEC now documents all required close-flatten keys, no-default behavior, supported ranges, and the cross ceiling ordering constraint.

SPEC.md
JuaniRios: Fixed in 81b19e01. SPEC now says dead-lettering records and pages but does not suppress later scans. The scan retries the standing delta, and alert dedup prevents repeated pages for the same cause.

SPEC.md
JuaniRios: Fixed in 81b19e01. hedge_scan_skipped_total is now documented as authoritative for scan and immediate-path skips. close_flatten_blocked_total is the close-flatten-only subset, and the two must not be summed.

SPEC.md
JuaniRios: Fixed in 81b19e01. SPEC now separates the 300-second ordinary cadence from the 60-second close-flatten cadence and states that position_check_interval also controls replacement timing.

PR #1180, opened 2026-08-12
fix: harden close-flatten pricing and hedge isolation

JuaniRios:
## What

- Track [RAI-1690](https://linear.app/makeitrain/issue/RAI-1690).
- Resolve extended-hours limits through a primary quote, position mark, and delayed SIP fallback.
- Widen close-flatten limits over the configured window.
- Isolate symbol price failures with bounded retries, dead letters, metrics, and deduplicated alerts.
- Persist close-flatten attribution and record terminal order outcomes.

## Why

- The previous SIP lookup returned a permanent 403 on the production entitlement.
- The previous latest-trade source stopped updating after the regular session.
- One failed symbol could stop hedging for every symbol.

## How

- Add optional `fetch_primary_limit_quote` and `fetch_position_mark` executor capabilities.
- Prefer the primary ask for buys and bid for sells, then fall through to the mark.
- Use hardcoded `delayed_sip` only after the primary source and mark fail.
- Recheck the exact extended-hours limit before submission.
- Classify failures by permanence, backpressure, and claim scope.
- Recover live claims before dead-lettering and reserve alerts before notification.

## Testing

- Added tests for the quote chain, mark validation, symbol matching, and error permanence.
- Added tests for cross widening, fresh preflight checks, retries, dead letters, alerts, claim recovery, and terminal outcomes.
- Workspace check, Clippy, formatting, Nix formatting, and changed-file hooks passed locally.

## Screenshots

N/A

## Anything else

- Alpaca config now requires `close_flatten_cross_max_bps`.
- Production, staging, and `s01-issuer` set the ceiling to 400 basis points.
- No primary quote provider is installed, so the current path starts at the position mark.
- This PR depends on [#1179](https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1179).

---
<a href="https://app.blacksmith.sh/ST0x-Technology/codesmith/st0x.liquidity/pr/1180"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789133952&installation_model_id=19370&pr_number=1180&repository=ST0x-Technology%2Fst0x.liquidity&return_to=https%3A%2F%2Fgithub.com%2FST0x-Technology%2Fst0x.liquidity%2Fpull%2F1180&signature=c7c87275796da0278f52b4b7c6050dde67d35208caca60686a297f4d869b0ef5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added configurable close-flatten price-crossing limits, with ramping up to 400 basis points during the flatten window.
  - Extended-hours hedge pricing now uses primary limit quotes and position marks for more accurate execution decisions.
  - Added improved tracking of close-flatten outcomes and hedge placement results.

- **Improvements**
  - Reduced extended-hours repricing timeout from 300 to 60 seconds.
  - Improved handling of temporary, permanent, invalid, and mismatched market-data responses.
  - Added safeguards and alerts for failed hedge attempts.


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1180?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1165** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1165?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1180** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1180?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1180?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1179** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1179?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1180).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

JuaniRios:
@coderabbitai full review

JuaniRios:
@coderabbitai full review

crates/execution/src/lib.rs
```diff
@@ -316,18 +346,36 @@ pub trait Executor: Send + Sync + 'static {
...
     ) -> Result<Option<Positive<Usd>>, Self::Error> {
         Ok(None)
     }
 
-    /// Fetches the latest validated bid and ask for a symbol. Returns `None`
-    /// when the executor does not support quote lookups.
+    /// Fetches the validated emergency quote for a symbol. Alpaca implements
+    /// this with the hardcoded `delayed_sip` feed, after both the optional
+    /// primary quote and the position mark fail. Returns `None` when the
+    /// executor does not support the fallback lookup.
```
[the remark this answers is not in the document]
JuaniRios: The complete source order belongs on `resolve_extended_hours_reference_price`, which owns and implements it. The trait methods document each source contract, and `hedge_price_source_total` records the selected source.

crates/execution/src/alpaca_broker_api/mock_api.rs
```diff
@@ -1097,17 +1134,24 @@ fn register_latest_trade_endpoint(server: &MockServer, state: &Arc<Mutex<MockSta
...
                     }
                 }),
-            )
+                MockMarketDataKind::Quote => json!({
+                    "symbol": symbol,
+                    "quote": {
+                        "bp": format_float_with_fallback(&price),
+                        "ap": format_float_with_fallback(&price),
+                    }
+                }),
```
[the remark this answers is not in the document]
JuaniRios: The configured mock price is a deterministic reference anchor, not a spread model. Perturbing it would change existing mock semantics without a configured spread. Explicit unequal bid and ask values already cover parsing and direction selection.

crates/execution/src/alpaca_broker_api/executor.rs
```diff
@@ -870,6 +869,145 @@ mod tests {
...
+            .unwrap_err();
+
+        assert!(matches!(error, AlpacaBrokerApiError::LatestQuote(_)));
+        assert_eq!(
+            error.backpressure(),
+            Some(Backpressure {
+                retry_after: Some(Duration::from_secs(15))
+            })
+        );
+    }
```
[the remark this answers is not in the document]
JuaniRios: The existing assertion already proves this request matched the fresh mock and received its `Retry-After: 15` response. There is no competing quote matcher in this test, so a separate hit-count assertion would be redundant.

src/trading/offchain/close_flatten.rs
```diff
@@ -46,13 +47,100 @@ impl CloseFlattenPolicy {
...
+        // `active_window` (which only yields a window when `now < closes_at`),
+        // but crossing at the ceiling is the safe reading of "no time left".
+        if total <= 0 {
+            return self.max_bps;
+        }
+
+        let elapsed = (now - window.started_at).num_seconds().clamp(0, total);
+        let progressed = u128::from(self.base_bps)
+            + u128::from(span) * u128::from(elapsed.unsigned_abs())
+                / u128::from(total.unsigned_abs());
```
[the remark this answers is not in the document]
JuaniRios: `total > 0` is checked immediately above, and `elapsed` is clamped into `0..=total`, so both values are proven non-negative. `unsigned_abs()` preserves that invariant without adding an impossible error branch.

JuaniRios:
Partially agree, but replacing five pre-existing CLI fixtures across four modules is outside this hedge-isolation change. Those fixtures override different inventory, Alpaca, and rebalancing state, so the refactor needs its own review.

src/trading/offchain/hedge.rs
```diff
@@ -232,94 +335,86 @@ async fn select_order_kind_for_current_session(
...
+            let limit_price = apply_slippage(reference.price.inner(), direction, cross_bps)
+                .map_err(TradeAccountingError::SlippageCalculation)?;
+
+            // The scan-time preflight approved this buy against an earlier
+            // reference -- potentially minutes stale by the time this job
+            // reaches perform(). Re-check cash sufficiency against the exact
+            // limit this order is about to submit, closing that staleness
+            // window for every extended-hours buy. Sells are unaffected by
+            // price, so only buys need this.
+            if direction == Direction::Buy
```
agryaznov, line 393: *_high_*: `extended_hours_preflight_at_submitted_price` is gated only on `direction == Direction::Buy`; it previously also required `close_flatten_active`. It therefore runs on every extended-hours buy, including the `select_order_kind_for_current_session` call made from `recover_pending_poll_status`'s `Pending` arm (line 835). On that path a prior attempt may already have a live order at the broker whose `MarkAccepted` commit was lost. That live limit buy reserves cash, so `get_account_funds().buying_power` is already reduced, the fresh preflight returns `Skipped` → `Ok(None)`, the idempotent re-drive is skipped, and no `PollOrderStatus` job is pushed. The live broker order then sits unpolled and its fill unrecorded until the next restart — precisely the failure `recover_pending_poll_status` exists to prevent.

**Fix:** skip the staleness re-check on the recovery path; the order being re-driven is deduped by `client_order_id`, so it is not a new cash commitment and does not need a preflight.

src/trading/offchain/hedge.rs
```diff
@@ -450,16 +830,24 @@ async fn recover_pending_poll_status(
...
                 direction,
                 market_session,
             )
-            .await?
+            .await
+            .map_err(|source| TradeAccountingError::ClaimedHedgeOrderKind {
+                symbol: symbol.clone(),
+                source: ClaimedHedgeOrderKindCause::classify(source),
+            })?
             else {
```
agryaznov, line 858: *_medium_*: The `else { return Ok(ClaimOutcome::NothingClaimed) }` arm is commented "the venue closed under the claim", but since [line 384](https://github.com/ST0x-Technology/st0x.liquidity/blob/092ef666/src/trading/offchain/hedge.rs#L384) `Ok(None)` also means "the exact-price preflight rejected this buy" — with a `Pending` order and its position claim still outstanding. `abandon_or_recover` (line 1495) treats that as `NothingClaimed` and calls `record_abandoned_hedge`, incrementing `hedge_dead_lettered_total` and paging. This contradicts the spec directly: `SPEC.md:510` states the dead-letter is recorded "only after recovery establishes that no broker order or claim remains". Scenario: attempt 1 leaves AAPL `Pending`; attempt 2 fails with a 403 `LimitQuoteFetch`; recovery re-prices, the buy fails preflight, and the operator is paged that the hedge was abandoned with nothing outstanding — while a claim and possibly a live order remain.

**Fix:** falls out of [line 384](https://github.com/ST0x-Technology/st0x.liquidity/blob/092ef666/src/trading/offchain/hedge.rs#L384); if the gate stays, distinguish "preflight rejected" from "venue closed" so the claim is not reported as released.

config/staging/st0x-hedge.toml
```diff
@@ -28,8 +28,9 @@ logs_endpoint = "http://st0x-observability.tail6094d7.ts.net:9428"
 # incoming orderbook fills.
 [broker]
 counter_trade_slippage_bps = 100
-extended_hours_reprice_timeout_secs = 300
+extended_hours_reprice_timeout_secs = 60
```
agryaznov: *_medium_*: `extended_hours_reprice_timeout_secs` drops 300 → 60 in prod, staging and `example.config.toml`. This is what makes `SPEC.md:537`'s "configured 60-second timeout" and ADR 0019 decision 2's ramp-cadence argument true, so in-window it is correct and necessary. But `request_extended_hours_reprice_timeout_cancellations` (`src/position_check.rs:891`) sweeps all extended-hours limit orders, not only those inside the 900-second window. Outside the window `CloseFlattenCrossRamp::cross_bps` returns the flat `base_bps`, so from 16:00 to 19:45 ET an unfilled hedge is now cancelled and re-placed at essentially the same price roughly every 60 seconds instead of every 300: ~210 cancel/replace pairs per symbol per session instead of ~42, each losing broker time priority and adding cancel + mark + place calls, for no widening of the cross. Neither `SPEC.md` nor ADR 0019 reasons about out-of-window cadence, so this side effect is undocumented and unjustified.

**Fix:** scope the shorter timeout to the close-flatten window, or document why the out-of-window churn is acceptable.

src/trading/onchain/trade_accountant.rs
```diff
@@ -561,10 +562,190 @@ pub(crate) enum TradeAccountingError {
...
+            Self::LimitQuoteFetch { .. } => SymbolScoped {
+                reason: SymbolScopedReason::LimitQuoteFetch,
+            },
+            Self::LimitQuoteUnavailable { .. } => SymbolScoped {
+                reason: SymbolScopedReason::LimitQuoteUnavailable,
+            },
+            Self::SlippageCalculation(_) => SymbolScoped {
+                reason: SymbolScopedReason::SlippageCalculation,
+            },
+            Self::CloseFlattenPreflightAtPrice { .. } => SymbolScoped {
```
agryaznov, line 729: *_medium_*: The variant is classified `ErrorScope::SymbolScoped`, making it dead-letterable per symbol. Its source is `preflight_counter_trade_at_price`, whose buy branch calls `get_account_funds` against the account-wide `/v1/trading/accounts/{id}/account` endpoint. An account-level permanent failure (403, or `AccountNotActive` — both `Permanence::Permanent`) is not per-symbol: every configured symbol dead-letters independently with `reason="close_flatten_preflight_at_price"` and fires its own operator page, instead of surfacing once as the process-scoped condition it is. Combined with [hedge.rs:384](https://github.com/ST0x-Technology/st0x.liquidity/blob/092ef666/src/trading/offchain/hedge.rs#L384) this now applies to every extended-hours buy, not only close-flatten ones, so the fan-out is wider than before this PR.

**Fix:** classify account-level permanence as process-scoped, or collapse the pages so an account-wide failure alerts once.

src/position_check.rs
```diff
@@ -97,6 +116,80 @@ pub(crate) enum CheckPositionsError {
...
+        counter!(
+            "close_flatten_blocked_total",
+            "symbol" => symbol.to_string(),
+            "reason" => reason
+        )
+        .increment(1);
+    }
+}
+
+fn should_page_reference_price_failure(error: &ReferencePriceError) -> bool {
```
agryaznov: *_low_*: `should_page_reference_price_failure` returns `true` unconditionally for `ReferencePriceError::Unavailable`. Under `BrokerCtx::DryRun` — which `crates/config/src/loader.rs:1155` and `CtxError::MissingCloseFlattenCrossMaxBps` treat as a real runtime mode with extended hours enabled — the executor is `MockExecutor`, whose `fetch_primary_limit_quote`, `fetch_position_mark` and `fetch_latest_quote` all return `Ok(None)` without overrides. So the first extended-hours scan pages once per symbol, and every extended-hours hedge job dead-letters with `limit_quote_unavailable`. "No source is wired" is a configuration state, not an incident.

**Fix:** suppress paging when no reference source is configured, or exclude `DryRun` from the page path.

src/conductor.rs
```diff
@@ -3405,16 +3409,18 @@ where
...
-    // which has the OrderPlacer and submits the limit order. CheckPositions
-    // remains the backstop. Preflight + clamp here, mirroring the scan path, so
-    // we never enqueue a hedge that would exceed buying power.
+    // Extended-hours counter-trades cannot be placed inline. Instead of waiting
+    // for the next CheckPositions scan (~1 min), enqueue an immediate PlaceHedge
+    // job. CheckPositions remains the backstop. Preflight buys from the same
+    // crossed reference the job will price from; the ordinary latest-trade
+    // preflight can understate a ramped limit's buying-power requirement.
     if execution.market_session == MarketSession::Extended {
+        let Some(preflight) = resolve_extended_hours_preflight(&execution, cqrs).await else {
```
agryaznov, line 3419: *_low_*: When `resolve_extended_hours_preflight` returns `None` (session lookup failed, no reference price, or the cross overflowed), the inline path does `return Ok(None)` after only a `warn!`, emitting no counter. The two sibling paths both instrument this exact condition — the scan with `hedge_scan_skipped_total`, the job with `hedge_dead_lettered_total` — and `SPEC.md` justifies the scan counter precisely because "the dead-letter counter cannot report a job that never existed". That reasoning applies identically here: a transient 503 from the position endpoint during extended hours drops the immediate `PlaceHedge` enqueue for that fill with no metric at all, leaving only the ~1-minute `CheckPositions` backstop and a log line.

**Fix:** emit `hedge_scan_skipped_total` (or an equivalent) on this path so all three share one instrumented contract.

agryaznov (changes requested):

src/trading/offchain/hedge.rs
JuaniRios: Good catch. Fixed in b398a194. Pending recovery now skips the exact-price buying-power preflight. The broker re-drive remains idempotent through client_order_id and always re-arms polling.

src/trading/offchain/hedge.rs
JuaniRios: Fixed in b398a194. Recovery can no longer return NothingClaimed because already-reserved buying power rejected the re-drive, so a live Pending order is not falsely dead-lettered or paged.

config/staging/st0x-hedge.toml
JuaniRios: Fixed in b398a194. Ordinary extended-hours orders keep the 300-second cadence. Close-flatten orders use a separate required 60-second timeout selected from the persisted close_flatten flag.

src/trading/onchain/trade_accountant.rs
JuaniRios: Fixed in b398a194. Permanent account-wide preflight failures are process-scoped. Transient and unclassified pricing failures remain symbol-scoped.

src/position_check.rs
JuaniRios: Fixed in b398a194. DryRun treats an unavailable reference source as expected and suppresses paging while keeping the skip metric.

src/conductor.rs
JuaniRios: Fixed in b398a194. Immediate extended-hours exits now increment hedge_scan_skipped_total for session lookup, reference source, and slippage failures. Close-flatten paths also keep close_flatten_blocked_total.

src/conductor.rs
```diff
@@ -3405,16 +3410,18 @@ where
...
+            return Ok(None);
```
agryaznov, line 3420: [low] This path drops the immediate `PlaceHedge` enqueue with only a `warn!` -- no counter. Its two siblings both instrument exactly this condition (`hedge_scan_skipped_total` on the scan, `hedge_dead_lettered_total` on the job), and the spec's rationale for the scan counter ("the dead-letter counter cannot report a job that never existed") applies identically here: a transient 503 during extended hours loses the fill's immediate hedge with no metric, leaving only the ~1-min CheckPositions backstop. Emit `hedge_scan_skipped_total` (or an equivalent) here so all three paths share one instrumented contract.

agryaznov (approved):
just one last small nit left, otherwise LGTM

PR #1184, opened 2026-08-12
fix: give partial-fill conversions a legal terminal transition

agryaznov:
# What

Gives every conversion outcome the bounded poll can produce a defined, legal transition in the rebalancing aggregate, so none of them strands funds, leaks the in-flight guard, or terminalizes a rebalance against an order that may still fill.

Closes RAI-1718. Stacked on #1174, which is what makes these outcomes reachable.

# Why

Bounding the conversion poll (#1174) is correct at the execution layer, but accepting a partial fill lets two states reach the aggregate that it had no legal transition for:

A conversion settling **below Alpaca's $51 withdrawal minimum**. The floor is enforced trigger-side on the *requested* amount, so a partial fill under it produced a withdrawal Alpaca rejects — terminalizing the transfer on a withdrawal error and misattributing the cause.

A conversion whose **cancel never settles** (`ConversionCancelNotSettled` / `ConversionOrderNotFound`). The order may still be live and still fill, so recording a failure releases the in-flight guard while real money can move, and lets the trigger arm a second conversion for the same imbalance.

# How

**The sub-minimum case is refused at two points, because only one of them can record a failure.** `execute_usd_to_usdc_conversion` checks before sending `ConfirmConversion` — the only moment `FailConversion` is legal, since `transition_fail_conversion` rejects it from `ConversionComplete` with `ConversionAlreadyCompleted`. `initiate_alpaca_withdrawal` checks again as defence for the resume path, where an aggregate persisted before this change can still arrive short and no failure transition remains; there it refuses the call without emitting a command and leaves the aggregate for reconciliation.

**Unresolved outcomes latch instead of retrying.** Both legs map the two broker errors to `ConversionOutcomeUnresolved` without emitting `FailConversion`, and both apalis jobs handle it by logging, notifying, and returning `Ok(())` — modelled on the existing `BurnSubmitInconclusive` arm. The `Ok` is the mechanism: it stops apalis re-entering the resume, which would otherwise reach the `Converting` arm and emit `FailConversion`, releasing the guard against a live order.

**Both legs now decide identically for identical broker state.** The post-deposit leg previously sent `FailConversion` for every error including a possibly-live order.

**`resume_converting` derives terminality from the same source as the polls.** It gated on `CryptoOrderOutcome::Pending` alone, so a `suspended` / `replaced` / `calculated` order seen on the first read was failed on sight — contradicting the single-source terminality rule #1174 introduced. `CryptoOrderOutcome::is_terminal()` is added so callers outside the execution crate ask the same question the polls do rather than re-deriving it from the variant.

**Sub-minimum USDC is accounted for, not silently stranded.** Both paths alert with the converted amount named, since the USDC sits in the Alpaca crypto wallet in a denomination the offchain cash inventory does not read.

# Testing

- The sub-minimum resume test now asserts the **aggregate state**, not merely that no HTTP call happened — the weaker assertion passed under both the correct and the broken design, which is what let the original defect through.
- Fresh-path sub-minimum terminalizes to `ConversionFailed` with no withdrawal attempted.
- Unresolved outcomes leave the aggregate at `Converting` on both legs, specifically not `ConversionFailed`.
- `resume_converting` waits out a `suspended` order rather than failing on sight.
- Three job-level tests cover both jobs latching on both new variants.

# Anything else

One judgement call worth review: on the fresh path a sub-minimum conversion emits `FailConversion`, which releases the guard, and the operator is alerted. The alternative — latching at `Converting` and holding the guard — would block all USDC rebalancing until someone intervenes. For under $51 stranded, alerting was judged better than wedging; the opposite trade is defensible.

None of this is reachable on master today. It becomes reachable only once partial fills are accepted, which is what the parent PR does.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USDC conversion handling for below-minimum amounts, unresolved broker statuses, suspended orders, and partial fills.
  * Prevented unsafe retries when conversion outcomes cannot be confirmed.
  * Added clearer reconciliation failures and operator alerts for unresolved conversions.
  * Improved recovery behavior when resuming conversions, including withdrawal-minimum validation and terminal order-state checks.
  * Preserved conversions in progress when cancellation does not settle, enabling accurate reconciliation rather than premature completion.

### DIFF
--- a/crates/execution/src/alpaca_broker_api/order.rs
+++ b/crates/execution/src/alpaca_broker_api/order.rs
@@ -352,6 +352,17 @@ enum TerminalCryptoOutcome {
 }
 
 impl CryptoOrderOutcome {
+    /// Whether this classification is an answer, or the order can still
+    /// change state.
+    ///
+    /// Exposed so callers outside this crate decide "keep waiting?" from the
+    /// same rule the polls use, rather than re-deriving it from the variant.
+    /// Matching on [`Self::Pending`] alone is not equivalent: a `Failed`
+    /// status the broker may still resume from is also not an answer.
+    pub fn is_terminal(self) -> bool {
+        self.terminal().is_some()
+    }
+
     /// The terminal outcome this classification represents, or `None` while
     /// the order can still change state.
     ///
--- a/src/rebalancing/usdc/job.rs
+++ b/src/rebalancing/usdc/job.rs
@@ -776,6 +776,27 @@ impl Job<TransferUsdcToHedgingCtx> for TransferUsdcToHedging {
                     warn!(target: "rebalance", ?error, "Failed to deliver USDC hedging inconclusive-burn alert");
                 }
             }
+            // The post-deposit conversion's fate is unknown and the order may
+            // still fill, so retrying would race a live order and recording a
+            // failure would terminalize the rebalance against one. Latched for
+            // the operator, matching the Alpaca->Base leg.
+            Err(UsdcTransferError::ConversionOutcomeUnresolved { id, source }) => {
+                error!(
+                    target: "rebalance",
+                    %id, %source,
+                    "Base->Alpaca USDC transfer: conversion outcome unresolved; the broker \
+                     order may still be live. Latched for operator reconciliation \
+                     (no auto-retry, no failure recorded)"
+                );
+                let message = format!(
+                    "USDC transfer {id}: post-deposit conversion outcome unresolved ({source}). \
+                     The broker order may still be live and may still fill; verify at Alpaca \
+                     before forcing this rebalance either way."
+                );
+                if let Err(error) = ctx.notifier.notify(&message).await {
+                    warn!(target: "rebalance", ?error, "Failed to deliver USDC hedging unresolved-conversion alert");
+                }
+            }
             // Deterministic vault-liquidity revert: the withdraw is atomic
             // (nothing left the vault) and re-issuing it just reverts again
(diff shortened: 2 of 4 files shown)


agryaznov:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1184?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1184** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1184?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1184?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1174** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1174?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1184).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

JuaniRios:
@coderabbitai full review

JuaniRios (approved):

PR #1201, opened 2026-08-13
fix: use Alpaca filled quantity for dividend tokenization

findolor:
## What

[RAI-1717](https://linear.app/makeitrain/issue/RAI-1717/use-alpacas-filled-quantity-for-dividend-tokenization)

Uses Alpaca's confirmed `filled_qty` as the authoritative quantity for both tokenization and wrapper donation in `dividend-bump`.

## Why

Alpaca truncates fractional equity quantities to its supported precision. The command previously continued with the original CLI quantity after the buy filled, so tokenization could request more shares than Alpaca actually acquired.

## How

- Adds a positive filled quantity to `OrderState::Filled` and populates it from Alpaca's order response.
- Fails closed if a filled order omits `filled_qty`, reports a non-positive value, or reports a value inconsistent with the broker order quantity.
- Returns the confirmed fill from the buy-and-poll workflow and passes that exact value through tokenization and donation.
- Prints requested, placed, and filled quantities when precision adjustments make them differ.
- Makes dry-run order IDs self-describing so filled quantities remain recoverable after executor recreation without retaining unbounded in-memory order history.
- Documents the invariant in `SPEC.md` and the operator behavior in `docs/cli-ops.md`.

## Testing

- Added Alpaca response tests covering preserved, missing, zero, and mismatched filled quantities.
- Added a dividend workflow regression test proving `0.0041` is used for tokenization and donation when the request was `0.004115451077565126`.
- Added buy workflow coverage for requested/placed/filled quantity reporting.
- Added dry-run market/limit/restart/error-path coverage for self-describing order IDs.
- Ran `nix run .#ci` successfully:
  - 4,515 backend tests passed (7 skipped)
  - workspace Clippy and formatting passed
  - DTO freshness, dashboard lint, and Svelte checks passed with 0 errors and 0 warnings

## Screenshots

Not applicable; this is CLI and execution-domain behavior.

## Anything else

Reviewed through the staged local review loop with Codex reviewers and DeepSeek V4 Pro (`opencode-go/deepseek-v4-pro`) across three rounds. All validated findings were resolved before submission.

## Summary by CodeRabbit

- **New Features**
  - Dividend bump workflows now use the broker-confirmed filled share quantity for tokenization and donation.
  - CLI output highlights requested, placed, and filled quantities when they differ.

- **Bug Fixes**
  - Invalid, missing, zero, or inconsistent fill quantities now stop processing before tokenization.
  - Dry-run order-status checks now fail clearly instead of returning simulated results.
  - Filled quantities are validated as positive values throughout order processing.

- **Documentation**
  - Updated dividend bump guidance to explain quantity validation, discrepancy reporting, and dry-run status behavior.


findolor:
* **#1201** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1201?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1201?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1201).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor:
@codex Please perform a full code review of this PR.

Review goals:
- Follow all applicable AGENTS.md instructions, including any nested AGENTS.md files that apply to changed files.
- Inspect repository convention files, contribution docs, lint/config rules, and nearby existing code patterns before judging the implementation.
- Verify the PR follows the repo's established architecture, naming, error handling, testing, styling, and abstraction conventions.
- Prioritize actionable findings: bugs, regressions, security issues, data-loss risks, race conditions, edge cases, missing validation, broken API contracts, and insufficient tests.
- Check whether the code introduces unnecessary duplication, weak abstractions, over-engineering, or behavior that should reuse existing helpers/components.
- Look for security and privacy issues, including unsafe input handling, auth/permission gaps, secret exposure, injection risks, SSRF/path traversal/XSS/CSRF where relevant, and overly broad logging.
- Review tests for meaningful coverage of the changed behavior, including failure paths and important edge cases.
- Call out compatibility or migration risks if schemas, APIs, configs, background jobs, or user-facing behavior changed.

Please structure the review like this:
1. Findings first, ordered by severity.
2. For each finding, include file/line references, the concrete risk, and a suggested fix.
3. If there are no blocking issues, say that explicitly.
4. Include any open questions or assumptions.
5. Keep summary brief and avoid restating the diff.

Do not leave style-only comments unless they conflict with repository conventions or materially affect maintainability.

src/offchain/order/poll_status.rs
```diff
@@ -343,6 +343,7 @@ impl PollOrderStatus {
                 price,
                 order_id,
                 executed_at,
+                ..
```
0xgleb: should fix: the production poller discards the new `shares_filled` here with `..`, `validate_broker_terminal_fill` returns None for the Filled arm, and `ReconcileOrderFill` carries no quantity, so the position still completes at the aggregate's requested share count. The truncation bug this PR exists to fix therefore remains unfixed on the hedging path, and the only code enforcing the new invariant is the test-only harness in `integration_tests/arbitrage.rs`. Either thread `shares_filled` into the `ReconcileOrderFill` payload and use it for `CompleteOffChainOrder` (adding the mismatch guard alongside `validate_broker_terminal_fill`), or record explicitly that the hedge path is out of scope and drop the harness-only guard so tests stop advertising behaviour production lacks.

src/cli/dividend.rs
```diff
@@ -10,49 +10,134 @@
...
 
     writeln!(
         stdout,
-        "Step 3/3: donating {quantity} {symbol} into the wrapper"
+        "Step 3/3: donating {filled_quantity} {symbol} into the wrapper"
     )?;
-    wrapper::donate_equity_command(stdout, symbol, quantity, ctx).await?;
+    operations
+        .donate(stdout, symbol, filled_quantity, ctx)
+        .await?;
```
0xgleb: critical: `alpaca_tokenize_command` returns `Ok(())` when Alpaca rejects the tokenization and when the 60-attempt balance poll times out, and this flow treats that as success and proceeds to donate `filled_quantity`. The donation is a bare ERC-20 transfer from the market-making wallet, so a rejected or slow tokenization silently converts into an irreversible transfer of hedging inventory into the wrapper, with no position-accounting event. The tokenize step should return a typed outcome (completed with the measured onchain delta / rejected / timed out), stop before the donate step on anything but completed, and donate the measured delta rather than the pre-tokenization quantity.

crates/execution/src/alpaca_broker_api/executor.rs
```diff
@@ -158,6 +158,25 @@ impl Executor for AlpacaBrokerApi {
...
+                    .inner()
+                    .inner()
+                    .eq(order_update.shares.inner().inner())?
+                {
+                    return Err(AlpacaBrokerApiError::FilledQuantityMismatch {
+                        order_id: ExecutorOrderId::new(order_id),
+                        ordered: order_update.shares,
+                        filled: shares_filled,
+                    });
+                }
```
0xgleb: critical: this method is not CLI-only, it is what the server's `PollOrderStatus` job calls for every hedge, and all three new failures are Permanent and non-backpressure, so after the 3-retry budget they open the circuit breaker, which is the RAI-1492 "hedging silently stops" outcome. `get_order_status` is a pure read of terminal data, so re-polling can never produce a different answer, and the `filled_qty == qty` invariant is uncited and proven only against hand-written httpmock fixtures. Consider recording the broker-reported fill and surfacing the divergence loudly (or hard-failing only the unsafe direction, `filled_qty > qty`), and cite the Alpaca order-lifecycle docs for the claim that `filled` implies fully filled. The zero-quantity case also degrades to the context-free `NotPositive` variant with no order id, unlike its two neighbours.

src/integration_tests/arbitrage.rs
```diff
@@ -211,7 +211,16 @@ async fn poll_submitted_orders<E: st0x_execution::Executor + Clone>(
...
                 executed_at,
+                shares_filled,
             } => {
+                let placed_shares = order.shares();
+                if shares_filled != placed_shares {
+                    return Err(format!(
+                        "order {order_id} reported filled quantity {shares_filled}, but the broker-accepted quantity was {placed_shares}"
+                    )
+                    .into());
+                }
```
0xgleb: should fix: this helper's doc promises it applies the same commands the production job chain would emit, but the new mismatch rejection has no production counterpart, so `poll_helper_rejects_mismatched_filled_quantity` asserts a property only the harness has. A reviewer reading that test will conclude mismatched fills are rejected in production when they are silently accepted. Either mirror the guard into `dispatch_for_broker_state`/`ReconcileOrderFill` or drop it here and test the executor-level invariant where it actually lives.

crates/execution/src/mock.rs
```diff
@@ -178,9 +192,32 @@ impl MockExecutor {
...
+            .parse()
+            .map_err(|source| MockOrderIdError::InvalidQuantity {
+                order_id: order_id.to_string(),
+                source,
+            })?;
+        Positive::new(shares).map_err(|source| MockOrderIdError::NonPositiveQuantity {
+            order_id: order_id.to_string(),
+            source,
+        })
     }
```
0xgleb: should fix: dry-run is a deployed broker mode with orders persisted through `MarkAccepted`, and any order accepted before this change carries a `TEST_<n>` id with no colon, so `shares_from_order_id` now errors with `MissingQuantity` on every poll of a pre-change order and the poll job wedges. Tolerate the legacy shape, fall back to the aggregate's placed quantity, or classify `MockOrderIdError` as terminal and non-retrying in `PollOrderStatus`.

crates/execution/src/mock.rs
```diff
@@ -178,9 +192,32 @@ impl MockExecutor {
...
 
-    fn generate_order_id(&self) -> String {
-        let id = self.order_counter.fetch_add(1, Ordering::SeqCst);
-        format!("TEST_{id}")
+    fn order_id(
+        client_order_id: &crate::ClientOrderId,
+        shares: Positive<crate::FractionalShares>,
+    ) -> String {
+        format!("TEST:{client_order_id}:{shares}")
+    }
```
0xgleb: minor: the quantity round-trips through a display-formatted string inside the order id, while `parse_order_id` still accepts any input that `get_order_status` will later reject, and `MockOrderIdError` carries `order_id: String` where `ExecutorOrderId` is used a few lines away. A `MockOrderId` newtype with `Display`/`FromStr` (and `Float::as_hex` for the quantity so the round trip is bit-exact rather than dependent on `format_float` staying reparseable) would move the rejection to the boundary and delete the error path in `get_order_status` entirely.

crates/execution/src/mock.rs
```diff
@@ -260,7 +297,7 @@ impl Executor for MockExecutor {
     ) -> Result<OrderPlacement<Self::OrderId>, Self::Error> {
         self.fail_if_unhealthy()?;
 
-        let order_id = self.generate_order_id();
+        let order_id = Self::order_id(&order.client_order_id, order.shares);
 
         debug!(
             target: "broker",
```
0xgleb: minor: the mock echoes the caller's exact quantity as placed and reports the same value back as filled, never applying `truncate_shares_to_alpaca_precision`, so a dry-run rehearsal of dividend-bump can never reproduce the quantity adjustment this PR exists to surface, and the new reporting block is dead under dry run. Applying the same truncation in the mock's place methods makes the rehearsal match live behaviour.

src/cli/dividend.rs
```diff
@@ -10,49 +10,134 @@
...
+        &self,
+        stdout: &mut Writer,
+        symbol: Symbol,
+        quantity: Positive<FractionalShares>,
+        ctx: &Ctx,
+    ) -> anyhow::Result<()>;
+
+    async fn donate<Writer: Write + Send>(
+        &self,
+        stdout: &mut Writer,
```
0xgleb: minor: `DividendBumpOperations` is a CLI-local trait over three one-line delegations, and it costs `Sync` supertraits and `Writer: Write + Send` bounds on unrelated commands. The success-path test also runs entirely against the recording stub, so the real buy -> tokenize -> donate wiring in `LiveDividendBumpOperations` is only exercised by the one failure test. Parameterizing on the executor/tokenizer/wrapper abstractions that already exist (the mock already supports `with_order_status`) would let a success-path test drive the real command with a filled quantity that differs from the request.

src/cli/trading.rs
```diff
@@ -207,6 +207,7 @@ fn write_order_status<W: Write>(stdout: &mut W, state: OrderState) -> anyhow::Re
             executed_at,
             order_id,
```
0xgleb: minor: the Filled arm skips `shares_filled` with `..` so order-status output never prints the broker-confirmed quantity, while the PartiallyFilled and Cancelled arms both print a Shares Filled line, and there is no test covering the Filled arm at all. Either print it or pin the intentional omission with a test.

src/cli/trading.rs
```diff
@@ -660,9 +662,27 @@ async fn place_market_order_until_filled<Exec: Executor, W: Write>(
...
+                let placed_matches_filled = placement
+                    .shares
+                    .inner()
+                    .inner()
+                    .eq(shares_filled.inner().inner())?;
+                if !requested_matches_placed || !placed_matches_filled {
+                    writeln!(stdout, "   Requested quantity: {requested_shares}")?;
+                    writeln!(stdout, "   Placed quantity: {}", placement.shares)?;
+                    writeln!(stdout, "   Filled quantity: {shares_filled}")?;
+                }
```
0xgleb: minor: with the real executor the placed-vs-filled divergence this report and its test model is unreachable, because placement returns the locally truncated quantity and the new Filled-arm check rejects any response where `filled_qty` differs from `qty`. The reachable case is requested-vs-placed truncation, so the test should model that instead of forcing an impossible combination through `with_placed_shares`.

src/cli/trading.rs
```diff
@@ -281,10 +282,10 @@ async fn get_broker_order_status<W: Write>(
...
             .await?)
         }
-        BrokerCtx::DryRun => {
-            let broker = MockExecutorCtx.try_into_executor().await?;
-            Ok(broker.get_order_status(&order_id.to_string()).await?)
-        }
+        BrokerCtx::DryRun => anyhow::bail!(
+            "order status is unavailable in dry-run mode because dry-run does not persist broker \
+             order state"
+        ),
```
0xgleb: nit: the docs justify rejecting dry-run status lookups with "dry-run does not persist broker order state", but this same PR makes dry-run ids self-describing precisely so the quantity survives executor recreation, and the dry-run server path resolves them fine. Either let the lookup work for ids the dry-run executor issued, or reword the rationale to the actual constraint.

SPEC.md
```diff
@@ -239,6 +239,21 @@ distinct — neither key alone can both mint and authorize.
...
   warning instead of staying silently inert.
 
+#### Dividend NAV Bump
+
+The issuer's `dividend-bump` command buys the underlying equity, tokenizes the
+acquired shares, and donates the resulting tokenized equity into its ERC-4626
+wrapper. The broker-confirmed filled quantity is authoritative for both the
+tokenization and donation steps; the originally requested quantity must never be
+reused after placement because Alpaca may truncate it to the precision the
+broker supports.
```
0xgleb: minor: "Alpaca may truncate it to the precision the broker supports" is uncited, and the only truncation in the code is our own client-side `truncate_shares_to_alpaca_precision` to the local `ALPACA_MAX_DECIMAL_PLACES = 9`, so the reported "Placed quantity" is our pre-send truncation, not broker behaviour. Also the filled-quantity invariant documented here is implemented in `get_order_status`, so it applies to every polled Alpaca order including the hedge loop, and belongs in the broker status-handling section with the dividend-bump section referring to it.

JuaniRios (approved):

PR #1207, opened 2026-08-14
RAI-1744: page trade history from the materialized views

agryaznov:
## What

Closes RAI-1744. `GET /trades` and the dashboard's WebSocket seed now page terminal trades out of the materialized `onchain_trade_view` and `offchain_order_view` projections, filtering, sorting and limiting in SQL.

## Why

Trade history rebuilt the full list on every request: `load_all_ids::<OnChainTrade>` plus one `load_entity` per id, ~10k aggregate replays, 6-7s on prod. The dashboard aborts at its 5s `FETCH_TIMEOUT_MS`, so the trade panel flickered.

The cost was almost entirely the onchain half. The offchain half was already a single indexed read of `offchain_order_view`.

## How

`OnChainTrade` was `PROJECTION = Nil`. Setting `PROJECTION = Table("onchain_trade_view")` hands the read model to the framework, which already maintains one for `OffchainOrder` — including ordering against commit order, retry, crash catch-up, schema-version rebuilds, and backfilling every pre-existing aggregate inside `StoreBuilder::build`.

`onchain_trade_view` turned out to already exist: created in `20251103172115`, json paths repaired in `20260212115445`, never written to and referenced by no Rust code. The read model was half-built and never wired. The migration is therefore a DROP + CREATE of a table that is empty in every environment.

Both views gain `STORED GENERATED` sort and filter keys — `occurred_at`, `symbol`, `venue`, and `tx_hash`/`log_index` derived from `view_id`. They are pure read keys: rows still hold the serialized aggregate, and `try_into_trade` runs at read time over the returned page only, so the wire shape keeps exactly one source of truth.

Key details worth a look:

- **`occurred_at` normalization.** chrono's serde emits `SecondsFormat::AutoSi` — 0, 3, 6 or 9 fractional digits — which is not sortable as stored: `...20.5Z` sorts before `...20Z`, and `...20.500Z` after `...20.500000Z`. The generated column pads to a constant 9 digits. A proptest pins SQL ordering against `DateTime<Utc>` ordering.
- **Ordering equivalence.** `ORDER BY occurred_at DESC, sort_group ASC, log_index DESC` reproduces `sort_trades_newest_first` exactly, including the numeric log-index tie-break within one transaction (`:20` before `:11` before `:2`). `tx_hash` is generated from `view_id` rather than re-formatted from the payload, so the sort group and the trade id's prefix are the same bytes by construction.
- **Query shape.** Each side is ordered and truncated to `offset + limit` on its own index before the union, so the merge sorts two bounded lists rather than the whole history.
- **Index steering.** `EXPLAIN QUERY PLAN` showed SQLite preferring `idx_offchain_order_view_status` over the ordering index whenever a `status IN (...)` term was present, then sorting every match in a temp b-tree. The predicate is now dropped entirely for protocols carrying every terminal outcome (implied by `occurred_at IS NOT NULL`) and written `+status` for the narrowing ones, where SQLite's unary `+` strips index affinity without changing meaning.
- **Venue mapping is self-policing.** The onchain `venue` column restates `OnChainTradeSource::trading_venue()` in SQL. Every returned row cross-checks the indexed venue against the replayed one and errors on mismatch, so an unmapped variant cannot silently become invisible to venue filters while still appearing in unfiltered history.

Response shapes, protocol semantics and the `limit`/`offset` contract are unchanged.

## Testing

21 tests in `trade_loader.rs`, covering the migrated coverage from the deleted replay path plus the new SQL-specific cases: comparator equivalence, legacy venue collapse per protocol, nanosecond-inclusive `since`/`until`, pagination without gaps or repeats, `usize::MAX` offset, unreadable-row skip, the venue-mismatch error, an exhaustive `OnChainTradeSource` mapping check that fails to compile when a variant is added, and `EXPLAIN QUERY PLAN` assertions on all three branch shapes.

Every test that previously covered trade history has an explicit home — migrated, rewritten against the real SQL, or retired with its replacement named. `failed_offchain_trade_history_respects_all_filters` moved from an in-memory predicate to the real query, which strengthens it.

### Outstanding: needs someone with prod access

**`nix run .#prodVerifyMigrations` has NOT been run — I do not have prod access. Please run it before merge if you do.**

This matters more than the usual migration check. The first boot after this deploys replays every persisted `OnChainTrade` aggregate to backfill `onchain_trade_view`, so "does all prod data still replay under current code?" is literally the operation that boot performs. `sqlx db reset -y` passes, but that only proves the migrations apply to an empty schema — it says nothing about the ~10k real aggregates.

`nix run .#stagingVerifyMigrations` is a weaker but real substitute if prod access is awkward.

Mitigating: the same binary already gates deploys (`deploy.nix:71`), so this fails **closed** at deploy time rather than corrupting anything. The cost of not running it up front is a failed deploy, not bad data. If it does fail, the fix is a repair migration — `migrations/20260701223808_repair_legacy_usdc_conversion_confirmed_events.sql` is the template.

## Anything else

**Behaviour change to review.** A corrupt payload used to be skipped per request; under a projection it fails `catch_up` loudly at boot. Strictly more correct, but it is what makes the verification above worth doing.

The `venue` CASE duplicates a Rust match in SQL. Guarded by the exhaustive variant test and the per-row cross-check, but a reviewer should agree it is the right trade against deserializing 21k payloads just to filter.

This supersedes the originally planned design (a bespoke checkpointed ingester over `events_since` + a `head_rowid` watermark, mirroring ADR 0018). Review found that version unsound — `events_since` hard-errors on the first undeserializable payload, so its "skip corrupt aggregates and advance" semantic was not implementable, and skipping on a `load_entity` failure would permanently drop a trade on a transient `SQLITE_BUSY`. The framework already provided the mechanism, so no new architectural pattern is introduced and no ADR is owed.


agryaznov:
* **#1207** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1207?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.liquidity/1207?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=1207).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

agryaznov:
@coderabbitai review

agryaznov:
@coderabbitai full review

src/dashboard/trade_loader.rs
```diff
@@ -1,832 +1,1379 @@
...
+            .collect::<Vec<_>>()
+            .join(" AND ")
+    }
+}
+
+/// `AssertSqlSafe`: the statement is assembled from `&'static str` fragments
+/// -- table and column names chosen by [`Side`] and this module's own literals
+/// -- and every caller-supplied value is a bound `?` placeholder. No user input
+/// reaches the SQL text.
+async fn count_matching(
```
agryaznov, line 295: **`total` counts rows the page can never return.** `count_matching` counts matching view rows; `convert_row` (:377) then drops any the page cannot turn into a `Trade`. `api.rs:640` mixes the two, deriving `has_more` from the converted length against this unconverted count. Master counted after the drop, so it was exact.

On the two rows `unreadable_rows_are_skipped_with_the_rest_of_the_page_intact` (:1157) already inserts: `total: 2, entries: [1 trade], has_more: true` for a history with one displayable trade. Master gave `total: 1, has_more: false`.

What gets dropped is not only corruption — `try_into_trade` fails on a non-positive amount (`onchain_trade.rs:325`) and on `terminal_quantity_provenance` offchain. Those are well-formed aggregates with no trade representation, so failing the request instead is not an option.

`total` is a UI counter: `trade-history-panel.svelte:479` renders it as "N of M" in the panel header (`102 of 24961` in prod). Not a financial value, and the client already treats it as soft — it increments on live arrivals and reconciles with `Math.max`. So the visible defect is not the number, it is that `hasMore = data.hasMore || nextEntries.length < reconciledTotal` (:238) keeps "Load more" alive forever and fetches empty pages.

Resolution: keep the SQL count, derive `has_more` from the raw fetched row count so the server flag is truthful, and have the client trust that flag rather than re-deriving it from a length comparison. Making `total` exact would mean expressing "convertible" as a SQL predicate over a serialized `Float` — the drift the venue cross-check exists to prevent.

src/dashboard/trade_loader.rs
```diff
@@ -1,832 +1,1379 @@
...
+            limit: usize::MAX,
+            offset: 0,
+        }
+    }
 
-    Ok(trades)
+    /// The newest page a dashboard client gets on connect.
+    pub(crate) fn newest(trade_protocol: TradeProtocol) -> Self {
+        Self {
+            limit: MAX_TRADES,
```
agryaznov, line 118: **The `MAX_TRADES` cap lost its test.** `load_trades_capped_at_max` (master, `trade_loader.rs:582`) seeded `MAX_TRADES + 1` fills and asserted the seed returned exactly `MAX_TRADES`. Nothing in the new suite asserts it.

`MAX_TRADES` now appears three times in the file (:31, :117, :985) and the only test-module use is a loop bound. Change this line to `usize::MAX` and every test still passes while every dashboard WebSocket connect seeds the client with the full history.

This is the one gap in the PR description's claim that every deleted test has an explicit home — the other 11 old loader tests do, and `api.rs` lost none.

src/dashboard/trade_loader.rs
```diff
@@ -1,832 +1,1379 @@
...
     }
 
+    /// Paging in SQL is only a win if each branch walks its ordering index and
+    /// stops at the limit. The offchain side is the fragile one: any
+    /// index-eligible `status` term makes SQLite prefer
+    /// `idx_offchain_order_view_status` and sort every match in a temp b-tree,
+    /// which is why the narrowed protocols write `+status`.
     #[tokio::test]
-    async fn load_trades_excludes_all_nonterminal_offchain_orders() {
+    async fn every_branch_walks_its_ordering_index_instead_of_sorting() {
```
agryaznov, line 1339: **This test asserts against SQL the code never emits.** The three statements are hand-written; production SQL comes from `branch_sql` (:352) via `side_filter` (:159), and nothing links them.

The offchain case here orders by `occurred_at DESC, view_id ASC`; `branch_sql` emits `occurred_at DESC, sort_group ASC, log_index DESC` wrapped in `SELECT * FROM (...)`, with `log_index` the literal `NULL`. Measured against the real schema:

```
real branch SQL : SEARCH ... USING INDEX idx_offchain_order_view_occurred_at | USE TEMP B-TREE FOR LAST TERM OF ORDER BY
this test's SQL : SEARCH ... USING INDEX idx_offchain_order_view_occurred_at
```

Both assertions still pass on the real SQL (the step is named "FOR LAST TERM OF ORDER BY", which does not contain the forbidden substring), and early `LIMIT` termination does survive — 200k offchain rows: `LIMIT 100` 0.000s, `LIMIT 200000` 0.529s. So no live perf bug.

But the guard cannot do its job: delete the `+` from `side_filter:184` and the plan regresses to `idx_offchain_order_view_status` + a full temp b-tree sort, with this test still green. Build the probe from `branch_sql(side, &filter)` so it is wired to the code that can regress. The outer union query is also uncovered.

One caveat worth recording: the `+status` steering only matters on a database without `ANALYZE` stats. With stats present SQLite picks the ordering index either way.

src/dashboard/trade_loader.rs
```diff
@@ -1,832 +1,1379 @@
...
+            let mut bounds = [format!("{hash}:2"), format!("{hash}:11")];
+            bounds.sort();
+            let [low, high] = &bounds;
+            let between = uuid > low.as_str() && uuid < high.as_str();
+            assert!(!between, "{uuid} must not sort between ids sharing a hash");
+        }
+    }
+
+    #[test]
+    fn a_malformed_onchain_view_id_is_a_row_error_not_a_panic() {
```
agryaznov: **This test never reaches the path it names.** `parse_row` (:420) deserializes the payload before parsing the id, and `{"Live":{}}` cannot produce an `OnChainTrade` — `symbol`, `amount`, `direction`, `price_usdc`, `block_timestamp` and `filled_at` are all missing. So it always returns `Payload(_)` and the `OnchainId` arm is dead.

The `|` in the `matches!` hides that, which is what AGENTS.md's "assert the exact variant" and "never use `||` in assertions unless outcomes are genuinely equivalent" are aimed at. Give it a valid payload with a malformed `view_id` if the id path is what should be covered.

migrations/20260814163552_onchain_trade_view_trade_history_keys.sql
```diff
@@ -0,0 +1,78 @@
...
+    -- Derived from view_id rather than re-formatted from the payload, so the
+    -- sort group and the trade id's prefix are the same bytes by construction.
+    -- That identity is what makes ordering by tx_hash equivalent to
+    -- `sort_trades_newest_first`'s comparison of whole ids.
+    tx_hash TEXT GENERATED ALWAYS AS
+        (substr(view_id, 1, instr(view_id, ':') - 1)) STORED,
+    log_index INTEGER GENERATED ALWAYS AS
+        (CAST(substr(view_id, instr(view_id, ':') + 1) AS INTEGER)) STORED
+);
+
```
agryaznov, line 71: **This index cannot serve the count, so the onchain `COUNT(*)` is a full table scan.** Its offchain counterpart is `WHERE occurred_at IS NOT NULL`, which covers the equivalent count; this one has no predicate. Measured:

```
COUNT(*) FROM onchain_trade_view  WHERE occurred_at IS NOT NULL  ->  SCAN onchain_trade_view
COUNT(*) FROM offchain_order_view WHERE occurred_at IS NOT NULL  ->  SEARCH ... USING INDEX (partial)
```

`count_matching` runs on every `/trades` request and every WebSocket connect, so this scans every row and its JSON payload pages, growing linearly with history. At ~10k rows that is milliseconds and not a regression — but SPEC.md and the commit message both claim the request costs "one bounded index scan per venue side", which is not what happens. Either add the partial predicate here or soften the claim.

migrations/20260814163608_offchain_order_view_trade_history_keys.sql
```diff
@@ -0,0 +1,79 @@
...
+    symbol TEXT GENERATED ALWAYS AS (
+        coalesce(
+            json_extract(payload, '$.Live.Filled.symbol'),
+            json_extract(payload, '$.Live.Failed.symbol'),
+            json_extract(payload, '$.Live.Cancelled.symbol')
+        )
+    ) STORED,
+
+    -- Mirrors the SupportedExecutor -> TradingVenue match in
+    -- OffchainOrder::try_into_trade.
```
agryaznov, line 61: **This mapping has no exhaustiveness guard, unlike its onchain twin.** `indexed_venue_matches_trading_venue_for_every_source` pins the onchain `CASE` to `OnChainTradeSource::trading_venue()` with a match that fails to compile when a variant is added. Nothing does that for this `CASE` against `SupportedExecutor`.

Every offchain test path uses `AlpacaBrokerApi`, so the `'DryRun' -> 'dry_run'` arm is never executed by the suite. The mapping is correct today (`crates/execution/src/lib.rs:469`), but drift here leaves `venue` NULL, which `convert_row:405` turns into a hard error rather than a skipped row — see the comment there for the blast radius.

src/dashboard/trade_loader.rs
```diff
@@ -1,832 +1,1379 @@
...
+        }
+    };
+
+    // The views derive `venue` in SQL, duplicating the mapping `try_into_trade`
+    // applies in Rust. Comparing them on every row returned makes that
+    // duplication self-policing: an unmapped source variant leaves the column
+    // NULL and would otherwise be silently invisible to venue filters while
+    // still appearing in unfiltered history.
+    if venue != Some(trade.venue.to_string().as_str()) {
+        return Err(TradeHistoryError::VenueMismatch {
```
agryaznov, line 434: **A NULL indexed venue is fatal, and now takes down more than `/trades`.** This propagates out of `fetch_page`, so one bad row fails the whole request with a 500. The comment above explains the fail-closed intent for history rendering.

But `query_trades` also backs `reconcile_undelivered` (`dashboard/event.rs:222`), whose error surfaces at `event.rs:610` as `DashboardTradeHandoffMonitorError::TerminalReconciliation`. So an unmapped source variant also breaks dashboard trade *delivery reconciliation* at startup, not just history.

Note the asymmetry this creates: a payload that fails to parse is skipped (`Ok(None)`), a payload that parses with an unmapped venue is fatal. That is defensible — it is exactly the drift case — but the wider blast radius should be stated here, and the offchain `CASE` needs the same exhaustiveness guard the onchain one has.

src/conductor.rs
```diff
@@ -252,14 +252,22 @@ async fn setup_vault_registry(
     ))
 }
 
+/// Builds the onchain trade store and its `onchain_trade_view` projection.
+///
+/// `build` catches the view up to the event log before returning, which on the
```
agryaznov, line 258: **Worth timing before merge: this backfill is serial and untransacted, inside a 5-minute deploy window.** `Projection::catch_up` iterates stale aggregates in a plain `for` loop, doing a full event replay plus a `SELECT` and an upsert per aggregate, with one implicit WAL commit each — no concurrency, no wrapping transaction.

The read path this replaces used `buffer_unordered(8)` and still cost 6-7s for ~10k aggregates. A one-time serial backfill with ~10k fsyncs is plausibly an order of magnitude slower. This runs before readiness, and `deploy.nix:52-76` fails activation if the unit does not report ready within 5 minutes.

Not confirmed — I could not measure it — but cheap to settle: run `nix run .#prodVerifyMigrations`, then time a `StoreBuilder::<OnChainTrade>::build` against the downloaded snapshot.

Partial failure is safe: `catch_up` is incremental and idempotent, and `record_version` is only written after recovery completes.

src/dashboard/trade_loader.rs
```diff
@@ -1,832 +1,1379 @@
...
-                    let error = TradeHistoryError::OnchainConversion { id, source };
-                    warn!(target: "dashboard", %error, "Skipping unrepresentable onchain trade");
-                    None
-                }
-            }
+    page.fetch_all(pool)
+        .await?
+        .into_iter()
+        .filter_map(|row| {
+            let view_id: String = row.get("view_id");
```
agryaznov: `Row::get` panics on a decode or column-name error, in a request path. Unreachable given the schema — `payload JSON NOT NULL`, `view_id` primary key, `venue` typed `Option`, `side` a literal — but `try_get` + `?` is a one-word change and `TradeHistoryError::Database(#[from] sqlx::Error)` already exists to carry it.

src/dashboard/trade_loader.rs
```diff
@@ -1,832 +1,1379 @@
...
+            filter.where_sql()
+        );
+        let mut counted = sqlx::query_scalar::<_, i64>(sqlx::AssertSqlSafe(sql));
+        for bind in &filter.binds {
+            counted = counted.bind(bind);
+        }
+        total += counted.fetch_one(pool).await?;
+    }
+
+    Ok(usize::try_from(total).unwrap_or(usize::MAX))
```
agryaznov: `unwrap_or` on a numeric conversion is the shape AGENTS.md bans. Unreachable in practice (`COUNT(*)` is never negative, and `i64 -> usize` cannot fail on a 64-bit target), but an error variant costs nothing here.

`clamp_to_i64` at :369 is the same shape and I would leave it — it is documented as deliberate saturation and covered by `offset_beyond_the_end_returns_an_empty_page`.

src/dashboard/trade_loader.rs
```diff
@@ -1,832 +1,1379 @@
...
 use st0x_finance::{FractionalShares, NotPositive};
 
 use super::TradeProtocol;
-use crate::offchain::order::{OffchainOrder, TradeConversionError};
-use crate::onchain_trade::{OnChainTrade, OnChainTradeId};
+use crate::offchain::order::{OffchainOrder, OffchainOrderId, TradeConversionError};
+use crate::onchain_trade::{OnChainTrade, OnChainTradeId, ParseOnChainTradeIdError};
+
+/// Trades returned to a dashboard client that asks for no explicit page.
+pub(crate) const MAX_TRADES: usize = 100;
```
agryaznov: `pub(crate)` but referenced only inside this module (:117, :985). Should be private — AGENTS.md asks for the most restrictive visibility, and it improves dead-code detection.

migrations/20260814163608_offchain_order_view_trade_history_keys.sql
```diff
@@ -0,0 +1,79 @@
+-- Add trade-history sort and filter keys to offchain_order_view so /trades can
+-- page terminal counter-trades in SQL instead of parsing every row per
+-- request. SQLite cannot ALTER TABLE ADD a STORED generated column, so this is
+-- DROP + CREATE -- safe here for the same reason the cancelling/cancelled
+-- migration gave: views are rebuilt from events at startup.
+--
+-- `status` and its index are preserved verbatim; /api/pending_orders depends
```
agryaznov: **"preserved verbatim" overstates what survives.** The `status` column and its index are byte-identical, and `/api/pending_orders` does depend on them — but it also depends on `ORDER BY rowid DESC LIMIT 100` (`api.rs:551`). DROP + CREATE discards rowids, and `catch_up` reinserts in `GROUP BY aggregate_id` order, i.e. UUID order.

So every pending order predating this migration comes back in arbitrary rowid order and that endpoint no longer shows them newest-first. Precedented — `20260212115839` and `20260626024335` did the same — and usually invisible because pending orders are few and new ones get increasing rowids. Worth narrowing the claim to the column and index.

agryaznov (commented):
Self-review of the read path, from two independent passes over the diff. Findings are ranked roughly by severity in the thread order below; the first is the only one that changes runtime behaviour.

What was verified clean, with evidence rather than argument: ordering equivalence with `sort_trades_newest_first` across 300 randomized mixed datasets (0 mismatches), per-branch truncation before the union across 200 datasets x 42 limit/offset combinations (0 mismatches), filter parity with the deleted `TradeFilter` across 720 protocol/venue/symbol/time combinations (0 mismatches, including `total`), and every chrono `AutoSi` width probed through the real migration against `sortable_timestamp`.

JuaniRios (approved):

src/dashboard/trade_loader.rs
agryaznov: Fixed in 0a5dfda4. `fetch_page` reports the rows it consumed and `query_trades` derives `has_more` from that, so the flag is truthful whatever the page drops; `api.rs` reads it instead of doing its own arithmetic. `total` stays the SQL count and now documents that it is an upper bound. The dashboard trusts the flag rather than re-deriving it from a length comparison, and the skip test pins both `total` and `has_more`.

src/dashboard/trade_loader.rs
agryaznov: Fixed in 0a5dfda4 — `the_default_page_is_capped_at_max_trades` seeds `MAX_TRADES + 1` fills and asserts the page, the total, and `has_more`.

src/dashboard/trade_loader.rs
agryaznov: Fixed in 0a5dfda4. The statements now come from `side_filter` and `branch_sql`, so losing the `+` fails the assertion. The offchain branch's `USE TEMP B-TREE FOR LAST TERM OF ORDER BY` is documented in the test rather than removed — its `log_index` is the constant NULL, so it sorts blocks of one row.

src/dashboard/trade_loader.rs
agryaznov: Fixed in 0a5dfda4 — the payload now comes from the projection rather than a literal, so the row deserializes and the id branch is actually reached. Assertion narrowed to the single variant.

migrations/20260814163552_onchain_trade_view_trade_history_keys.sql
agryaznov: Fixed in 0a5dfda4 — the index is partial on `occurred_at IS NOT NULL`, so `count onchain` goes from `SCAN onchain_trade_view` to `SEARCH ... USING INDEX idx_onchain_trade_view_order`. SPEC.md and the module doc now describe a bounded range for the page plus an index-only count.

migrations/20260814163608_offchain_order_view_trade_history_keys.sql
agryaznov: Fixed in 0a5dfda4 — `indexed_venue_matches_trading_venue_for_every_executor` gives the offchain mapping the same compile-forcing guard. It earned its keep immediately: `DryRun` reached the view for the first time, and the helper layering bug that hid it (`fill` re-places with a hardcoded executor) had to be fixed to make it pass.

src/dashboard/trade_loader.rs
agryaznov: Fixed in 0a5dfda4 — the comment now names delivery reconciliation as part of the blast radius and says why the asymmetry against the unreadable-payload skip is deliberate.

src/dashboard/trade_loader.rs
agryaznov: Fixed in 0a5dfda4 — `try_get` with `?`, carried by the existing `Database(#[from] sqlx::Error)` variant.

src/dashboard/trade_loader.rs
agryaznov: Fixed in 0a5dfda4 — a `CountOutOfRange` variant carrying the count and the `TryFromIntError`. `clamp_to_i64` left as is, per the reasoning above.

src/dashboard/trade_loader.rs
agryaznov: Fixed in 0a5dfda4 — private.

migrations/20260814163608_offchain_order_view_trade_history_keys.sql
agryaznov: Fixed in 0a5dfda4 — the comment now claims only the column and index, and records the rowid caveat with its effect on `/api/pending_orders`.

src/conductor.rs
agryaznov: Not fixed — there is no code change to make here, only a measurement I cannot take without prod access.

One data point since: the live dashboard reports 24961 trades, not the ~21k the issue assumed, so the backfill is larger than estimated. Still needs `nix run .#prodVerifyMigrations` and a timed `StoreBuilder::<OnChainTrade>::build` against the snapshot before merge.

Leaving this thread open.

agryaznov:
successful `nix run .#prodVerifyMigrations` run confirmed by @JuaniRios , good to merge

